Issue #29, opened 2019-05-10
[Story #6] User configures Adobe Stock integration


sidolov:
## Story
As a Merchant I want to configure Adobe Stock integration on my Magento instance

## Business Value
no need to maintain API credentials on Magento side
 
## Preconditions 
* Admin User is logged into Admin Panel
* User configured new Adobe Stock integration on adobe.io console

## Acceptance Criteria
* User goes Media Gallery and opens Adobe Stock. User sees message that Adobe Stock disabled and API is missing with the link to Store-Configuration-Advanced-System. User clicks the link and navigates to Adobe Stock Integration configuration options
* User enables Adobe Stock
* User sees the following required fields in empty state:
  * API Keys (x-api-key)
* User sees a message that explains how to get these credentials with a link to [Adobe.io](https://console.adobe.io)
* User copies API key generated on Adobe.io side and inserts it into the API key field
* User sees a message that API key is valid. If user inserted wrong API key, error message displayed.  <strike>Configuration can't be saved with not valid API key</strike>
* <strike>User copies name that they chose for the integration when created it on adobe.io console</strike>
* User saves configuration 
* If API key is invalid, user sees message about that next the API Keys field. Configuration is saved with the wrong key
* User goes Media Gallery and opens Adobe Stock. If API key is invalid, user sees message about that. No image search possible. 
 
https://xd.adobe.com/view/0a8fc470-497b-4fd6-4043-62560c21d8db-7cbe/screen/bf0d5d64-e91b-4b77-9926-a3ca7946d808/Configuration

TomashKhamlai:
@sidolov, considering this acceptance criterion "User sees the following required fields in the empty state: - API Keys (x-api-key)" does it means that this scenario is valid?

User enables Adobe Stock Integration
User copies credentials from console.adobe.io and saves the configuration
User flushes cache
User disables Adobe Stock Integration and saves the configuration
User flushes cache
User enables Adobe Stock Integration
User sees the following required fields in the empty state: - API Keys (x-api-key)

or this criterion means only that Magento doesn't include any credentials out of the box?

sidolov:
@TomashKhamlai I believe this statement valid only for the first time when you don't have any credentials at all. Once you filled the field and saved the configuration you should see the filled field (hidden, password style)

PR #50, opened 2019-05-12
#11 Implement setup script for user metadata

diazwatson:
https://github.com/magento/adobe-stock-integration/issues/11

PR #52, opened 2019-05-12
#29 User configures Adobe Stock integration

diazwatson:
Resolved Issue: #29
Blocked by: https://github.com/magento/magento2/pull/22844


AdobeStockImageAdminUi/etc/adminhtml/system.xml
```diff
@@ -0,0 +1,37 @@
...
+                </field>
+                <field id="api_key" translate="label" type="text" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Api Key</label>
+                    <config_path>adobe_stock/integration/api_key</config_path>
+                    <validate>required-entry</validate>
+                    <depends>
+                        <field id="enabled">1</field>
+                    </depends>
+                </field>
+                <field id="adobe_stock_test_connect_wizard" translate="button_label" sortOrder="68" showInDefault="1" showInWebsite="0" showInStore="0">
```
sivaschenko: Why 68? :)

AdobeStockImageAdminUi/etc/adminhtml/system.xml
```diff
@@ -0,0 +1,37 @@
...
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Config:etc/system_file.xsd">
+    <system>
+        <section id="system">
+            <group id="adobe_stock_integration" translate="label" type="text" sortOrder="1100" showInDefault="1" showInWebsite="1" showInStore="1">
```
sivaschenko: I think these settings make sense only on a global level, as they are used to configure admin panel functionality, so `showInWebsite` and `showInStore` should be `0` for all of them

AdobeStockImageAdminUi/Model/Block/Adminhtml/System/Config/TestConnection.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->addData(
+            [
+                'button_label' => __($originalData['button_label']),
+                'html_id'      => $element->getHtmlId(),
+                'headers'      =>
+                    $this->json->serialize(
+                        [
+                            'x-api-key'                   => $this->config->getApiKey(),
+                            'X-Product'                   => $this->config->getProductName(),
+                            'Access-Control-Allow-Origin' => '*',
```
sivaschenko: I don't think this header is necessary on the client side

AdobeStockAsset/etc/db_schema.xml
```diff
@@ -0,0 +1,109 @@
+<?xml version="1.0"?>
```
sivaschenko: let's remove the db_schema from this pull request as it will be delivered with https://github.com/magento/adobe-stock-integration/pull/50

sivaschenko (changes requested):
Hi @diazwatson , thanks for the pull request!

Please see my review comments

AdobeStockImageAdminUi/Model/Block/Adminhtml/System/Config/TestConnection.php
diazwatson: After discussed a different approach with @sidolov I need to refactor all this to use the recently introduced request builder instead of a hard coded url to validate the api key

AdobeStockAsset/etc/db_schema.xml
diazwatson: ups

AdobeStockImageAdminUi/etc/adminhtml/system.xml
diazwatson: make sense

AdobeStockImageAdminUi/etc/adminhtml/system.xml
diazwatson: You got me, I was doing copy&paste  ¯\\_ (ツ)_/¯

AdobeStockImage/Model/GetImageList.php
```diff
@@ -58,6 +58,7 @@ public function __construct(
     }
 
     /**
+     * @api
```
sivaschenko: Let's keep only the interface API, the implementation should not be an API. What do you think? cc @sidolov

AdobeStockImageAdminUi/Controller/Adminhtml/System/Config/TestConnection.php
```diff
@@ -0,0 +1,119 @@
...
+use Magento\Framework\Filter\StripTags;
+
+class TestConnection extends Action
+{
+    /**
+     * Authorization level of a basic admin session.
+     *
+     * @see _isAllowed()
+     */
+    const ADMIN_RESOURCE = 'Magento_AdobeStockImageAdminUi::adobe_stock_integration';
```
sivaschenko: I would suggest to create only one ACL resource for now. To match other configuration specific ACLs - it's would be good to configure it the following way:

        <resources>
            <resource id="Magento_Backend::admin">
                <resource id="Magento_Backend::stores">
                    <resource id="Magento_Backend::stores_settings">
                        <resource id="Magento_Config::config">
                            <resource id="Magento_AdobeStockAsset::config" title="Adobe Stock Section" translate="title" />
                        </resource>
                    </resource>
                </resource>
            </resource>
        </resources>

sivaschenko (changes requested):
Thanks for the updates @diazwatson ! Please see my review comments, also I think that all the code in this pull request should be moved to AdobeStockAsset module as it is general configuration for the integration and is not specifically related to Image. @sidolov can you please confirm my opinion?

AdobeStockImageAdminUi/Controller/Adminhtml/System/Config/TestConnection.php
```diff
@@ -0,0 +1,119 @@
...
+     */
+    public function execute()
+    {
+        $result = [
+            'success'      => false,
+            'errorMessage' => '',
+        ];
+
+        try {
+            $response = $this->getImageList->execute($this->getSearchCriteria());
```
sivaschenko: Considering that the configuration is general and not specifically related to image I would avoid dependency on AdobeStockImage module. For this purpose, I propose to introduce a `getAssetList` or `testConnection` API to AdobeStockAssetApi and use it

diazwatson: Hey guys, I have applied the requested changes. Could you please review when you get a chance?

**Note:** I am not very happy with the implementation of `testConnection` method, can you see a better way to implement it?

sivaschenko: Thanks for the contribution, @diazwatson ! I think test connection implementation is perfect for now, also, we can improve it later if there will be ideas.

The only adjustments I would do now - is changing the return type to bool

sivaschenko (approved):

AdobeStockImageAdminUi/Controller/Adminhtml/System/Config/TestConnection.php
diazwatson: Thanks @sivaschenko I'll change that tonight

sivaschenko: @diazwatson actually, I did this already. Pull request is merged! Please take a look at my changes and feel free to create a new pull request if you'd like to adjust anything

PR #53, opened 2019-05-15
Introduced request builder

sidolov:
- Removed redundant block
- Added strict types to existing methods
- Introduced request builder


AdobeStockAsset/Model/Client.php
```diff
@@ -32,32 +32,35 @@ public function __construct(ConfigInterface $config)
...
      */
-    public function search()
+    public function execute(\Magento\AdobeStockAssetApi\Api\Data\RequestInterface $request)
     {
-        // TODO THIS IS A STUB
-        $results_columns = Constants::getResultColumns();
-        $search_params = new SearchParameters();
-        $search_params->setWords('tree')->setLimit(32)->setOffset(0);
+        // TODO: THIS IS A STUB. SHOULD BE REFACTORED
+        $words = $request->getData('filters')['words'] ?? 'image';
```
sivaschenko, line 42: Is "image" word the best query to use by default? Can we perform a request without words filter at all?

AdobeStockAsset/Model/Client.php
```diff
@@ -32,32 +32,35 @@ public function __construct(ConfigInterface $config)
...
+        $searchParams->setOffset($request->getData('offset'));
+
+        $resultsColumns = Constants::getResultColumns();
+        $resultColumnArray = [];
+        foreach ($request->getData('resultColumns') as $column) {
+            $resultColumnArray[] = $resultsColumns[$column['field']];
+        }
 
         $request = new SearchFilesRequest();
         $request->setLocale('En_US');
```
sivaschenko, line 58: I think locale should be passed through the `RequestInterface`. (let's add a TODO if we'd like to extract it later)

AdobeStockAsset/etc/request.xsd
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
```
sivaschenko: Can we use DI configuration (arguments) to avoid introducing a new schema? What are the benefits of custom schema comparing to the DI configuration approach?

AdobeStockAsset/Model/Request/Request.php
```diff
@@ -0,0 +1,33 @@
...
+    public function __construct(array $data)
+    {
+        $this->data = $data;
+    }
+
+    /**
+     * @param string $key
+     * @return mixed|null
+     */
+    public function getData(string $key)
```
sivaschenko: Are we going to use concrete getters later?

AdobeStockAsset/Model/Request/Builder.php
```diff
@@ -0,0 +1,127 @@
...
+     */
+    public function setSort(array $sort): void
+    {
+        $this->data['sort'] = $sort;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function bind(string $placeholder, $value): void
```
sivaschenko: `$placeholder` is a bit confusing name for this parameter (considering that it's not a placeholder with the current implementation), I'd rename to `$name` or  `$key`

AdobeStockAsset/Model/Request/Builder.php
```diff
@@ -0,0 +1,127 @@
...
+        }
+        $requestName = $this->data['requestName'];
+
+        /** @var array $requestConfig */
+        $requestConfig = $this->config->getRequestConfig($requestName);
+        if ($requestConfig === null) {
+            throw new NonExistingRequestNameException(new Phrase("Request name '%1' doesn't exist.", [$requestName]));
+        }
+
+        $requestConfig = $this->binder->bind($requestConfig, $this->data);
```
sivaschenko, line 114: `binder->bind` looks a bit confusing. I'd adjust the naming to i.e. `dataBuilder->getRequestData` (Having the class name `Magento\AdobeStockAsset\Model\Request\Builder\Data`)

AdobeStockAsset/Model/Request/Builder.php
```diff
@@ -0,0 +1,127 @@
...
+        return $this->convert($requestConfig);
+    }
+
+    /**
+     * @param array $data
+     * @return \Magento\AdobeStockAssetApi\Api\Data\RequestInterface
+     */
+    private function convert(array $data) : \Magento\AdobeStockAssetApi\Api\Data\RequestInterface
+    {
+        // TODO: Define request interface and refactor line below.
```
sivaschenko, line 125: I don't think this private method is needed. Also, there should be no setters defined in the `RequestInterface` in my opinion

AdobeStockAsset/Model/Request/Builder/Binder.php
```diff
@@ -0,0 +1,41 @@
...
+     * @param array $requestData
+     * @return array
+     */
+    public function bind(array $requestConfig, array $requestData) : array
+    {
+        return [
+            'size' => (int) $requestData['size'] ?? (int) $requestConfig['size'],
+            'offset' => (int) $requestData['offset'] ?? (int) $requestConfig['from'],
+            'filters' => $this->buildFilters($requestData['placeholder'] ?? [] , $requestConfig['filters']),
+            'resultColumns' => $requestData['resultColumns'] ?? $requestConfig['resultColumns'],
```
sivaschenko, line 22: I don't see a way to set `resultColumns` to builder's data

AdobeStockAsset/Model/Request/Builder.php
```diff
@@ -0,0 +1,127 @@
...
+    {
+        $this->data['size'] = $size;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function setOffset(int $offset): void
+    {
+        $this->data['offset'] = $offset;
```
sivaschenko, line 79: Considering that these keys are used in builder and binder classes and will be used in request class, we should declare corresponding constants in the RequestInterface

sivaschenko (changes requested):
@sidolov nice one!

Please see my review comments

AdobeStockAsset/Model/Client.php
sidolov: We should define the search word because this field is required for request

AdobeStockAsset/Model/Request/Builder.php
sidolov: Fixed

AdobeStockAsset/Model/Request/Builder.php
sidolov: I totally agree with you, setter added as stub right now. We will define the request interface soon

AdobeStockAsset/Model/Request/Builder.php
sidolov: Added todo to the builder class

Issue #7, opened 2019-05-01
Admin user can search images by keywords


sidolov:
### Description
- Add a search field to the image listing
- Integrate search by keywords to the request builder
- Cover changes with MFTF test
- Add locale support to request (locale should be the same as on Magento admin)

### Documentation
Story: [User searches stock images by keywords](https://github.com/magento/adobe-stock-integration/issues/23)
App Design Document: [Adobe Stock Image Search Design](https://github.com/magento/adobe-stock-integration/wiki/Adobe-Stock-Image-Search-Design)

### QA notes
- Admin has the ability to search images by keywords

PR #59, opened 2019-05-24
Localize the image search by keywords for admin

valentinboyanov:
### Description (*)

Add Magento admin locale to search API requests so admin user can see localized results.

1. magento/adobe-stock-integration#7 Admin user can search images by keywords

1. Change locale configuration for 'Default config': Stores > Configuration > General > Locale options > Locale
2. Search stock images by keywords in admin (UI Component)

### DIFF
--- a/AdobeStockAsset/Model/Client.php
+++ b/AdobeStockAsset/Model/Client.php
@@ -11,7 +11,6 @@ use AdobeStock\Api\Models\SearchParameters;
 use AdobeStock\Api\Request\SearchFiles as SearchFilesRequest;
 use Magento\AdobeStockAssetApi\Api\ClientInterface;
 use Magento\AdobeStockAssetApi\Api\Data\ConfigInterface;
-use Magento\Framework\Locale\ResolverInterface;
 
 /**
  * DataProvider for cms ui.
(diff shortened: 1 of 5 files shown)


sivaschenko (changes requested):
Hi @valentinboyanov , thanks for the pull request!
 
To be honest, I think that `Client` class is too low level and too specific to interact with a locale resolver.

The locale should be retrieved from the `RequestInterface`.

Could you update the changes and set the locale in `\Magento\AdobeStockImage\Model\GetImageList::execute` to request builder and then retrieve it frorm the the `RequestInterface` in `Client`

AdobeStockAsset/Model/Client.php
```diff
@@ -92,4 +100,9 @@ private function getClient()
             $this->config->getTargetEnvironment()
         );
     }
+
+    private function locale(): string
```
sidolov: Is there a reason to use a private method instead of direct usage of the dependency?

sidolov (changes requested):

AdobeStockAsset/Model/Client.php
valentinboyanov: First I extracted the locale value to a method in order to express the concept. And then I chose the LocaleResolver to provide the expected value. But now I think that changing the name of the collaborator I will achieve the same avoiding one level of indirection :)

sidolov (approved):

PR #60, opened 2019-05-27
Add Backend locale for requests

diazwatson:


sivaschenko:
Hi @diazwatson looks like it's a dup to https://github.com/magento/adobe-stock-integration/pull/59

diazwatson:
It is indeed

PR #87, opened 2019-06-13
Extracted client modules

sivaschenko:
### Description (*)

The code responsible for interaction with the SDK is moved to a separate module.
Refactored the approach to fill SearchParameters in a flexible and extensible way.


AdobeStockClient/Model/Client.php
```diff
@@ -0,0 +1,201 @@
...
+
+        foreach ($itemData as $key => $value) {
+            if ($value === null) {
+                continue;
+            }
+            $attribute = $this->attributeValueFactory->create();
+            $attribute->setAttributeCode($key);
+            if (is_bool($value)) {
+                // for proper work of form and grid (for example for Yes/No properties)
+                $value = (string)(int)$value;
```
YevSent, line 149: It's better to use `boolvar()`.

AdobeStockClient/Model/Client.php
```diff
@@ -0,0 +1,201 @@
...
+            $this->config->getApiKey(),
+            $this->config->getProductName(),
+            $this->config->getTargetEnvironment()
+        );
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function testConnection()
```
YevSent, line 182: What the purpose of this method?

AdobeStockClient/Model/Client.php
```diff
@@ -0,0 +1,201 @@
...
+            $attribute->setValue($value);
+            $attributes[$key] = $attribute;
+        }
+        return $attributes;
+    }
+
+    /**
+     * @return null
+     */
+    private function getAccessToken()
```
YevSent, line 162: Does it make sense to have an additional method to return `null`? At least, the method should contain a description of why access token is not needed for a search.

AdobeStockClient/Model/SearchParametersProvider/Pagination.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Magento\AdobeStockClient\Model\SearchParametersProvider;
+
+use AdobeStock\Api\Models\SearchParameters;
+use Magento\AdobeStockClientApi\Api\SearchParameterProviderInterface;
+use Magento\Framework\Api\SearchCriteriaInterface;
+
+class Pagination implements SearchParameterProviderInterface
+{
+    /**
+     * @inheritDoc
```
YevSent: `inheritdoc` should be used instead.

AdobeStockClient/Model/SearchParametersProviderPool.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Magento\AdobeStockClient\Model;
+
+use AdobeStock\Api\Models\SearchParameters;
+use Magento\AdobeStockClientApi\Api\SearchParameterProviderInterface;
+use Magento\Framework\Api\SearchCriteriaInterface;
+
+class SearchParametersProviderPool implements SearchParameterProviderInterface
```
YevSent: This class rather `Composite` than `Pool`.

AdobeStockClient/composer.json
```diff
@@ -0,0 +1,29 @@
...
+    "license": [
+        "OSL-3.0",
+        "AFL-3.0"
+    ],
+    "autoload": {
+        "files": [
+            "registration.php"
+        ],
+        "psr-4": {
+            "Magento\\AdobeStockClinet\\": ""
```
YevSent: A type, `AdobeStockClient` should be instead.

AdobeStockClientApi/composer.json
```diff
@@ -0,0 +1,21 @@
...
+    "license": [
+        "OSL-3.0",
+        "AFL-3.0"
+    ],
+    "autoload": {
+        "files": [
+            "registration.php"
+        ],
+        "psr-4": {
+            "Magento\\AdobeStockClinetApi\\": ""
```
YevSent: A typo, `AdobeStockClientApi` should be instead.

AdobeStockClient/Model/SearchParametersProvider/Premium.php
```diff
@@ -0,0 +1,26 @@
...
+use Magento\Framework\Api\SearchCriteriaInterface;
+
+class Premium implements SearchParameterProviderInterface
+{
+    /**
+     * @inheritDoc
+     */
+    public function apply(SearchCriteriaInterface $searchCriteria, SearchParameters $searchParams): SearchParameters
+    {
+        foreach ($searchCriteria->getFilterGroups() as $filterGroup) {
```
YevSent: I would propose to extract all blocks of
```php
foreach ($searchCriteria->getFilterGroups() as $filterGroup) {
            foreach ($filterGroup->getFilters() as $filter) {
                if ($filter->getField() === 'premium_price_filter') {
                    $searchParams->setFilterPremium($filter->getValue());
                    return $searchParams;
                }
            }
        }
```

To a separate class as there are similar and the difference only in field name and search parameter setter method's name.

AdobeStockClient/registration.php
```diff
@@ -0,0 +1,11 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+\Magento\Framework\Component\ComponentRegistrar::register(
```
YevSent: Please, avoid FQCN usage.

AdobeStockClient/Model/SearchParametersProviderPool.php
sivaschenko: I called it pool to clarify how this class should be used. It's not expected to have multiple levels of subsequent classes for applying criteria to search parameters.

In order to avoid confusion, I'd keep it a `Pool`

AdobeStockClient/Model/SearchParametersProvider/Pagination.php
sivaschenko: This annotation tag was generated by PHPStorm in camel case. Also, guides suggest using camel case for this annotation: https://docs.phpdoc.org/guides/inheritance.html#the-inheritdoc-tag

AdobeStockClient/Model/Client.php
sivaschenko: That's a placeholder for future implementation, corresponding phpdoc added

AdobeStockClient/Model/Client.php
sivaschenko: This method is intended to verify the connection to Adobe Stock API, currently used in system configuration for `Test Connection` button, see the phpdoc in `ClientInterface`

AdobeStockClient/Model/Client.php
sivaschenko: The goal is to have a string type variable, not boolean

AdobeStockClient/composer.json
sivaschenko: Thanks! :)

AdobeStockClientApi/composer.json
sivaschenko: Thanks! :)

AdobeStockClient/Model/SearchParametersProvider/Premium.php
sivaschenko: Done

AdobeStockClient/registration.php
sivaschenko: Done

sivaschenko:
@joni-jones @sidolov thanks for the review! Requested changes are done, please review

AdobeStockClient/Model/SearchParametersProvider/Pagination.php
YevSent: It depends on PHPStorm configuration, in most cases it generates `inheritdoc`. Also, in most cases, Magento uses `inheritdoc` but not `inheritDoc` (~400 vs ~5000 usages).

phpDocumentor describes the usage of `{@inheritDoc}` but not `@inheritDoc`.

AdobeStockClient/Model/SearchParametersProviderPool.php
YevSent: But its usage shows it as composite because you call the interface and configure the preference via di.xml as composite, not as a pool. As for me, the `Pool` more confusing name rather than `Composite`, what signs of the pool this class has?

sidolov (approved):

Issue #18, opened 2019-05-01
Implement search filter for the "colors" attribute


sidolov:
### Description
- Implement "colors" search filter UI element
- Implement "colors" filter for the API endpoint on the backend

### Documentation
Story: [User filters images by image color](https://github.com/magento/adobe-stock-integration/issues/37)
App Design Document: [Adobe Stock Image Search Design](https://github.com/magento/adobe-stock-integration/wiki/Adobe-Stock-Image-Search-Design)

### QA notes
- User has the ability to filter results by "colors" attribute

PR #99, opened 2019-06-14
#18 Implement search filter for the color attribute

diazwatson:
### Description (*)

#18 Although it seems to be working there are two things I am not happy about it:
Screenshot 2019-06-14 at 01 51 16
Screenshot 2019-06-14 at 01 51 33

- Rules defined in `Magento_Ui:etc/ui_configuration.xsd` do not allow a colorPicker component to be added inside a listing component (should we extend `ui_configuration.xsd` to include this type of configuration?).
- As you can see, the label of the filter is aligned to the right, this is due to the class automatically added to the component (`admin__field-label`). the right class should be `admin__form-field-label`

I am going to continue reviewing this to see if I can fix the issues or see if I can come up with a better  way to implement this filter.


diazwatson:
Hey Guys, I have refactored to use the template component class approach instead of the color picker but for some reason the label doesn't show. Has anyone seen similar issue with a Ui Component  before?

AdobeStockImageAdminUi/Ui/Component/Listing/Filter/Color.php
```diff
@@ -0,0 +1,123 @@
+<?php
```
larsroettig, line 1: Please Use **strict_type** 

```php
<?php
/**
 *
 * Copyright © Magento, Inc. All rights reserved.
 * See COPYING.txt for license details.
 */
declare(strict_types=1);
```

AdobeStockImageAdminUi/Ui/Component/Listing/Filter/Color.php
```diff
@@ -0,0 +1,123 @@
...
+use Magento\Framework\View\Element\UiComponent\ContextInterface;
+use Magento\Framework\View\Element\UiComponentFactory;
+use Magento\Ui\Component\Filters\FilterModifier;
+use Magento\Ui\Component\Form\Element\Input as ElementInput;
+use Magento\Ui\Component\Filters\Type\AbstractFilter;
+use Magento\Ui\Model\ColorPicker\ColorModesProvider;
+
+class Color extends AbstractFilter
+{
+    const NAME = 'filter_input';
```
larsroettig, line 28: why do you add their const for one-time usage? Usually, we have const only interfaces`

AdobeStockImageAdminUi/Ui/Component/Listing/Filter/Color.php
```diff
@@ -0,0 +1,123 @@
...
+
+    const COMPONENT = 'input';
+
+    /**
+     * Wrapped component
+     *
+     * @var ElementInput
+     */
+    protected $wrappedComponent;
```
larsroettig: OPTIONAL: you should use private there to enforce setter and getter

AdobeStockImageAdminUi/Ui/Component/Listing/Filter/Color.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Provides color picker modes configuration
+     *
+     * @var ColorModesProvider
+     */
+    private $modesProvider;
+
+    public function __construct(
```
larsroettig, line 54: Missing doc block static should fail

AdobeStockImageAdminUi/Ui/Component/Listing/Filter/Color.php
```diff
@@ -0,0 +1,123 @@
...
+        );
+
+        $this->applyFilter();
+
+        $this->initColorPickerConfig();
+
+        parent::prepare();
+    }
+
+    private function initColorPickerConfig()
```
larsroettig: Missing docblock what do

larsroettig (changes requested):

diazwatson:
Hi @larsroettig 
I believe all changes you requested are now covered, could you please review and let me know if any?

Cheers

sivaschenko (approved):

larsroettig (approved):

sidolov (approved):

Issue #84, opened 2019-06-13
Uncaught Error: You cannot apply bindings multiple times to the same element.


TomashKhamlai:
### Preconditions:
1. php bin/magento config:set adobe_stock/integration/enabled 1
2. php bin/magento config:set adobe_stock/integration/api_key \<key\>
3. php bin/magento config:set cms/wysiwyg/enabled disabled
4. php bin/magento config:set admin/security/admin_account_sharing 1
5. php bin/magento config:set admin/security/use_form_key 0
6. php bin/magento cache:flush

### Steps to reproduce:
1. Navigate to Content > Elements > Pages
2. Click Add New Page button
3. Fill Title with "Test CMS Page"
4. Expand the "Content" section
5. Fill Content Heading with "Test Content Heading"
6. Click "Insert Image..." button
7. Click "Cancel"
8. Click "Insert Image..." button

### Expected result:
:heavy_check_mark: JS works as expected

### Actual result
:x: JS error in knockout.js::3287
```js
Uncaught Error: You cannot apply bindings multiple times to the same element.
    at applyBindingsToNodeInternal (knockout.js:3287)
    at applyBindingsToNodeAndDescendantsInternal (knockout.js:3233)
    at Object.ko.applyBindings (knockout.js:3444)
    at _super (wrapper.js:73)
    at Object.applyBindings (bound-nodes.js:120)
    at Object.applyBindings (wrapper.js:78)
    at bindings.js:84
    at Array.forEach (<anonymous>)
    at jQuery.fn.init.$.fn.applyBindings (bindings.js:81)
    at panel.js:18
```
  | applyBindingsToNodeInternal | @ | knockout.js:3287
-- | -- | -- | --
  | applyBindingsToNodeAndDescendantsInternal | @ | knockout.js:3233
  | ko.applyBindings | @ | knockout.js:3444
  | _super | @ | wrapper.js:73
  | applyBindings | @ | bound-nodes.js:120
  | (anonymous) | @ | wrapper.js:78
  | (anonymous) | @ | bindings.js:84
  | $.fn.applyBindings | @ | bindings.js:81
  | (anonymous) | @ | panel.js:18
  | (anonymous) | @ | main.js:31
  | execCb | @ | require.js:1650
  | context.execCb | @ | resolver.js:145
  | check | @ | require.js:866
  | enable | @ | require.js:1143
  | init | @ | require.js:774
  | (anonymous) | @ | require.js:1416
[...]

PR #104, opened 2019-06-14
fixed Uncaught Error: You cannot apply bindings multiple times to the…

konarshankar07:
… same element.

### Description (*)
As this bug was related to KnockOut JS and they are working to fix this issue so I've tweaked some code  to display modal without getting the error

1. magento/adobe-stock-integration#84: Uncaught Error: You cannot apply bindings multiple times to the same element. 

1. Navigate to Content > Elements > Pages
2. Click Add New Page button
3. Fill Title with "Test CMS Page"
4. Expand the "Content" section
5. Fill Content Heading with "Test Content Heading"
6. Click "Insert Image..." button
7. Click "Cancel"
8. Click "Insert Image..." button
9. There will be no error in KnockOut Js in the browser console


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/adobe-stock-integration?pullRequest=104) <br/>All committers have signed the CLA.

sidolov (changes requested):
@konarshankar07 , unfortunately, we can't accept such changes, all inline JS code should be removed from the templates and moved to the separate file like it was before the changes. 
Mentioned bug reproducing because during the media gallery slide panel loading were performed two requests to the server instead of one. That's the issue that should be fixed and delivered to the magento2 repository instead of Adobe Stock integration.

konarshankar07:
@sidolov 
I see your point but I took the same logic applied in the Magento core files(app/code/Magento/ConfigurableProduct/view/adminhtml/templates/catalog/product/edit/super/wizard.phtml) and also make sure your it won't happen again from now onwards.

Thanks

konarshankar07:
@sidolov 
I think the problem is the applyBindings() function invokes each time when we open the modal and we get the error "You cannot apply bindings multiple times to the same elements". For the first time, The modal content is already added and then after closing, we are not removing the modal content so we getting the error when we try to open the modal again so applyBindings() is called for the second time and element is already present so this way we are getting the error.

Thanks

AdobeStockImageAdminUi/view/adminhtml/templates/panel.phtml
```diff
@@ -5,16 +5,17 @@
...
 </div>
 
-<script type="text/x-magento-init">
-    {
-        "#adobe-stock-images-search-modal": {
-            "Magento_AdobeStockImageAdminUi/panel": {}
-        }
-    }
+<script>
+    require(['jquery', 'mage/apply/main'], function ($, main) {
```
sivaschenko: I'd avoid inline scripts if it's possible.

sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request! I'd avoid inline script if it's possible.

Also, I think that this particular bug might need to be fixed in gallery functionality in magento/magento2 repository: ensuring that the script is triggered and executed only once.

Please let me know what do you think

AdobeStockImageAdminUi/view/adminhtml/templates/panel.phtml
konarshankar07: Yes, I agree with you and @sidolov  regarding this pull request we need to avoid inline js and I don't think that this issue is with Magento 2 as gallery functionality will work without error in magento/magento2 repository. I think applyBindings() is triggered every time we open our Adobe stock modal so we need to fix this. I'm working on an alternative way with using the knockout js template to render the sliding modal. Let me know you have any other idea to fix this issue.

Thanks

sidolov: @konarshankar07 , we should fix gallery and make only one request when modal will be opened first time and reuse already loaded content in all cases, including the case when modal was closed and opened again.

konarshankar07: Yes, I will try to apply your logic. But can we implement different without using the following code? I mean are there any other ways we can add our Uicomponent in the modal content.
https://github.com/magento/adobe-stock-integration/blob/7556d544db268309817c32e9cdc1ee8f98c9246b/AdobeStockImageAdminUi/view/adminhtml/web/panel.js#L18
```
return function (config, element) {
        $(element).modal({
            type: 'slide',
            buttons: [],
            title: $t('Adobe Stock')
        }).applyBindings();
    };
```

sidolov: @konarshankar07 I'm not sure that we have another valid way to add it, at least this one is more easiest. And we still have a bug when content loading more than once after modal init

konarshankar07:
Hello @sidolov and @sivaschenko ,

I managed to found another way to use applyBindings without js error. Please check and let me know if you still want to make any changes. I might have an issue with coding standards. Please comment regarding my coding standards and also help me out how can I improve it so it won't happen next time.

Thanks

sidolov (changes requested):
Please, don't try to resolve this issue on the Adobe Stock integration side. This issue should be fixed in the Magento open source core code by second request elimination.

konarshankar07:
@sidolov 

Honestly, I don't think its from the Magento Open source core code cause you can check this [comment](https://github.com/magento/adobe-stock-integration/issues/84#issuecomment-501730385) as it is not reproduced in the core magneto without extension. As I've checked and found that our previous Adobe Stock button implementation is creating the second request binding. You can also check from here
```
Uncaught Error: You cannot apply bindings multiple times to the same element.
    at applyBindingsToNodeInternal (knockout.js:3287)
    at applyBindingsToNodeAndDescendantsInternal (knockout.js:3233)
    at Object.ko.applyBindings (knockout.js:3444)
    at _super (wrapper.js:73)
    at Object.applyBindings (bound-nodes.js:120)
    at Object.applyBindings (wrapper.js:78)
    at bindings.js:84
    at Array.forEach (<anonymous>)
    at jQuery.fn.init.$.fn.applyBindings (bindings.js:81)
    at panel.js:18
```
As you can see the last line says that the error is from the panel.js so I don't think that its causing due to Magento Open source core code.  

Thanks

sivaschenko:
@konarshankar07 I believe the problem with reproducing it on vanilla Magento is that there is no indicator (no code) to outline the issue. It would be good to update the issue steps to reproduce to include the code snippet that should be executed only once.

sidolov:
Hi @konarshankar07 , I have fixed the issue on the Magento side. You may take a look at fix: https://github.com/magento/magento2/pull/23621

konarshankar07:
Hello @sidolov ,

Awesome job!!! I found the same issue with other buttons like"Insert widget" and "Insert Variable". Do you want me to fix that issue?

Thanks

sidolov:
Hi @konarshankar07 , sure, you are welcome to fix mentioned bugs in the magento2 repository! Thank you!

sivaschenko:
Thanks for collaboration! Closing this pull request as the issue was fixed in the core repository.

PR #123, opened 2019-06-22
FEATURE: add exception manager for adobe stock integration client

coderimus:
### Description
For more isolated exception operation an exception manager class is added. The goal of it in the more detail exception management which is mostly independent of the integration library and provide functionality for making exception handling more informative.

### Fixed Issues
N/A

### Manual testing scenarios
N/A


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/adobe-stock-integration?pullRequest=123) <br/>All committers have signed the CLA.

AdobeStockClientApi/Api/ExceptionManagerInterface.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ *
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\AdobeStockClientApi\Api;
+
+interface ExceptionManagerInterface
```
larsroettig: Missing doc `api interfaces` must be documented it should be also an `@api` annotations there

larsroettig (changes requested):
Please fixe note from than looks good to me the only `const` is not so often used in Magento currently if we are adding const we must maintain them for that.  @sidolov What do you think to add const for that

AdobeStockClient/Model/ExceptionManager.php
```diff
@@ -0,0 +1,90 @@
...
+            default:
+                $this->buildDefaultErrorMessage();
+        }
+    }
+
+    private function buildForbiddenConnectionMessage()
+    {
+        $url = $this->urlBuilder->getUrl('adminhtml/system_config/edit/section/system');
+        $this->errorMessage = sprintf(
+            'Adobe Stock API not configured. Please, proceed to <a href="%s">Configuration → System → Adobe Stock Integration.</a>',
```
sidolov: The client shouldn't know about admin UI and how to set up the API key. 
BTW, we may have the only API for Adobe Stock Integration and disable AdminUi module at all.

AdobeStockClient/Model/Client.php
```diff
@@ -124,6 +119,39 @@ public function search(SearchCriteriaInterface $searchCriteria): SearchResultInt
...
+            $searchRequest = new SearchFilesRequest();
+            $searchRequest->setLocale($this->localeResolver->getLocale());
+            $searchRequest->setSearchParams($searchParams);
+            $searchRequest->setResultColumns($resultColumnArray);
+
+            $client = $this->getClient()->searchFilesInitialize($searchRequest, $this->getAccessToken());
+            $response = $client->getNextResponse();
+
+            return $response;
+        } catch (\Exception $exception) {
```
sidolov, line 149: We need to handle `\AdobeStock\Api\Exception\StockApi` exceptions first.

AdobeStockClient/Model/Client.php
```diff
@@ -124,6 +119,39 @@ public function search(SearchCriteriaInterface $searchCriteria): SearchResultInt
...
+            $this->exceptionManager->processException(
+                $exception,
+                ExceptionManagerInterface::FORBIDDEN_CONNECTION_ERROR_CODE
```
sidolov: Why are we trying to handle all possible problems in the code as a Forbidden error? In this case, we can just throw Localized exception without any managers

sidolov (changes requested):
Could you please provide benefits of ExceptionManager? The current implementation looks like we need just verify the exception from the service and throw `LocalizedException` with a corresponded message.

coderimus:
> Could you please provide benefits of ExceptionManager? The current implementation looks like we need just verify the exception from the service and throw `LocalizedException` with a corresponded message.

@sidolov  The main inspiration to make this manager was an idea to have a service which can manage exception happened in the client and make them as much information as possible according to [those standards](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html#5-exceptions)

With the Exception manager, we also have a layer between the API library and Magento 2 module which allows as to be independent of their implementation which currently has not so informative exceptions according to Magento needs.

AdobeStockClient/Model/ExceptionManager.php
coderimus: Yes, thank you for note. I changed the way this method handles what test should be shown.

AdobeStockClient/Model/Client.php
coderimus: I extended this part: we handle AdobeStock exception first, then Exception and if something extraordinary happened just throw a default exception.

sidolov:
@coderimus , I got your point. Sure, we need to handle exceptions properly, but ExceptionManager sounds a little bit weird. Exceptions it's a basic mechanism and we don't want to make simple things complicated. 

Here are the steps of how we can achieve proper exceptions handling without additional classes and interfaces: 

1. First of all, we need to handle exceptions thrown by SDK in our client (I see that's already done in the latest changes) 
2. Throw our own Magento exception from the Client with the correct message (for example throw `\Magento\Framework\Exception\AuthenticationException` if the error caused by bad API key). We don't need to throw LocalizedException on that level, we need to show as much info as we can. 
3. Handle exceptions thrown by the client in the presentation layer and throw LocalizedException to the end user with a generic message (`\Magento\AdobeStockImage\Model\GetImageList`)

By this approach, we will eliminate additional logic for exception handling and make it more simple and transparent.

coderimus:
Hi! @sivaschenko @sidolov 
I changed the way how exceptions are handling in the `Client` class and also covered `GetImgeList` execute method with the localized exception. Please, review.

sidolov (approved):

Issue #116, opened 2019-06-21
Configuration can be saved with not valid API key


TomashKhamlai:
Related to #29

### Preconditions:
1. Go to **Steps to reproduce** section

### Steps to reproduce:
1. Execute magentoCLI command
`php bin/magento config:set adobe_stock/integration/api_key ""`
2. Execute magentoCLI command
`php bin/magento config:set adobe_stock/integration/enabled 0`
3. Before - LoginAsAdmin
4. User navigates to Stores > Settings > Configuration
5. Page reloads
6. User expands `Advanced` tab
7. User clicks on `System` section
8. Page reloads
9. Location is `admin/system_config/edit/section/system/`
10. User expands "Adobe Stock Integration" expandable section
11. User sets "Enabled Adobe Stock" to "Yes"
12. Find input field labeled by `Api Key` label
13. Input field is in empty state
14. User fills next string to the `API Key` input
`adobeadobeadobeadobeadobeadobe`
15. User clicks `Test Connection` button
16. `Key is invalid. Try a different key.` appears
17. User clicks `Connection failed! Test again?` button
18. `Key is invalid. Try a different key.` appears
19. User clicks `Save Config` button

### Expected result:
:heavy_check_mark: 1. Configuration is successfully saved

### Actual result
:x: 1. `Key is invalid. Try a different key.` appears

sivaschenko:
I don't think thats an issue. @tkacheva I think we should allow saving API key without verification (i.e. to be able to workaround missing network connection to Adobe Stock). However, the ability to test the API key before saving the configuration will be introduced in https://github.com/magento/adobe-stock-integration/pull/126

What do you think about it?

PR #125, opened 2019-06-23
#116 : Fixed issue for Configuration can be saved without valid API key  

konarshankar07:

1. magento/adobe-stock-integration#116: Configuration can be saved with not a valid API key

1. Execute magentoCLI command
2. php bin/magento config:set adobe_stock/integration/api_key ""
3. Execute magentoCLI command
4. php bin/magento config:set adobe_stock/integration/enabled 0
5. Before - LoginAsAdmin
6. User navigates to Stores > Settings > Configuration
7. Page reloads
8. User expands Advanced tab
9. User clicks on System section
10. Page reloads
11. Location is admin/system_config/edit/section/system/
12. User expands "Adobe Stock Integration" expandable section
13. User sets "Enabled Adobe Stock" to "Yes"
14. Find input field labeled by Api Key label
15. Input field is in empty state
16. User fills next string to the API Key input adobeadobeadobeadobeadobeadobe
17. User clicks Test Connection button Key is invalid. Try a different key. appears
18. User clicks Connection failed! Test again? button Key is invalid. Try a different key. appears
19. User clicks Save Config button
20. User gets the error message "Key is invalid. Try a different key" and value is not saved


AdobeStockAsset/Model/Config/Backend/AdobeStockApi.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public function beforeBeforeSave(\Magento\Config\Model\Config\Backend\Encrypted $subject)
+    {
+        $value = $subject->getValue();
+        $this->testConnection($value);
+    }
+    /**
+     * @inheritdoc
+     */
+    public function testConnection($value)
```
sidolov: We already have `testConnection` method here: `\Magento\AdobeStockClient\Model\Client::testConnection`

AdobeStockAsset/etc/di.xml
```diff
@@ -10,5 +10,8 @@
     <preference for="Magento\AdobeStockAssetApi\Api\Data\AssetInterface" type="Magento\AdobeStockAsset\Model\Asset"/>
     <preference for="Magento\AdobeStockAssetApi\Api\Data\KeywordInterface" type="Magento\AdobeStockAsset\Model\Keyword"/>
     <preference for="Magento\AdobeStockAssetApi\Api\Data\AssetSearchResultsInterface" type="Magento\AdobeStockAsset\Model\SearchResults" />
+    <type name="Magento\Config\Model\Config\Backend\Encrypted">
```
sidolov: Implemented plugin will be called for every field in Magento that has `Magento\Config\Model\Config\Backend\Encrypted` backend model. I'm not sure that we want to run suck check on all encrypted fields. It's incorrect behavior.

sidolov (changes requested):

AdobeStockAsset/etc/di.xml
konarshankar07: Hello @sidolov ,

Yes, you are right. I'm so stupid to use plugin here. I will update this code.

AdobeStockAsset/Model/Config/Backend/AdobeStockApi.php
konarshankar07: Hello @sidolov 

We need to refactor this function if we want to use because of this function we can test after system configuration value has been saved but as per this task requirement, we need to check before saving the system configuration. You can check `\Magento\AdobeStockClient\Model\Client::getClient` on line 173. Please check let know how you want to implement before saving the system configuration

Thanks

sivaschenko:
Hi @konarshankar07 on the last demo we discussed test connection button and decided that it will be better to implement it so that admin can test the connection before even saving the API key. There is a PR created that implements that functionality: https://github.com/magento/adobe-stock-integration/pull/126

Lets close this PR and collaborate to complete that pull request

konarshankar07:
Hello @sivaschenko ,

I also had the same problem and discussed with @sidolov regarding the same issue that user needs to save the API key to test from the admin and I also worked on the same issue and waiting for the update from @sidolov but I think he is on Adobe summer break. So If you think that we need to close this pull request then it's fine.

Thanks

sivaschenko:
Closing if favour to https://github.com/magento/adobe-stock-integration/pull/126

PR #126, opened 2019-06-24
FEATURE: improve test button notification presentation

coderimus:
### Description
During configuration, I faced with the issue that it is not so obvious what happened after the "Test Connection" is pressed: is it tests an API key which is just typed or the saved one.

As a suggestion, I added next in this PR:
1. Add a note under the button: `Works only for the saved value. Please, save the API key before the test.`
1. Add notification message box under the test button

For the failure API check:
image

For the success API check:
image

### Fixed Issues
N/A

### Manual testing scenarios
Try to save valid and invalid API key


coderimus:
Hi @sidolov @sivaschenko ,
I made adjustments according to our conversation about the provided functionality. Next was done:

1. `testConnection` method now has an optional parameter - `testConnection(AdobeStock $connectionInstance = null)` I added this to give an ability to send a connection instance generated by `ConnectionFactory`.  If this param is `null` config settings will be used. What do you think about this?
1. `Magento\AdobeStockAsset\Model\Config\Backend\Encrypted` was added as a model which handles the `beforeSave` action and validate API key value which is tried to be saved. If it is not valid - do not save it and keep the one which was inserted. 
1. Move js code from phtml to the `web\js` source and rewrite it from prototype to jQuery.

Please, review :)

Thank you,
Alex

coderimus:
@sidolov @sivaschenko static tests fixed.

AdobeStockClientApi/Api/ClientInterface.php
```diff
@@ -27,7 +28,9 @@ public function search(SearchCriteriaInterface $searchCriteria): SearchResultInt
     /**
      * Perform a basic request to Adobe Stock API to check network connection, API key, etc.
      *
+     * @param AdobeStock|null $adobeStock
+     *
      * @return bool
      */
-    public function testConnection(): bool;
+    public function testConnection(AdobeStock $adobeStock = null): bool;
```
sivaschenko: The idea of `ClientInterface` is to separate the SDK implementation from the Magento modules. AdobeStock classes should not be exposed outside the `ClientInterface`.

`testConnection` method may accept the API key as a parameter though.

AdobeStockAsset/Model/Config/Backend/Encrypted.php
```diff
@@ -0,0 +1,152 @@
...
+     */
+    public function beforeSave()
+    {
+        $this->_dataSaveAllowed = false;
+        $value = (string)$this->getValue();
+        if ($this->isAPiKeyValid($value)) {
+            parent::beforeSave();
+        } else {
+            $this->setValue('');
+            $message = __('API key is invalid and can not be saved. Please, check it and try again.');
```
sivaschenko: I don't think we should prevent admin user from saving API key if it cannot be validated.

The validation can be failed because of network restrictions or issues. Admin user should be able to configure Magento instance that has limited network access.

Considering the above I'd use the `Magento\Config\Model\Config\Backend\Encrypted`  backend model and rely on admin user ability to use the "Test Connection" button

AdobeStockAsset/Controller/Adminhtml/System/Config/TestConnection.php
```diff
@@ -34,60 +37,132 @@ class TestConnection extends Action
...
+     *
+     * @param bool $isConnectionEstablished
+     *
+     * @return string
+     */
+    private function getResultMessage(bool $isConnectionEstablished): string
+    {
+        $message = $isConnectionEstablished ?
+            __('API key is valid. Connection successfully established.')
+            : __('Invalid API Key. Connection refused.');
```
sivaschenko: I'd simplify these messages to "Connection Successful!" and "Connection Failed!" (as Exception can be thrown not only in the case when API Key is invalid but in many other circumstances)

AdobeStockAsset/view/adminhtml/templates/system/config/testconnection.phtml
```diff
@@ -6,13 +6,23 @@
...
-    "fieldToCheck": "success",
-    "failedText": "<?= $block->escapeHtmlAttr(__('Connection failed! Test again?')) ?>", "validation": {}}}'>
-    <span id="<?= $block->escapeHtmlAttr($block->getData('html_id')) ?>_result">
+<div id="<?= $block->escapeHtmlAttr($block->getApiKeyValidationResultContainerId()) ?>" class=""></div>
+<button class="scalable" type="button" id="<?= $block->escapeHtmlAttr($block->getHtmlId()) ?>">
+    <span id="<?= $block->escapeHtmlAttr($block->getHtmlId()) ?>_result">
         <?= $block->escapeHtml($block->getData('button_label')) ?>
     </span>
 </button>
+<script type="text/x-magento-init">
```
sivaschenko, line 15: I'd try to apply a Magento UI components approach for rendering this button. To do this it's needed to:

1. Move button HTML with knockout annotations to a .html template file, i.e.

         <div data-bind="text: message"></div>
         <button data-bind="click: testConnection"><span data-bind="text: buttonLabel"></span></button>

2. Create a component js responsible for providing data for a template, i.e.:

        define([
        'ko',
        'uiComponent',
        'jquery'
        ], function (ko, Component, $) {
        'use strict';

        return Component.extend({
            defaults: {
                buttonLabel: '',
                message: '',
            },
        /**
         * @public
         */
        testConnection: function () {
            $.ajax({type: "POST",
    ....

3. Initialize the component from the phtml template

        <div data-bind="scope: 'testAdobeStockConnection'">
        
        </div>
        <script type="text/x-magento-init">
        {
            "*": {
                "Magento_Ui/js/core/app": {
                    "components": {
                        "testAdobeStockConnection": {
                            "component": "Magento_AdobeStockAsset/js/connection",
                            "template" : "Magento_AdobeStockAsset/connection"
                        }
                    }
                }
            }
        }
        </script>


Disclaimer: The code provides in these examples have never been tested :)

sivaschenko (changes requested):
Thanks for update @coderimus , please see my review comments

coderimus:
@sivaschenko I implemented requested and discussed changes. Please, review the latest version of this implementation.

sivaschenko (approved):

PR #133, opened 2019-06-24
FEATURE: add connection adapter and separate connection initialization from client object

coderimus:
### Description
Currently, there is no ability to handle client initialization. It is encapsulated in the [Client class](https://github.com/magento/adobe-stock-integration/blob/develop/AdobeStockClient/Model/Client.php#L170) and strictly used config settings. But there are some cases when this connection initialization should be managed by plugins:
1. validate is admin user session alive and this user has access to make an API call
1. validation before saving the API key in the system settings.

In both case with the currently existed implementation to achieve the described result, for example, session validation, you will need to handle different methods such as search or create attributes. Which is not so flexible and stable.

To avoid this and decouple Client object with responsibilities  next was done:

1. A new class added - `ConnectionAdapter`. Its main role is to initialize a connection to the Adobe Stock API. It has one public method `initializeConnection` which instantiate an `AdobeStock` object from the PHP library from Adobe. Also, this class has additional data interface which provides required parameters setters and getters.
1.` Client` class calls `ConnectionAdapter` to initialize connection wich move this responsibility away from it.

**Benefits**
1. Possible to use plugins over the API connection
1. Decouple responsibility

### Fixed Issues
N/A

### Manual testing scenarios
Try to initialize connection by visiting image grid and test connection in settings


AdobeStockClient/Model/ConnectionAdapter.php
```diff
@@ -0,0 +1,94 @@
...
+     * @var string
+     */
+    private $targetEnvironment;
+
+    /**
+     * @inheritdoc
+     */
+    public function initializeConnection(): AdobeStock
+    {
+        $connection = new AdobeStock(
```
sidolov: It's not a good practice to introduce temporal coupling. Your initialization method depends on data that should be provided before the method call.

AdobeStockClientApi/Api/ConnectionObjectInterface.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
sidolov: Unused interface.

sidolov (changes requested):
We extracted client logic to the separate modules to achieve one major goal - incapsulate all dependencies to the 3d party SDK in one place, so if you want to use another SDK - you just need to remove client module and implement the new one. 
I'm not sure that we need to extract connection to the separate class with own interface, you may customize connection by client replacement or Config pluginization in case when you need to modify client params.

AdobeStockClientApi/Api/ConnectionObjectInterface.php
coderimus: This interface consists of methods which can be used for current adobe stock integration and just added to the ConnectionAdapterInterface. This will not affect any other 3rd party implementation if the only initiateConnection method is needed to be changed via preference.

coderimus:
@sidolov thank you for your review. This is my explanation:

> We extracted client logic to the separate modules to achieve one major goal - incapsulate all dependencies to the 3d party SDK in one place, so if you want to use another SDK - you just need to remove client module and implement the new one.

Yes, I see your point and totally agree with you that this is a nice idea. But it is good to have some entry point for customisations even inside one encapsulated module. This adapter is a part of a current adobeClient module. It doesn't pretend to be a separate instance. Its role is just to decouple 2 entities: client and connection.

> I'm not sure that we need to extract connection to the separate class with own interface, you may customize connection by client replacement or Config pluginization in case when you need to modify client params.

I also thought about this but came to the next conclusion: the Client class provides a client object which has the main functionality for operating with the Adobe Stock service. At the same time, it initializes a connection to this service which looks like another responsibility - I mean this connection initialization. By having an adaptor for this we achieve 2 benefits:
1. Separate connection initialization from the client
1. Add the ability to handle connection initialization process by not modifying the client or config class. I think that plugin over the Config class is not so good because it provides you with limited variants: you can change just a config value and that's all.  For example, if you need to validate is customer able to establish connection or not you will need to plugin/observe every public method. While if you have an entry point for connection initialization you can simply manage it without any mutation or change for other objects.

coderimus:
@sidolov @sivaschenko I provided a new way for connection initialization through the fabric pattern. Please, review.

AdobeStockClient/Model/Client.php
```diff
@@ -155,25 +171,38 @@ private function createAttributes(string $idFieldName, array $itemData): array
...
-    private function getAccessToken()
+    private function getClient(): AdobeStock
     {
-        return null;
+        try {
+            $apiKey = $this->config->getApiKey();
+            $productName = $this->config->getProductName();
+            $targetEnvironment = $this->config->getTargetEnvironment();
+            /** @var ConnectionFactory $connectionInstance */
+            $connectionInstance = new ConnectionFactory($apiKey, $productName, $targetEnvironment);
```
sivaschenko: I'd avoid using `new` operator for i.e. unit testing. It's better to pass these three arguments directly to `createConnection` method. `ConnectitonFactory` should be injected via DI. Also, it would be good to cache the created connection in the corresponding property.

AdobeStockClient/Model/Client.php
```diff
@@ -155,25 +171,38 @@ private function createAttributes(string $idFieldName, array $itemData): array
...
 
     /**
-     * TODO: Implement retriving of an access token
-     *
-     * @return null
+     * @return AdobeStock
+     * @throws IntegrationException
      */
-    private function getAccessToken()
+    private function getClient(): AdobeStock
```
sivaschenko: Please rename the method to `getConnection`

AdobeStockClient/Model/ConnectionFactory.php
```diff
@@ -0,0 +1,66 @@
...
+    public function createConnection(): AdobeStock
+    {
+        try {
+            return new AdobeStock(
+                $this->apiKey,
+                $this->productName,
+                $this->targetEnvironment
+            );
+        } catch (\Exception $exception) {
+            $message = 'An error occurred during Stock API connection initialization: %1';
```
sivaschenko: I don't think this try catch is needed if you'd like to move all the exception related logic to a separate private method in the Client (https://github.com/magento/adobe-stock-integration/pull/123)

sivaschenko (changes requested):
Thanks for updates @coderimus Please see my review comments

AdobeStockClient/Model/ConnectionFactory.php
coderimus: Yes, it's handling will be moved to the exception related logic. Here I put it just as a note for myself to not to forget that this variant also should be taken into account.

AdobeStockClient/Model/Client.php
coderimus: Thank you for review. Will change.

coderimus: @sidolov requested changes implemented. Please, check when you will have time :)

sivaschenko (approved):

AdobeStockClient/Model/ConnectionFactory.php
```diff
@@ -0,0 +1,28 @@
...
+class ConnectionFactory
+{
+    /**
+     * @param string $apiKey
+     * @param string $productName
+     * @param string $targetEnvironment
+     *
+     * @return AdobeStock
+     */
+    public function createConnection(string $apiKey, string $productName, string $targetEnvironment): AdobeStock
```
sidolov: Let's rename to `create`, we already have `ConnectionFactory` in the class name, it would be better to simplify the method name.

sidolov (changes requested):
@coderimus one minor issue left, the implementation looks good! Thank you!

AdobeStockClient/Model/ConnectionFactory.php
sivaschenko: Fixed

sidolov (approved):

Issue #149, opened 2019-07-03
System configuration outside system section


diazwatson:
### Preconditions (*)

1. Checkout `develop` branch

### Steps to reproduce (*)

1. Go to Stores -> Configuration -> Advanced -> System

### Expected result (*)

![Screenshot 2019-07-03 at 11 16 58](https://user-images.githubusercontent.com/1080386/60584159-678c4280-9d84-11e9-8a4d-f6593db2bcde.png)

### Actual result (*)

![Screenshot 2019-07-03 at 10 53 33](https://user-images.githubusercontent.com/1080386/60584206-87236b00-9d84-11e9-9541-89d66393d106.png)

Issue #93, opened 2019-06-13
Specify correct ACL resources for UI component and backend resources


sidolov:
### Description
- Specify correct ACL resources for UI component and backend resources

### Documentation
Story: [User controls access to Adobe Stock images from Admin Panel in ACL](https://github.com/magento/adobe-stock-integration/issues/42)

### QA notes
1. User opens Resources tree and sees Adobe Stock Config node under Settings - Configuration - System Section.
2. User includes this resource into the role. That will allow admin user with this role to set up an integration with Adobe stock
3. User sees Adobe Stock node under Content - Elements - Media Gallery. This node has 2 child nodes:
  * Save preview images
  * License images

4. User includes Save preview images resource into the role. This will allow admin user with the role assigned only search for stock images and same their previews into Media Gallery. License image is not accessible.
5. If user doesn't select any of the 2 options, user with the role won't see the access to Adobe Stock from Media Gallery
6. If user selects the parent node Adobe Stock, both child nodes get selected.

vpodorozh:
#kharkiv2019

PR #150, opened 2019-07-03
#149 Move conf back to system section

diazwatson:
#149 

### Preconditions (*)

1. Checkout `develop` branch

### Steps to reproduce (*)

1. Go to Stores -> Configuration -> Advanced -> System

### Expected result (*)

Screenshot 2019-07-03 at 11 16 58

### Actual result (*)

Screenshot 2019-07-03 at 10 53 33


vpodorozh (changes requested):
Hi @diazwatson - it was moved outside the `system` section in order to add ACL, check this issue #93  .
Why should we move it back to `system` section?
Thx!

diazwatson:
Hi @vpodorozh you are right that is what issue #93 tells to do, although issue #29 also establish that the  Adobe Stock Integration configuration options should go inside Store-Configuration-Advanced-System.

I think we need @sidolov for clarification

sivaschenko:
That's correct, thanks @diazwatson according to requirements the configuration should be located in the system section, I missed this point.

As there is no way to specify ACL for a system configuration group we'll have to use `Magento_Config::config_system`. So looks like `Magento_AdobeStockAsset::config` should be removed.

sivaschenko (approved):

diazwatson:
@sidolov thanks for clarifying this.

PR #158, opened 2019-07-06
Fix phpdoc and add return types to AssetInterface

hws47a:
### Description (*)
- Improve `Magento\AdobeStockAssetApi\Api\Data\AssetInterface`:
  - Fix incorrect PhpDoc: set methods return void instead of $this
  - Add void return type to methods with appropriate phpdoc
  - change setId and setExtensionAttributes method signatures to return void instead of self as it should be by design.
- Change `Magento\AdobeStockAsset\Model\Asset` to fit it's interface

### DIFF
--- a/AdobeStockAsset/Model/Asset.php
+++ b/AdobeStockAsset/Model/Asset.php
@@ -30,9 +30,9 @@ class Asset extends AbstractExtensibleObject implements AssetInterface
     /**
      * @inheritdoc
      */
-    public function setId($value)
+    public function setId($value): void
     {
-        return $this->setData(self::ID, $value);
+        $this->setData(self::ID, $value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/adobe-stock-integration?pullRequest=158) <br/>All committers have signed the CLA.

sivaschenko:
Hi @hws47a , thanks for the pull request!

The proposed changes do make sense in the context of strict CQRS principle following. However, could you please share what is this approach based on? Are there any particular standards or guidelines that you've used?

hws47a:
@sivaschenko 

**For explicit return types:**
Magento technical guidelines:
> 1.2. Explicit return types MUST BE declared on functions.

**For changed `$this` to `void` in `setId` and `setExtensionAttributes`:**
I cannot find where exactly I've seen that it's a recommended way.

As example I can show two newest Magento core classes:
Inventory: https://github.com/magento/inventory/blob/2.3-develop/app/code/Magento/InventoryApi/Api/Data/SourceInterface.php
GraphQlCache: https://github.com/magento/graphql-ce/blob/2.3-develop/app/code/Magento/GraphQlCache/Model/CacheableQuery.php

Anyway, all of other methods returns `void` and I believe class should follow same design for all methods.

sivaschenko:
Great, thank you! I am ok with convention that all setters in adobe-stock-integration API will return void. @sidolov any objections?

sidolov (approved):

Issue #163, opened 2019-07-09
Writing integration test for module AdobeStockClient for client


engcom-Golf:
### Description
Write integration test for class `\Magento\AdobeStockClient\Model\Client`

PR #165, opened 2019-07-10
ASI-163: Writing integration test for module AdobeStockClient for client

engcom-Golf:
### Description (*)
Integration test added for module AdobeStockClient.

1. magento/adobe-stock-integration#163: Writing integration test for module AdobeStockClient for client

1. Run integration test:
- \Magento\AdobeStockClient\Test\Integration\Model\ClientTest


diazwatson (commented):
@engcom-Golf couple of question regarding this PR.
I have ran the test and noticed that only 60% of the file is covered.

- Are we aim to provide 100% for each file?
- Is it worth to also test the exceptions?

sivaschenko:
I'd agree with @diazwatson Client is the most critical part of the integration, so I'd prefer testing both positive and negative scenarios to cover Client code completely

engcom-Golf:
@sivaschenko @diazwatson Hello. I've covered other cases, you may check it again.

diazwatson:
Hi @engcom-Golf you now reached **77% covered**, nice!

Screenshot 2019-07-17 at 13 19 20

How difficult would be to also include coverage for `testConnection` method and the exception thrown by the `createAttribute` method? I think that would give us 100% coverage for the Client.!

I can see the `testConnection` method has a **_Todo_** note saying we should refactored so maybe things are going to check in the future but even so I think covering this method tests would actually help us to refactor.

engcom-Golf:
@diazwatson Oh, sorry, I have not seen catch in method createAttrbutes, I'll add case for this case.
As for method testConnection - I didn't cover this method because I creating mock for connection, so I see no reason to check data which I added in mock.

engcom-Golf:
@diazwatson You may check it again :)

diazwatson (approved):

Issue #103, opened 2019-06-14
stock.adobe.io returns 500 for some requests even when they looks like valid


TomashKhamlai:
### Preconditions
Admin is authorized using Session-based authentication

### Steps to reproduce:
1. Perform request from the browser without URL encoding
`{{base_url}}/admin/mui/index/render/?namespace=adobe_stock_images_listing&search=popular&paging[pageSize]=4&paging[current]=97848&filters[placeholder]=true&isAjax=true`
2. Verify that
```
Server error: `GET https://stock.adobe.io/Rest/Media/1/Search/Files?locale=en_US&search_parameters%5Blimit%5D=4&search_parameters%5Boffset%5D=391388&search_parameters%5Bwords%5D=popular&result_columns%5B0%5D=nb_results&result_columns%5B1%5D=id&result_columns%5B2%5D=comp_url&result_columns%5B3%5D=thumbnail_500_url&result_columns%5B4%5D=width&result_columns%5B5%5D=height` resulted in a `500 Internal Server Error` response: {"error":"UNIVERSAL_SEARCH error: [400] {\"status\":[{\"scope\":\"stock\",\"status\":400}],\"metrics\":{\"total_hits\":0 (truncated...)
```
is returned.
3. Perform different request from the browser with URL encoding
`{{base_url}}/admin/mui/index/render/?namespace=adobe_stock_images_listing&search=pop&paging%5BpageSize%5D=4&paging%5Bcurrent%5D=97748&filters%5Bplaceholder%5D=true&isAjax=true`

### Expected result:
1. `stock.adobe.io` is more error resistant
2. Adobe Stock Integration is smart enough to block request that leads to errors

### Actual result:
1. Server code - 500
```
Server error: `GET https://stock.adobe.io/Rest/Media/1/Search/Files?locale=en_US&search_parameters%5Blimit%5D=4&search_parameters%5Boffset%5D=390988&search_parameters%5Bwords%5D=pop&result_columns%5B0%5D=nb_results&result_columns%5B1%5D=id&result_columns%5B2%5D=comp_url&result_columns%5B3%5D=thumbnail_500_url&result_columns%5B4%5D=width&result_columns%5B5%5D=height` resulted in a `500 Internal Server Error` response: {"error":"UNIVERSAL_SEARCH error: [400] {\"status\":[{\"scope\":\"stock\",\"status\":400}],\"metrics\":{\"total_hits\":0 (truncated...)
```
[...]

PR #227, opened 2019-07-15
ASI-103: stock.adobe.io returns 500 for some requests even when they looks like valid

engcom-Golf:
### Description (*)
In case of a search error, reset the result.

1. magento/adobe-stock-integration#103: stock.adobe.io returns 500 for some requests even when they looks like valid

### DIFF
--- a/AdobeStockClient/Model/Client.php
+++ b/AdobeStockClient/Model/Client.php
@@ -129,7 +129,7 @@ class Client implements ClientInterface
             }
             $totalCount = $response->getNbResults();
         } catch (Exception $e) {
-            $this->logger->critical($exception->getMessage());
+            $this->logger->critical($e->getMessage());
         }
 
         $searchResult = $this->searchResultFactory->create();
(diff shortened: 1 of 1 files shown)


AdobeStockClient/Model/Client.php
```diff
@@ -126,28 +126,32 @@ public function search(SearchCriteriaInterface $searchCriteria): SearchResultInt
...
 
-            $items = [];
-            /** @var StockFile $file */
-            foreach ($response->getFiles() as $file) {
-                $itemData = (array)$file;
-                $itemData['thumbnail_url'] = $itemData['thumbnail_240_url'];
-                $itemData['preview_url'] = $itemData['thumbnail_500_url'];
-                $itemId = $itemData['id'];
-                $attributes = $this->createAttributes('id', $itemData);
+            try {
```
sivaschenko: It would be better to avoid multiple try-catch levels and separate them to a separate try-catch blocks if possible

sivaschenko (changes requested):
@engcom-Golf please take a look at my review comments

AdobeStockClient/Model/Client.php
engcom-Golf: @sivaschenko Fixed, please check it again

sivaschenko (approved):

Issue #128, opened 2019-06-24
Client error: Api Key is invalid if User is allowed to click on by client and is not allowed by server


TomashKhamlai:
### Preconditions:
1. php bin/magento config:set adobe_stock/integration/enabled 1
2. php bin/magento config:set adobe_stock/integration/api_key \<key\>
3. php bin/magento config:set cms/wysiwyg/enabled disabled
4. php bin/magento config:set admin/security/admin_account_sharing 1
5. php bin/magento config:set admin/security/use_form_key 0
6. php bin/magento cache:flush
7. php bin/magento config:set adobe_stock/integration/api_key ""
8. php bin/magento cache:flush

### Steps to reproduce:
1. Navigate to Content > Elements > Pages
2. Click Add New Page button
3. Fill Title with "Test CMS Page"
4. Expand the "Content" section
5. Fill Content Heading with "Test Content Heading"
6. Click "Insert Image..." button
7. Click "Search Adobe Stock" button

### Expected result:
:heavy_check_mark: 1. Step 7 is impossible - "Search Adobe Stock" is not visible on the page
:heavy_check_mark: 2. or/and "Invalid API key" alert appears because of validation.
(Link to settings page is +1 for usability:{{base_url}}/admin/admin/system_config/edit/section/system/group/adobe_stock_integration/)

### Actual result
:x: 1. Error: 
```
Client error: GET https://stock.adobe.io/Rest/Media/1/Search/Files?locale=en_US&search_parameters%5Blimit%5D=20&search_parameters%5Boffset%5D=0&result_columns%5B0%5D=id&result_columns%5B1%5D=nb_results&result_columns%5B2%5D=comp_url&result_columns%5B3%5D=thumbnail_500_url&result_columns%5B4%5D=width&result_columns%5B5%5D=height resulted in a 403 Forbidden response: {"error_code":"403003","message":"Api Key is invalid"}
```
![image](https://user-images.githubusercontent.com/11827230/60022404-6bd9a100-969c-11e9-9682-f53f6537bb5a.png)

Related to: #85

PR #228, opened 2019-07-16
ASI-128: Client error: Api Key is invalid if User is allowed to click on by client and is not allowed by server

engcom-Golf:
### Description (*)
Change error message and add URL to API key page configuration.

1. magento/adobe-stock-integration#128: Client error: Api Key is invalid if User is allowed to click on by client and is not allowed by server

### DIFF
--- a/AdobeStockImageAdminUi/Model/Listing/DataProvider.php
+++ b/AdobeStockImageAdminUi/Model/Listing/DataProvider.php
@@ -99,7 +99,7 @@ class DataProvider extends \Magento\Framework\View\Element\UiComponent\DataProvi
             $message = __('Error with message: %1', $e->getMessage());
             if ($e->getCode() === 403) {
                 $message = __(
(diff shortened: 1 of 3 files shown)


AdobeStockImageAdminUi/Model/Listing/DataProvider.php
```diff
@@ -70,19 +79,33 @@ public function __construct(
...
+                $result->getItems(),
+                $result->getTotalCount(),
+                $this->getSearchCriteria(),
+                'id'
+            );
+        } catch (\Exception $e) {
+            $message = __('Error with message: %1', $e->getMessage());
+            if ($e->getCode() === 403) {
+                $message = __(
+                    'You have problem with API key, please check it on <a href="%url'
```
sivaschenko:
```suggestion
                    'Please ensure the Adobe Stock API key is specified correctly on the <a href="%url'
```

sivaschenko (approved):

Issue #231, opened 2019-07-17
Add a tooltip to Isolated Assets filter


sivaschenko:
A tooltip has to be added to the Isolated Assets filter.

*Current UI:*

![image](https://user-images.githubusercontent.com/2028541/61377017-42c2bf80-a89a-11e9-8206-819523094b9b.png)

*Desired UI:*

![image](https://user-images.githubusercontent.com/2028541/61376918-11e28a80-a89a-11e9-87f8-c3fd46f0592a.png)

PR #236, opened 2019-07-18
#231 : Add a tooltip to Isolated Assets filter

konarshankar07:

1. magento/adobe-stock-integration#231: Add a tooltip to Isolated Assets filter
2. ...

image

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
@@ -79,7 +79,7 @@
             min-width: 70%;
         }
     }
-    .admin__field {
+    .admin__adobe-stock-image-checkbox {
         .admin__field-control._with-tooltip {
             .admin__form-field-label {
                 display: inline-block;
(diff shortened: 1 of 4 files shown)


AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
```diff
@@ -79,4 +79,14 @@
             min-width: 70%;
         }
     }
+    .admin__field {
```
sivaschenko: Please do not use general selectors for styling concrete elements, rely only on the specific selector to the target element instead

sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request, please see my comment

AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
konarshankar07: Hello @sivaschenko 
I've made sure that our adobe CSS won't break any style for core Magento elements as I've added label inside the tooltip but I've also added a new selector as per your requested changes.

Thanks

sivaschenko (approved):

Issue #96, opened 2019-06-13
User saves image attributes to the DB


sidolov:
### Description
- User saves image preview to the Magento media gallery

### Documentation
Story: [User saves image preview](https://github.com/magento/adobe-stock-integration/issues/26)

### QA notes
* The following image attributes associated with the stock image are saved into Magento DB:
  * id - Asset's unique identifier 
  * title - Asset's title
  * creator_id - Unique identifier for the asset's creator
  * creator_name - Asset creator's name	
  * creation_date - Date on which the asset was created		
  * country_name	
  * is_licensed - license applied
  * comp_url - URL to the watermarked version of the asset
  * comp_width - Width in pixels of the asset's complementary (unlicensed) image
  * comp_height	- Height in pixels of the asset's complementary (unlicensed) image	
  * category - JSON structure with information about the category assigned to the asset
    * category{id}
    * category{name}
  * keywords - List of localized keywords for the asset
  * media_type - custom Magento attribute. Value is based on stock media_type_id.
  * content-type - Mime type of the asset's content. E.g.image/jpeg
  * details_url - URL to the Adobe Stock details page for the asset. If you pass the Authorization header with the call, Adobe Stock generates an SSO jump URL.
  * premium_level_id- Asset's premium (pricing) level.

### Dependency
Has dependency on https://github.com/magento/adobe-stock-integration/issues/95

PR #248, opened 2019-07-23
96 User saves image attributes to the DB

coderimus:
### Description
This PR contains changes according to the [96: User saves image attributes to the DB](https://github.com/magento/adobe-stock-integration/issues/96) issue

**Current state - WORK IN PROGRESS - image is possible to be saved - left small adjastments**

* `SaveImagePriview` api endpoint added

* `Category`, `Creator`, `Media type`, `Premium level` introduced as components of an asset object to provide more flexibility for saving asset strategy.

* More fields were added to the search result response for `GetImageList`. Now it contains additional information about the image such as creator name, creator id, category name, category id, title etc.

* `Creator`, `Category`, `MediaType` and `PremiumLevel` repositories along with the data classes added.

* New filter `MediaId` added. It is only for internal needs such as a part of an image flow when you have to get image data for save. Will not be added to the UI filters of an Adobe Stock image grid.

**Implementation notes:**

At the `Magento\AdobeStockImage\Model\IsMediaValidForSave\IsImageExistsConditionChain` has next logic :
```php
        if (0 === $searchResults->getTotalCount()) {
            return false;
        }

        if (1 < $searchResults->getTotalCount()) {
            return false;
        }

        return true;
```
The current implementation of the Adobe Stock API by itself doesn't have a direct filter for adobe media id and not documented (as I understood) adobe id identifier minimum length. So this can leads to next: if the `$mediaId` parameter is equal, for example, to 9999 the result will not be empty (because there is no such image, but it will return similar images where 9999 is a part of id, image picture or else. Provided condition covers this situation and throws an exception.

### Fixed Issues
N/A

### Manual testing scenarios
Will be added after changing status from being draft.


AdobeStockImage/Model/SaveImagePreview.php
```diff
@@ -0,0 +1,87 @@
...
+        try {
+            $searchResult = $this->getImage->execute($mediaId);
+            if (!$this->isMediaExistsConditionChain->execute($searchResult)) {
+                return false;
+            }
+            $items = $searchResult->getItems();
+            /** @var AssetInterface $item */
+            $asset = reset($items);
+            //@TODO replace with the save image file class call (or move it to save asset processor when it will be merged into develop.
+            $imagePath = '';
```
sivaschenko: Please use the `\Magento\AdobeStockImage\Model\Storage::save` that is already merged to develop.

AdobeStockImage/Model/SaveImagePreview.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function execute(int $mediaId, string $destinationPath): bool
+    {
+        try {
+            $searchResult = $this->getImage->execute($mediaId);
+            if (!$this->isMediaExistsConditionChain->execute($searchResult)) {
```
sivaschenko: I don't think it is necessary to introduce a new class
```suggestion
            if (!$searchResults->getTotalCount()) {
```

AdobeStockImage/Model/SaveImagePreview.php
```diff
@@ -0,0 +1,87 @@
...
+                return false;
```
sivaschenko: let's throw a `LocalizedException` with a message instead of just returning `false`

AdobeStockImage/Model/SaveImagePreview.php
```diff
@@ -0,0 +1,87 @@
...
+            //@TODO replace with the save image file class call (or move it to save asset processor when it will be merged into develop.
+            $imagePath = '';
+            $asset->setPath($imagePath);
+            $this->savePreviewImageAssetStrategy->execute($asset);
+
+            return true;
+        } catch (\Exception $exception) {
+            $message = __('Image was not saved: %1', $exception->getMessage());
+            $this->logger->critical($message);
+            return false;
```
sivaschenko, line 165: let's throw a `LocalizedException` with a message instead of just returning `false`

AdobeStockImage/Model/IsImageValidToSaveCondition/IsImageExistsConditionChain.php
```diff
@@ -0,0 +1,37 @@
...
+declare(strict_types=1);
+
+namespace Magento\AdobeStockImage\Model\IsImageValidToSaveCondition;
+
+use Magento\AdobeStockAssetApi\Api\Data\AssetSearchResultsInterface;
+
+/**
+ * Class IsMediaExistsConditionChain
+ */
+class IsImageExistsConditionChain
```
sivaschenko: I don't think this class is necessary

AdobeStockImage/Model/GetImageList.php
```diff
@@ -104,4 +141,75 @@ public function execute(SearchCriteriaInterface $searchCriteria): AssetSearchRes
...
+     */
+    private function convertDocumentToAsset(Document $item): AssetInterface
+    {
+        //@TODO granulate this to components and make asset part construction possibly in builder
+
+        /** @var AssetInterface $asset */
+        $asset = $this->assetFactory->create();
+
+        $asset->setId($item->getId());
+        $asset->setThumbnailUrl($item->getCustomAttribute('thumbnail_url')->getValue());
```
sivaschenko: I'd use a multidimensional array containing mapping and simply go through this mapping using a recursive method with a foreach loop.

It's better to place this conversion in AdobeStockClient module as fields are part fo Adobe Stock SDK interface

AdobeStockImage/Model/GetImage.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function execute(int $mediaId): AssetSearchResultsInterface
+    {
+        try {
+            $searchCriteria = $this->constructSearchCriteria($mediaId);
+            /** @var AssetSearchResultsInterface $result */
+            $result = $this->imageList->execute($searchCriteria);
+
+            return $result;
+        } catch (\Exception $exception) {
```
sivaschenko: I don't think handling exception here is necessary `GetImageListInterface::execute` throws a localized exception that is ready for end-user

AdobeStockClient/Model/Client.php
```diff
@@ -216,17 +220,24 @@ private function getSearchRequest(SearchCriteriaInterface $searchCriteria): Sear
...
         $resultsColumns = Constants::getResultColumns();
         $resultColumnArray = [];
-        foreach ($this->config->getSearchResultFields() as $field) {
-            $resultColumnArray[] = $resultsColumns[$field];
+        try {
+            foreach ($this->config->getSearchResultFields() as $field) {
+                $resultColumnArray[] = $resultsColumns[$field];
+            }
+        } catch (Exception $exception) {
+            $message = __('An error occurred during result fields matching: %1', $exception->getMessage());
```
sivaschenko: It would be better to log error stating that "Cannot retrieve the field %field%. It's not available in Adobe Stock SDK." and skip this filed instead of terminating the whole process

AdobeStockClient/Model/Client.php
```diff
@@ -159,11 +159,13 @@ public function search(SearchCriteriaInterface $searchCriteria): SearchResultInt
                 $items[] = $this->convertStockFileToDocument($file);
             }
             $totalCount = $response->getNbResults();
-        } catch (Exception $e) {
-            if (strpos($e->getMessage(), 'Api Key is invalid') !== false) {
-                throw new AuthenticationException(__($e->getMessage()), $e, $e->getCode());
+        } catch (Exception $exception) {
```
sivaschenko, line 162: Because of Adobe Stock API specifics, it was decided to intentionally return the empty result instead of exception if the search fails for reasons other than incorrect credentials.

AdobeStockAssetApi/Api/Data/AssetInterface.php
```diff
@@ -23,10 +25,10 @@ interface AssetInterface extends \Magento\Framework\Api\ExtensibleDataInterface
     const ADOBE_ID = "adobe_id";
     const CREATED_AT = 'created_at';
     const UPDATED_AT = 'updated_at';
-    const MEDIA_TYPE_ID = 'media_type_id';
+    const MEDIA_TYPE = 'media_type';
```
sivaschenko: Are you sure it is necessary to introduce `media_type` and `premium_level` extensible entities?

It's enough to have hardcoded id to name for these fields in my opinion

AdobeStockAsset/Model/Components/SaveCreatorAssetComponent.php
```diff
@@ -0,0 +1,75 @@
...
+use Magento\AdobeStockAssetApi\Api\CreatorRepositoryInterface;
+use Magento\AdobeStockAssetApi\Api\Data\CategorySearchResultsInterface;
+use Magento\AdobeStockAssetApi\Api\Data\CreatorInterface;
+use Magento\Framework\Api\SearchCriteriaBuilder;
+use Magento\Framework\Exception\AlreadyExistsException;
+
+/**
+ * Class SaveCreatorAssetComponent
+ */
+class SaveCreatorAssetComponent
```
sivaschenko: These `Save*Component` classes are unnecessary. The repository should be completely and fully responsible for this operation

sivaschenko (changes requested):
Thanks for working on this task @coderimus ! Please see my review comments

AdobeStockImage/Model/SaveImagePreview.php
coderimus: @sivaschenko thank you for your review. I have just pushed updated code where this functionality was included. It works as expected but I will request some consultation for managing the destination path. Will write tomorrow my question.

AdobeStockAssetApi/Api/Data/AssetInterface.php
coderimus: @sivaschenko  I decided to introduce this implementation because of the next: the asset looks like a complex object consists of the several required components: the category, creator, media type and premium level.
They have different complicity structure but in fact, they all are required to be saved and provide internal id's to link them into the asset object.

Having this as a point I decided to use the strategy for saving images (preview and licensed in future) where we, at the first step, save asset components and then asset with the related entities. From my point of view, having this approach we provide more flexibility for managing the asset save process, use components independently for different scenarios (I think we will need this when the licensed image part will be in progress), cover with tests should be less complicated.

AdobeStockImage/Model/GetImageList.php
coderimus: I think that it is a little bit risky to use a recursive mapping because the API can be changed or even separated on several calls or else. If we want to have a mapping we should use some adapter to manage this to be saved from Adone API changes.

Or use components :) Where every part of it is an independent object which can be map/get/iterate whatever else to get the asset data in a needed format.

AdobeStockAsset/Model/Components/SaveCreatorAssetComponent.php
coderimus: Let me describe the idea of having this approach: if the asset or a category save action would be a simple and straight forward operation I would totally agree with you that their repositories can handle such responsibility. But we have 2 scenarios: 

**The first one:**
for example, category object. The adobe_id is not unique and, as a result, every time you use the `categoryRepository->save($category)` you generate new entry with the same `adobe_id` and possible the same name. As I can see, we do not have the possibility to build one-to-many relation between asset and categories. So this approach is handled inside the save component class.

**The second one**
I decided to propose those services to exclude saving logic from repositories and make them as simple as possible to not make any hidden knowledge. As I can see we will have 2 different types of assets: preview image and licensed. They are similar in some parts but will have different business logic. Having this save services for assets components we can delegate any kind of validation, manipulation and business logic to services and keep repositories with the simplest responsibility - provide data to the presentation layer and save it to domain.

AdobeStockImage/Model/SaveImagePreview.php
coderimus: I will change to the `NotFoundException` and throw `LocalizedException` at the presentation layer. Here, in the domain, as I remember, it is not suggested to throw localized exceptions.

AdobeStockImage/Model/SaveImagePreview.php
coderimus: I added it as an entry point of possible condition extensibility and reuse for a licensed image. Should remove it?

coderimus:
@sivaschenko  I have a question: how we want to proceed with the download process when the same image is triggered to be downloaded? For example, image **A** is just downloaded to the folder **A-folder**. User press next button, press the return button at the image list and press the download button for image **A**. What is the expected behaviour?

sivaschenko:
This PR can be merged once the second Save button is removed from the UI.

coderimus:
@sivaschenko [with this commit](https://github.com/magento/adobe-stock-integration/pull/248/commits/7f95f49ca0fe24eb4186cf952ad2328c87e6e0ab) I have fixed the issue when you tried to save the asset data with the same adobe_id: `unique constraint violation found` With the provided fix, we identify is this asset is new or not. If new - `$asset->isObjectNew(true);` Otherwise, no and just update requested asset data.

coderimus:
@sivaschenko with the [latest commit](https://github.com/magento/adobe-stock-integration/pull/248/commits/fece3da93a91a3b502390539636cd64d4ca0cc7e) I provided functionality or saving media type and premium level information. It is required for saving asset data because of fk. According to our previous discussion, I simplified the premium level and media type objects with excluding extension attributes and save them with every asset save action call. I have to add repositories and models to be able to exclude resource models in future.

sivaschenko (dismissed):

sivaschenko (approved):

PR #249, opened 2019-07-24
Refactor: clean up the product image delete plugin and small fix for listing xml

coderimus:
### Description
The main goal of this PR is to refactor the `AdobeStockImage\Plugin\Product\Gallery\Processor` with the next changes:

* Change visibility scope from protected to private for properties
* Use only search criteria builder factory with the `addFilter` instead of injecting filter builder and use `addFilters` (we have the only one - path :) )
* Log an exception if any appeared during the plugin method run
* Adjust Unit Tests according to the refactored changes
* Set a sort order 
* Remove unused connection of a class at `AssetRepository`

N/A

N/A

### DIFF
--- a/AdobeStockAsset/Model/Config.php
+++ b/AdobeStockAsset/Model/Config.php
@@ -15,6 +15,9 @@ use Magento\Framework\App\Config\ScopeConfigInterface;
  */
 class Config
 {
+    /**
+     * Path to enable/disable adobe stock integration in the system settings.
+     */
     private const XML_PATH_ENABLED = 'adobe_stock/integration/enabled';
 
     /**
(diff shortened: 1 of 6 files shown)


AdobeStockAsset/Model/Config.php
```diff
@@ -39,6 +42,6 @@ public function __construct(ScopeConfigInterface $scopeConfig)
      */
     public function isEnabled(): bool
     {
-        return (bool)$this->scopeConfig->getValue(self::XML_PATH_ENABLED);
+        return (bool)$this->scopeConfig->isSetFlag(self::XML_PATH_ENABLED);
```
sivaschenko: `isSetFlag` returns bool, no need to cast it.
```suggestion
        return $this->scopeConfig->isSetFlag(self::XML_PATH_ENABLED);
```

sivaschenko (changes requested):

sivaschenko (approved):

PR #254, opened 2019-07-25
Adjust WebApi tests for AdobeStockAsset module

serhiyzhovnir:
### Description (*)
This PR adjust Web API tests for  AdobeStockAsset module.
Also, it fixes the issues with running Web API tests with SOAP.
The new test `testDeleteWithException` was added.

1. Run Web API tests with REST and SOAP configs

### DIFF
--- a/AdobeStockAsset/Model/Asset.php
+++ b/AdobeStockAsset/Model/Asset.php
@@ -53,15 +53,21 @@ class Asset extends AbstractExtensibleModel implements AssetInterface
     /**
      * @inheritdoc
      */
-    public function getMediaTypeId(): int
+    public function getMediaTypeId(): ?int
     {
-        return (int) $this->getData(self::MEDIA_TYPE_ID);
+        $mediaTypeId = $this->getData(self::MEDIA_TYPE_ID);
+
+        if (!$mediaTypeId) {
+            return null;
+        }
+
+        return (int) $mediaTypeId;
     }
 
     /**
      * @inheritdoc
      */
-    public function setMediaTypeId(int $mediaTypeId): void
+    public function setMediaTypeId(int $mediaTypeId = null): void
     {
         $this->setData(self::MEDIA_TYPE_ID, $mediaTypeId);
     }
@@ -117,15 +123,21 @@ class Asset extends AbstractExtensibleModel implements AssetInterface
     /**
      * @inheritdoc
      */
-    public function getPremiumLevelId(): int
+    public function getPremiumLevelId(): ?int
     {
-        return (int) $this->getData(self::PREMIUM_LEVEL_ID);
+        $premiumLevelId = $this->getData(self::PREMIUM_LEVEL_ID);
+
+        if (!$premiumLevelId) {
+            return null;
+        }
+
+        return (int) $premiumLevelId;
     }
 
     /**
      * @inheritdoc
      */
-    public function setPremiumLevelId(int $premiumLevelId): void
+    public function setPremiumLevelId(int $premiumLevelId = null): void
     {
         $this->setData(self::PREMIUM_LEVEL_ID, $premiumLevelId);
     }
--- a/AdobeStockAsset/Model/Asset.php
+++ b/AdobeStockAsset/Model/Asset.php
@@ -55,19 +55,13 @@ class Asset extends AbstractExtensibleModel implements AssetInterface
      */
     public function getMediaTypeId(): ?int
     {
-        $mediaTypeId = $this->getData(self::MEDIA_TYPE_ID);
-
-        if (!$mediaTypeId) {
-            return null;
-        }
-
-        return (int) $mediaTypeId;
+        return $this->hasData(self::MEDIA_TYPE_ID) ? (int) $this->getData(self::MEDIA_TYPE_ID) : null;
     }
 
     /**
      * @inheritdoc
      */
-    public function setMediaTypeId(int $mediaTypeId = null): void
+    public function setMediaTypeId(?int $mediaTypeId): void
     {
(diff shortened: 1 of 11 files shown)


AdobeStockAsset/Model/Asset.php
```diff
@@ -53,15 +53,21 @@ public function setId($value): void
...
+        }
+
+        return (int) $mediaTypeId;
     }
 
     /**
      * @inheritdoc
      */
-    public function setMediaTypeId(int $mediaTypeId): void
+    public function setMediaTypeId(int $mediaTypeId = null): void
```
sivaschenko: What is the use case when we'd like to call this method without an argument?

AdobeStockAsset/Model/Asset.php
```diff
@@ -53,15 +53,21 @@ public function setId($value): void
...
+    public function getMediaTypeId(): ?int
     {
-        return (int) $this->getData(self::MEDIA_TYPE_ID);
+        $mediaTypeId = $this->getData(self::MEDIA_TYPE_ID);
+
+        if (!$mediaTypeId) {
+            return null;
+        }
+
+        return (int) $mediaTypeId;
```
sivaschenko: What do you think about simplifying this to
```suggestion
        return ($this->getData(self::MEDIA_TYPE_ID) === null) ? null : (int) $this->getData(self::MEDIA_TYPE_ID);
```

sivaschenko (dismissed):
Hi @serhiyzhovnir thanks for improvements, please see my review comments

AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
```diff
@@ -68,17 +115,22 @@ private function deleteAsset(int $assetId)
                 'httpMethod'   => Request::HTTP_METHOD_DELETE,
             ],
             'soap' => [
-                'service'        => self::SERVICE_NAME,
+                'service' => self::SERVICE_NAME,
                 'serviceVersion' => self::SERVICE_VERSION,
-                'opertion'       => self::SERVICE_NAME . 'DeleteById',
+                'operation' => self::SERVICE_NAME . self::SERVICE_OPERATION
```
diazwatson, line 120: I think it was easier to read with the indentation the way it was

diazwatson (dismissed):

AdobeStockAsset/Model/Asset.php
hws47a: I think the idea is to reset the value to an initial one, so it should be
```php
public function setMediaTypeId(?int $mediaTypeId): void
```

AdobeStockAsset/Model/Asset.php
```diff
@@ -53,15 +53,21 @@ public function setId($value): void
     /**
      * @inheritdoc
      */
-    public function getMediaTypeId(): int
+    public function getMediaTypeId(): ?int
     {
-        return (int) $this->getData(self::MEDIA_TYPE_ID);
+        $mediaTypeId = $this->getData(self::MEDIA_TYPE_ID);
+
+        if (!$mediaTypeId) {
```
hws47a: In this case, you'll get `null` even when you have some value set, for example, `0`
`$mediaTypeId === null` should be used instead or `$this->hasData(…)` method.

So it could look like
```php
return $this->hasData(self::MEDIA_TYPE_ID) ? (int)$this->getData(self::MEDIA_TYPE_ID) : null;
```

hws47a (dismissed):

AdobeStockAsset/Model/Asset.php
sivaschenko: I think the default value should be set explicitly instead of passing `null`

AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
serhiyzhovnir: Hi @diazwatson 
Sure it was easier to read with indentation.
But we have a few tests and few arrays in the tests which do not contain indentations and I think that we need to use a general approach for arrays code style for this project.
The `GetByIdTest` and `SaveTest` do not contain indentations for `$serviceInfo` array and I think if we apply the required changes here where we need to add the indentation the way like it was it will lead to different code styles for arrays for this project.
My point here is to use one approach for arrays code style.
What the general code style for arrays in this project?
@sivaschenko, can you, please, help us with this question?
Thank you!

AdobeStockAsset/Model/Asset.php
serhiyzhovnir: Hi @hws47a 
Thanks for good suggestion!
The changes were applied for `setMediaTypeId ` and  `getPremiumLevelId` methods.

AdobeStockAsset/Model/Asset.php
serhiyzhovnir: Hi @sivaschenko @hws47a 
The code was adjusted and now the definition of `setMediaTypeId ` method looks like the following:
```
public function setMediaTypeId(?int $mediaTypeId): void
```
Thank you!

AdobeStockAsset/Model/Asset.php
serhiyzhovnir: Hi @sivaschenko 
We have used the approach provided by @hws47a.
The method looks like the following for us:
```
public function getMediaTypeId(): ?int
{
    return $this->hasData(self::MEDIA_TYPE_ID) ? (int) $this->getData(self::MEDIA_TYPE_ID) : null;
}
```
Please, let us know if we need to change something.
Thank you!

sidolov (approved):

sidolov (approved):

Issue #259, opened 2019-07-26
Add comment with Redirect URL and Redirect URL pattern next to Client Secret field in Stores Configuration


sivaschenko:
In order to create an integration in console.adobe.com Magento admin needs to enter Redirect URL and Redirect URL pattern.

We should generate values for those fields in Magento and add them as a comment next to Client Secret field in Stores -> Configuration -> System -> Adobe Stock Integration to make it easier for Magento admin to create an integration.

PR #261, opened 2019-07-28
#259 : Add comment with Redirect URL and Redirect URL pattern next to Client Secret field in Stores Configuration

konarshankar07:

1. magento/adobe-stock-integration#259: Add comment with Redirect URL and Redirect URL pattern next to Client Secret field in Stores Configuration

### DIFF
--- a/AdobeStockAsset/Model/System/Config/Comment.php
+++ b/AdobeStockAsset/Model/System/Config/Comment.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+declare(strict_types=1);
+
+namespace Magento\AdobeStockAsset\Model\System\Config;
+
+use Magento\Config\Model\Config\CommentInterface;
+use Magento\Framework\UrlInterface;
+
+/**
+ * Class Comment
+ * @package Magento\AdobeStockAsset\Model\System\Config
+ */
+class Comment implements CommentInterface
+{
+    /**
+     * @var UrlInterface
+     */
+    private $url;
+
+    /**
+     * Comment constructor.
+     * @param UrlInterface $url
+     */
+    public function __construct(
+        UrlInterface $url
+    ) {
+        $this->url = $url;
+    }
+
+    /**
+     * @param string $elementValue
+     * @return string
+     */
+    public function getCommentText($elementValue)
+    {
+        $redirect_url = $this->url->getUrl('adobe_stock/oauth/callback');
+        $redirect_url_pattern = $this->url->getUrl('admin/adobe_stock/oauth/callback/');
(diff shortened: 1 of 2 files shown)


sivaschenko (changes requested):
@konarshankar07 thanks for the pull request. However, the idea is to generate redirect_uri and pattern for admin user based on callback controller url

konarshankar07:
@sivaschenko 

Sorry, I think I haven't understood this task properly. So want to save the redirect url path on system config field?

sivaschenko:
@kronarshankar the expected values shown to admin user should look like:

redirect url: `https://<magento_host>/<admin_path>/adobe_stock/oauth/callback`
redirect url pattern: `https://<magento_host>/<admin_path>/adobe_stock/oauth/callback/`

Where `<magento_host>` and `<admin_path>` are taken from the database, also the url have to include index.php if "Use webserver rewrites" is turned off and secret key if it's turned off.

Simply saying the `UrlInterface` have to be used to get the url that will be shown to admin user

AdobeStockAsset/Model/System/Config/Comment.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * @param string $elementValue
+     * @return string
+     */
+    public function getCommentText($elementValue)
+    {
+        $redirect_url = $this->url->getUrl('adobe_stock/oauth/callback');
+        $redirect_url_pattern = $this->url->getUrl('admin/adobe_stock/oauth/callback/');
```
konarshankar07: Even though I have added `admin` in the redirect_url_pattern but it showing
image

sivaschenko: Please use camelcase for variables names, i.e. `$redirectUrl`.

Redirect url pattern requires escaping, I think you can simply take `$redirectUrl` and add `\\` before each dot, i.e. `$redirectUrlPattern = str_replace('.', '\\.', $redirectUrl)`.

AdobeStockAsset/Model/System/Config/Comment.php
```diff
@@ -0,0 +1,47 @@
...
+        $commentText = sprintf("Configure an Adobe Stock account on the <a href=\"https://console.adobe.io/\" target=\"_blank\">Adobe.io</a> site to retrieve an Private key (Client secret).</br>
+        In order to create an integration in console.adobe.com Magento admin needs to enter %s and %s.", $redirect_url, $redirect_url_pattern);
```
sivaschenko:
```suggestion
        Note: the Default redirect URI is %s and the redirect URI pattern is %s.", $redirectUrl, $redirectUrlPattern);
```

sivaschenko (changes requested):
@konarshankar07 well done! just a couple of comments

AdobeStockAsset/Model/System/Config/Comment.php
sivaschenko: sorry, there is no need to add `admin` to the redirect pattern, that was my typo

sivaschenko (dismissed):

sivaschenko (dismissed):

sivaschenko (approved):

PR #267, opened 2019-07-29
General code improvements

sidolov:
General code improvements

### DIFF
--- a/AdobeStockAsset/Model/Keyword.php
+++ b/AdobeStockAsset/Model/Keyword.php
@@ -43,10 +43,9 @@ class Keyword extends AbstractExtensibleModel implements KeywordInterface
     }
 
     /**
-     * {@inheritdoc}
-     *
+     * @inheritdoc
      */
-    public function getExtensionAttributes()
+    public function getExtensionAttributes(): KeywordExtensionInterface
     {
         return $this->_getExtensionAttributes();
     }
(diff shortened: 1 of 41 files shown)


AdobeStockAsset/Model/Keyword.php
```diff
@@ -40,4 +41,21 @@ public function setKeyword(string $keyword): void
     {
         $this->setData(self::KEYWORD, $keyword);
     }
+
+    /**
+     * {@inheritdoc}
```
sivaschenko:
```suggestion
     * @inheritdoc
```

sivaschenko: Also please add extensible methods to keyword interface

AdobeStockImageAdminUi/view/adminhtml/web/js/action/authorization.js
```diff
@@ -76,20 +80,21 @@ define([], function () {
...
-                            });
-                        } else {
-                            reject(new Error(responseData[authConfig.response.messageIndex]));
-                        }
+                        resolve({
+                            isAuthorized: true,
+                            lastAuthSuccessMessage: responseData[authConfig.response.messageIndex]
+                        });
+                    } else {
+                        stopHandle();
```
sivaschenko: stopHandle should be applied if only if `responseData` is  defined

sivaschenko (changes requested):
@sidolov thanks for the improvements, please see my review comments

sivaschenko (approved):

PR #292, opened 2019-08-03
Code cleanup

diazwatson:
### Description (*)

- Exposed throws of `\AdobeStock\Api\Exception\StockApi`
- `setMediaAdi` expect parameter to be int, string given
- Third parameter should be provided to clarify if types safety important in this context.
- Removed unused imports
- Handled `Null pointer exception may occur here`
- Minor spacing and formatting changes

### DIFF
--- a/AdobeStockAssetApi/Api/Data/CategoryInterface.php
+++ b/AdobeStockAssetApi/Api/Data/CategoryInterface.php
@@ -64,4 +64,12 @@ interface CategoryInterface extends \Magento\Framework\Api\ExtensibleDataInterfa
      * @return void
      */
     public function setExtensionAttributes(CategoryExtensionInterface $extensionAttributes): void;
+
+    /**
+     * Set object as new
+     *
+     * @param bool $true
+     * @return void
+     */
+    public function isObjectNew(bool $true): void;
 }
--- a/AdobeStockAssetApi/Api/Data/CategoryInterface.php
+++ b/AdobeStockAssetApi/Api/Data/CategoryInterface.php
@@ -64,12 +64,5 @@ interface CategoryInterface extends \Magento\Framework\Api\ExtensibleDataInterfa
      * @return void
      */
     public function setExtensionAttributes(CategoryExtensionInterface $extensionAttributes): void;
-
-    /**
-     * Set object as new
-     *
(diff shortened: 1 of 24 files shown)


AdobeStockAssetApi/Api/Data/CategoryInterface.php
```diff
@@ -64,4 +64,12 @@ public function getExtensionAttributes(): CategoryExtensionInterface;
...
      */
     public function setExtensionAttributes(CategoryExtensionInterface $extensionAttributes): void;
+
+    /**
+     * Set object as new
+     *
+     * @param bool $true
+     * @return void
+     */
+    public function isObjectNew(bool $true): void;
```
sivaschenko: Let's not add `isObjectNew` methods to an interface as using resource models for save operations is a temporary solutions

AdobeStockClient/Model/SearchParametersProvider/Sorting.php
```diff
@@ -29,7 +29,7 @@ public function apply(SearchCriteriaInterface $searchCriteria, SearchParameters
     {
         $resultsOrders = Constants::getSearchParamsOrders();
         foreach ($searchCriteria->getSortOrders() as $sortOrder) {
-            if (false !== ($sortOrderField = array_search($sortOrder->getField(), $resultsOrders))) {
+            if (false !== ($sortOrderField = array_search($sortOrder->getField(), $resultsOrders, null))) {
```
sivaschenko, line 32: Why do you think we need to pass `null` explicitly?

sivaschenko (dismissed):
@diazwatson Nice improvements! Thank you! Just a couple of comments

AdobeStockClient/Model/SearchParametersProvider/Sorting.php
diazwatson: Since `null` is the default value, it won't change anything but removes the warning:
> Third parameter should be provided to clarify if types safety is important in this context.

Btw I just noticed that the unit test that covers this fails after the change.
Should we pass `false`, `true` to pass the test and remove the warning or simply revert this change?

What do you think?

sidolov (approved):

Issue #283, opened 2019-08-02
Replace Preview Download controller with webapi endpoint


sivaschenko:
A post web API endpoint pointed to `\Magento\AdobeStockImage\Model\SaveImagePreview::execute` should be created and used instead of `AdobeStockImage/Controller/Adminhtml/Preview/Download.php` controller.

PR #297, opened 2019-08-05
Replace Preview Download controller with webapi endpoint

engcom-Golf:
### Description (*)
Replaced Preview Download controller with webapi endpoint

1. magento/adobe-stock-integration#283: Replace Preview Download controller with webapi endpoint

1. Go to Category -> Content-> Insert Image -> Adobe Stock Image -> Try To save Preview


AdobeStockImage/Model/SaveImagePreview.php
```diff
@@ -113,11 +110,13 @@ public function execute(int $adobeId, string $destinationPath): void
             $path = $this->storage->save($asset->getPreviewUrl(), $destinationPath);
             $asset->setPath($path);
             $this->saveAsset($asset);
+            return $asset->getThumbnailUrl();
```
sivaschenko: The returned URL should be a Magento URL, i.e. `local.magento.com/media/file.png`. It should be retrieved based on asset path

sivaschenko (changes requested):
@engcom-Golf thanks for the PR, please see my comment

AdobeStockImage/Model/SaveImagePreview.php
Nazar65: Ok

sivaschenko: @Nazar65 after changing the frontend approach in https://github.com/magento/adobe-stock-integration/pull/302 the service no longer has to return the url.

It should `return void`

Nazar65: Ok i will update.

sivaschenko:
Blocked by https://github.com/magento/magento2/issues/9138

AdobeStockImage/etc/webapi.xml
```diff
@@ -13,4 +13,10 @@
             <resource ref="Magento_AdobeStockAsset::actions_view"/>
         </resources>
     </route>
+    <route url="/V1/adobestock/saveImagePreview" method="POST">
+        <service class="Magento\AdobeStockImageApi\Api\SaveImagePreviewInterface" method="execute"/>
+        <resources>
+            <resource ref="anonymous"/>
```
sivaschenko, line 19: This WebAPI endpoint should be protected by an appropriate ACL

sivaschenko (changes requested):
Thanks for the pull request @engcom-Golf ! Please see my review comments

sivaschenko:
We are unable to implement this change without a proper webapi ACL and admin session authentication support from Magento framework.

Closing this PR considering the above.

PR #300, opened 2019-08-05
Cleanup

diazwatson:
### Description (*)

My previous PR #292 breaks one of the tests 🤦‍♂  because the 3rd parameter has to be boolean
Screenshot 2019-08-05 at 23 40 33

I have re-checked the [php docs](https://www.php.net/manual/en/function.array-search.php) and I am now passing `false` (default value if no value is passed) and that fix the issue
Screenshot 2019-08-05 at 23 47 40

### DIFF
--- a/AdobeStockClient/Model/SearchParametersProvider/Sorting.php
+++ b/AdobeStockClient/Model/SearchParametersProvider/Sorting.php
@@ -29,7 +29,7 @@ class Sorting implements SearchParameterProviderInterface
     {
         $resultsOrders = Constants::getSearchParamsOrders();
         foreach ($searchCriteria->getSortOrders() as $sortOrder) {
-            if (false !== ($sortOrderField = array_search($sortOrder->getField(), $resultsOrders, null))) {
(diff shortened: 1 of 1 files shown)


AdobeStockClient/Model/SearchParametersProvider/Sorting.php
```diff
@@ -29,7 +29,7 @@ public function apply(SearchCriteriaInterface $searchCriteria, SearchParameters
     {
         $resultsOrders = Constants::getSearchParamsOrders();
         foreach ($searchCriteria->getSortOrders() as $sortOrder) {
-            if (false !== ($sortOrderField = array_search($sortOrder->getField(), $resultsOrders, null))) {
+            if (false !== ($sortOrderField = array_search($sortOrder->getField(), $resultsOrders, false))) {
```
sivaschenko: There is no need to pass optional parameters value, I think that's why they are optional :) Let's remove the last argument
```suggestion
            if (false !== ($sortOrderField = array_search($sortOrder->getField(), $resultsOrders))) {
```

sivaschenko (changes requested):

sivaschenko (approved):

Issue #299, opened 2019-08-05
Move default filter value setting from ContentType filter to GetImageList service


sidolov:
`\Magento\AdobeStockClient\Model\SearchParametersProvider\ContentType::apply` method is currently setting the default filter states for SDK:

        $searchParams->setFilterContentTypePhotos(true);
        $searchParams->setFilterContentTypeIllustration(true);

That logic should not be present in the filter implementation, but should be implemented as part of constructing a `SearchCriteria` in `\Magento\AdobeStockImage\Model\GetImageList::execute`

PR #303, opened 2019-08-06
Move default filter value  from ContentType to GetImageList service

engcom-Golf:
### Description (*)
Move default filter value setting from ContentType filter to GetImageList service

1. magento/adobe-stock-integration#299: Move default filter value setting from ContentType filter to GetImageList service
2. ...

Filter Image by subCategory


coderimus:
Hello @Nazar65  ,

Thank you for managing this issue and the provided solution. I have an idea, which possibly, can make implementation more flexible and in the "Magento Adobe Stock Integration" style :)

The idea is in moving the default filters to the di and provide them as parameters to the `GetImageList` constructor. This approach can give us the next benefits:
* we avoid hardcode at the `GetImageList` class
* "teach" the `setDefaultFilters` method to build default filter group by iterate data getting from di
* provide an ability to change parameters through the configuration

For example,
```xml
    <type name="Magento\AdobeStockImage\Model\GetImageList">
        <arguments>
            <argument name="defaultFilters" xsi:type="array">
                <item name="content_type_filter" xsi:type="array">
                    <item name="condition" xsi:type="string">or</item>
                    <item name="field" xsi:type="string">illustration</item>
                </item>
                <item name="content_type_filter" xsi:type="array">
                    <item name="condition" xsi:type="string">or</item>
                    <item name="field" xsi:type="string">photo</item>
                </item>
            </argument>
        </arguments>
    </type>
```

What do you think about this?

@sidolov @sivaschenko

engcom-Golf:
Hi @coderimus thank you for your proposal, i'm completely agree with you :)

engcom-Golf:
@coderimus done, but not as i wanted, because `filterBuilder` have method `getData` which returns all data, but method `setData()` needs to be specified key, so there is no ability to set data as `setData(defaultFilters);`
But now its more extensible) thanks.

AdobeStockImage/Model/GetImageList.php
```diff
@@ -23,21 +25,77 @@ class GetImageList implements GetImageListInterface
...
+        }
+        if (!$isContentTypeFilter) {
+            foreach ($this->defaultFilters as $filter) {
+                $filters[] = $this->filterBuilder
+                    ->setField($filter['type'])
+                    ->setConditionType($filter['condition'])
+                    ->setValue($filter['field'])
+                    ->create();
+            }
+            $searchCriteria->setFilterGroups([$this->filterGroupBuilder->setFilters($filters)->create()]);
```
sivaschenko: The existing filter groups will be lost after this set operation

sivaschenko (approved):

Issue #277, opened 2019-08-02
Save keywords asset data along with the asset.


coderimus:
**DESCRIPTION**
An asset object also contains the keyword data which should be stored as well as other asset data components on evert asset save action.

**IMPLEMENTATION NOTES**
Implement the keyword repository and related classes for managing base save/get data operation.
Save asset keywords after the asset save. To do this implement keyword save in the scope of asset repository save action. Not after save asset flow.

PR #304, opened 2019-08-06
277: save asset keywords + provide keywords repository

coderimus:
### Description
This PR is a provided solution for issue [277: Save keywords asset data along with the asset.](https://github.com/magento/adobe-stock-integration/issues/277)

This implementation has 2 main parts:
* save keywords along with the asset object by using latest Magneto 2 approach: use resource services for data storage manipulation to avoid as much inheritance as possible. Also, save keywords action provided as a multiply items save process. As an example, the Inventory module was taken as a guide.
* Keyword repository functionality added because of possible keywords management at the Magento side or service requests or webapi requests. It also implemented with some new features: use resource services (command classes) to proxy repository requests. Here, as I can see, it is not possible at this moment to completely avoid using Abstract DB classes in the inheritance because it does 2 main things, from my point of view: manage load process and before/after save events.

**The save keywords workflow**: save asset and in the save method of the asset repository run save keywords action. In it, as a first step, we see what keywords were assigned to the asset and select them from the keywords storage. Then we compare results and unset not unique values. Act like this to avoid saving not consistent data.

**QUESTIONS**
Is it a required state for a keyword object to implement extension attributes? For the getList and get asset actions?

N/A

### Manual testing scenarios
Save asset and check keywords data.


coderimus:
@sivaschenko about unit tests: will make a separate issue for this after the architect and implementation approve. The tests are have to be complex for this part of the project and do not want to spend a lot of resources on them before approve what to test :)

AdobeStockAssetApi/Api/KeywordRepositoryInterface.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    /**
+     * Save Keyword data
+     *
+     * @param \Magento\AdobeStockAssetApi\Api\Data\KeywordInterface $keyword
+     * @return int
+     * @throws \Magento\Framework\Validation\ValidationException
+     * @throws \Magento\Framework\Exception\CouldNotSaveException
+     */
+    public function save(\Magento\AdobeStockAssetApi\Api\Data\KeywordInterface $keyword): int;
```
sivaschenko:
```suggestion
    public function save(\Magento\AdobeStockAssetApi\Api\Data\KeywordInterface $keyword): void;
```

AdobeStockAssetApi/Api/Data/KeywordInterface.php
```diff
@@ -8,16 +8,19 @@
...
 {
-    const ID = "id";
-    const KEYWORD = "keyword";
+    const ID = 'id';
+
+    const KEYWORD = 'keyword';
+
+    const ASSET_KEYWORD_ASSET_ID = 'asset_id';
+
+    const ASSET_KEYWORD_KEYWORD_ID = 'keyword_id';
```
sivaschenko: I don't think the last two constants make sense in the keyword scope

AdobeStockAsset/Model/ResourceModel/Keyword/SaveMultiplyAndAssignToAsset.php
```diff
@@ -0,0 +1,183 @@
...
+     * @param AssetInterface $asset
+     * @return void
+     *
+     * @throws CouldNotSaveException
+     */
+    public function execute(AssetInterface $asset): void
+    {
+        try {
+            $keywords = $asset->getKeywords();
+            $assignedKeywords = $this->getAlreadyAssignedKeywords($asset);
```
sivaschenko: I don't think there is any need to know which keywords are already assigned and which are not during the save operation.

The save keyword operation may be as simple as:
`insert ignore into adobe_stock_keyword (keyword) values ('apple');`
`insert ignore into adobe_stock_asset_keyword (keyword_id, asset_id) values (24, 42);`

AdobeStockAsset/Model/ResourceModel/Keyword/SaveMultiplyAndAssignToAsset.php
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Assign keywords to asset.
+     *
+     * @param AssetInterface $asset
+     * @param array          $keywords
+     */
+    private function assignKeywordsToAsset(AssetInterface $asset, array $keywords)
+    {
+        if (!empty($keywords)) {
+            $keywordIds = $this->selectKeywordsIdsByNames($keywords);
```
sivaschenko: I don't think there is any need to know keyword ids during save/assign operation

The save keyword operation may be as simple as:
`insert ignore into adobe_stock_keyword (keyword) values ('apple');`
`insert ignore into adobe_stock_asset_keyword (keyword_id, asset_id) values (24, 42);`

AdobeStockAsset/Model/AssetRepository.php
```diff
@@ -86,6 +96,7 @@ public function __construct(
     public function save(AssetInterface $asset): void
     {
         $this->resource->save($asset);
+        $this->saveMultiplyAndAssignToAsset->execute($asset);
```
sivaschenko: As there is a keyword repository - it should be responsible for saving keywords, all extra entry points to save keyword action should be removed to have a single entry point.

```suggestion
        $this->keywordRepository->save($assetId, $asset->getKeywords());
```

AdobeStockAsset/Model/KeywordRepository.php
```diff
@@ -0,0 +1,93 @@
...
+use Magento\AdobeStockAsset\Model\Keyword\Command\SaveInterface;
+use Magento\AdobeStockAssetApi\Api\Data\KeywordInterface;
+use Magento\AdobeStockAssetApi\Api\Data\KeywordSearchResultsInterface;
+use Magento\AdobeStockAssetApi\Api\KeywordRepositoryInterface;
+use Magento\Framework\Api\SearchCriteriaInterface;
+
+/**
+ * @inheritdoc
+ */
+class KeywordRepository implements KeywordRepositoryInterface
```
sivaschenko: There should be a single entry point to operations with keywords. Either a repository or separate services but not both. I'd go with repository not to overengineer the trivial keyword operations

sivaschenko (changes requested):
Hi @coderimus thanks for your pull request, please see my review comments

sivaschenko (approved):

sivaschenko:
@coderimus the repository was removed from the pull request as there are concerns about the implementation (the keyword implementation was partially based on standard resource model and partially on the custom resource model) approach and it is not needed currently. Let's discuss the changes and approach to the repository.

Anyway, this pull request is a great job! I appreciate your thoughtful and wide implementations!

Issue #282, opened 2019-08-02
Allow image entity to have multiple image paths


sivaschenko:
The same image from adobe stock can be saved to multiple folders in media gallery.

To support this behaviour image entity should have the ability to have multiple paths.

For this purpose:
 -`path` field of `adobe_stock_asset` table should be removed
 - `adobe_stock_asset_path` table should be added with many-to-one relation to `adobe_stock_asset`
 - three columns are necessarry for `adobe_stock_asset_path` table: int `id`, int `asset_id` and text `path`. `asset_id` foreign key to `adobe_stock_asset.id`, `path` should be unique

sivaschenko:
This issue is currently under discussion, let's temporary suspend the work on it

PR #306, opened 2019-08-07
Issue 282 allow image entity to have multiple image paths

serhiyzhovnir:
### Description 
This PR allows an image entity to have multiple image paths.

### Fixed Issues

1. magento/adobe-stock-integration#282: Allow image entity to have multiple image paths

### Manual testing scenarios 
1. Save/Load/Delete the `AssetInterface` entity with multiple image paths

### DIFF
--- a/AdobeStockAsset/Model/Asset.php
+++ b/AdobeStockAsset/Model/Asset.php
@@ -165,17 +165,17 @@ class Asset extends AbstractExtensibleModel implements AssetInterface
     /**
      * @inheritdoc
      */
-    public function getPath(): ?string
+    public function getPaths(): ?array
     {
-        return (string) $this->getData(self::PATH);
+        return $this->getData(self::PATHS);
     }
 
     /**
      * @inheritdoc
      */
-    public function setPath(string $path): void
+    public function setPaths(array $paths): void
     {
-        $this->setData(self::PATH, $path);
+        $this->setData(self::PATHS, $paths);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


coderimus:
Hello @serhiyzhovnir ,

Thank you for the well-organised solution provided in this PR. I have one comment which I would like about to talk.

Provided implementation manages to save asset image paths at the asset resource level. This way means that the business logic operates on a low level, which is responsible only for persistence data manipulation management. As I understood this done because of 2 possible decisions:

1. "Paths" is not an independent part of the asset entity and should be saved/deleted right with the asset save/delete action.

2. Resource model allows handling after/before actions which in this case looks like a possible solution: trigger save/delete path operation at the asset object resource manipulation.

But this leads to the next problems: inheritance in use and manage business logic on a low level.

I can recommend taking the next suggestions as an improvement:

1. Move the save/delete asset paths functionality to the separate resource services which will be responsible for those operations. Those services should use connection service in injection. This approach allows minimising inheritance and provides the ability to make save/delete process more transparent.

2. Asset repository should be responsible for asset paths related functionality. The asset repository save and delete methods should call the asset paths save/delete action.

3.*** Use extension attributes (or asset collection) to add paths data to the asset object.

*** This point is the most tricky here. 
@sivaschenko @sidolov

sivaschenko:
@serhiyzhovnir @coderimus The issue is currently under discussion, let's temporary suspend the work on it

sivaschenko:
Thanks for the pull request @serhiyzhovnir ! However, the decision was made to change the approach to allow only one image path and implement "Locate saved image" functionality instead.

See https://github.com/magento/adobe-stock-integration/issues/312 for details

Sorry for this, I shouldn't have created an issue before the final decision was made. However, please pay attention that issues on [our project board](https://github.com/magento/adobe-stock-integration/projects/1) in "Ready for Grooming" status might not have been finalized. In case you'd like to work on the issue in that status anyway - you can always contact me in slack ;)

Unfortunately, I have to close this pull request considering the above. Sorry.

Issue #293, opened 2019-08-03
Use custom services to save entities to the database instead of resource models


sivaschenko:
Use custom services to save entities to the database instead of resource models

Services should use `\Magento\Framework\App\ResourceConnection` for accessing database

Details TBD

PR #311, opened 2019-08-07
Use custom services to save entities to the database instead of resource models

engcom-Golf:
### Description (*)
Use custom services to save entities to the database instead of resource models.

Uses resource services for data storage manipulation. Also, save  action provided as a multiply items save process. 

1. magento/adobe-stock-integration#293: Use custom services to save entities to the database instead of resource models

1.Try to save Preview image.


coderimus:
@engcom-Golf @sidolov @sivaschenko @okorshenko I think that this pr is dependent on [#330 the issue about delete some columns from asset table](https://github.com/magento/adobe-stock-integration/issues/330) Maybe we should deliver it first and then based on the result update provided implementation in this pr?

sivaschenko:
@engcom-Golf @Nazar65 @coderimus I think we should remove database columns first and then adjust this PR to avoid the need to create 3rd PR to adjust the resource models to new DB schema

coderimus:
@Nazar65 @engcom-Golf @sivaschenko 
Ok, cool. I will provide requested data base changes with code adjustments in the nearest 2 days.

engcom-Golf:
@coderimus @sivaschenko well i'm updated PR to the latest changes.

sivaschenko (approved):

Issue #285, opened 2019-08-02
Remove unnecessary conversions of asset when loading it from API


sivaschenko:
Remove unnecessary conversions of the asset when loading it from API

### Displaying assets in the grid UI

| Class::method                                                               | Type before conversion                              | Type after conversion                               | Comment |
|-----------------------------------------------------------------------------|-----------------------------------------------------|-----------------------------------------------------|---------|
| \Magento\AdobeStockClient\Model\Client::search                              | \AdobeStock\Api\Models\StockFile                    | \Magento\Framework\Api\Search\DocumentInterface     |         |
| \Magento\AdobeStockImage\Model\GetImageList::execute                        | \Magento\Framework\Api\Search\DocumentInterface     | \Magento\AdobeStockAssetApi\Api\Data\AssetInterface |         |
| \Magento\AdobeStockImageAdminUi\Model\Listing\DataProvider::getSearchResult | \Magento\AdobeStockAssetApi\Api\Data\AssetInterface | \Magento\Framework\Api\Search\DocumentInterface     |         |

### Retrieving asset information for save operation:

| Class::method                                                               | Type before conversion                              | Type after conversion                               | Comment |
|-----------------------------------------------------------------------------|-----------------------------------------------------|-----------------------------------------------------|---------|
| \Magento\AdobeStockClient\Model\Client::search                              | \AdobeStock\Api\Models\StockFile                    | \Magento\Framework\Api\Search\DocumentInterface     |         |
| \Magento\AdobeStockImage\Model\GetImageList::execute                        | \Magento\Framework\Api\Search\DocumentInterface     | \Magento\AdobeStockAssetApi\Api\Data\AssetInterface |         |

## Solution

Replace `\Magento\AdobeStockImageAdminUi\Model\Listing\DataProvider` with custom implementation of `\Magento\Framework\View\Element\UiComponent\DataProvider\DataProviderInterface` that works with `\Magento\AdobeStockAssetApi\Api\Data\AssetInterface`

diazwatson:
@sivaschenko ready to discuss this when you ready :)

sivaschenko:
@diazwatson the description has been updated with the details

diazwatson:
Has dependency on #298

PR #325, opened 2019-08-07
#285 Remove unnecessary conversions of asset when loading it from API

diazwatson:

1. magento/adobe-stock-integration#285 : Remove unnecessary conversions of asset when loading it from API

1. 
```
vendor/bin/phpunit --configuration dev/tests/unit/phpunit.xml.dist vendor/magento/module-adobe-stock-image-admin-ui/Test/Unit
```
2. 
```
vendor/bin/phpunit --configuration dev/tests/static/phpunit.xml.dist dev/tests/static/testsuite/Magento/Test/Php/LiveCodeTest.php
```


sivaschenko:
Hi @diazwatson , how are things going with this implementation? Feel free to ping me in slack if you need any help

diazwatson:
@sivaschenko please have a look when you get a chance

AdobeStockImage/Model/SaveImagePreview.php
```diff
@@ -208,6 +227,21 @@ private function getImagesByAdobeId(int $adobeId): AssetSearchResultsInterface
             ->setSortOrders([])
             ->create();
 
-        return $this->getImageList->execute($searchCriteria);
+        $searchResult = $this->getImageList->execute($searchCriteria);
+
+        $items = [];
+        /** @var Document $item */
+        foreach ($searchResult->getItems() as $item) {
```
sivaschenko: As finally only one asset is needed here, the algorithm can be simplified. only the first item can be converted and returned if it exists (no need for the foreach loop and searchResult creation).

In this case the method should be renamed to `getImageByAdobeId` and return AssetInterface.

`execute` method should be adjusted accordingly

sivaschenko (approved):

PR #333, opened 2019-08-09
Refactor and clean up the comment model

coderimus:
### Description
Refactor and clean up of the model comment. As a core improvement of this implementation, I provide using the Phrase for comment text translation, code style minor updates and the URLs fixes to exclude the key data.

The comment appearance is
image

N/A

Check that the comment is represented correctly according to original issue description.


coderimus:
@sivaschenko you asked is it possible to get an admin URL without a secret key. It is possible, but I think that the way we can do it for this particular situation is not so suitable. The `getUrl` method generates an URL and adds a secret key to it based on config settings. [Check this link to for this](https://github.com/magento/magento2/blob/2.3-develop/app/code/Magento/Backend/Model/Url.php#L211) So, to avoid adding a secret key we can use directly the URL Magento framework model. But this looks too complicated for our needs. What do you think about this?

sivaschenko (approved):

Issue #326, opened 2019-08-07
Permanent connection error message in system configuration if API key is not valid


sivaschenko:
Permanent connection error message in system configuration if API key is not valid

PR #336, opened 2019-08-09
326: add permanent error message on invalid API key

coderimus:
### Description (*)
According to the issue description, the error message that signalised about incorrect connection has to be displayed permanently.

I decided to implement this in the following way: at the block which renders data for the test connection button template, the connection is tested and result provided to the js plugin. If the connection failed, the error message shown.

The validate-api-key js code also refactored.

#326 

1. Set incorrect API key
2. Save setting page
3. Check that the error message appears and can not be closed.


sivaschenko:
@coderimus I moved the implementation to UI component to simplify bindings

sivaschenko (approved):

PR #342, opened 2019-08-12
Cleanup and code standards adjustments

coderimus:
### Description (*)
This PR provides some code clean up changes like remove logger from the test connection controller (we do no log anything in it). Return type hinting together with parameters type adjustments.

N/A

N/A


AdobeStockAsset/Model/ResourceModel/Keyword.php
```diff
@@ -33,25 +34,34 @@ class Keyword
      */
     private $keywordFactory;
 
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
```
sivaschenko: I don't see logger usages in this file, why should it be added?

sivaschenko (changes requested):
Thanks for adjustments @coderimus ! Just one question in the comments

AdobeStockAsset/Model/ResourceModel/Keyword.php
coderimus: Oh, I missed it. This is a mistakenly adde stuff. Will clean up it in a hour.

sivaschenko (approved):

Issue #340, opened 2019-08-12
Cover with unit tests the keyword save and get functionality


coderimus:
Cove with the unit tests the save and get asset keyword functionality. Integration tests coverage.

sivaschenko:
Not actual anymore. Separate tasks created for each class unit test coverage.

PR #366, opened 2019-08-16
Integration tests: save asset keyword

coderimus:
### Description (*)
This PR provides integration tests coverage for asset keyword entity. In the scope of this PR, next provided:

#### Implementation notes:
next fixtures provided:
`assets_api.php` - provides saved asset object.
`objects/asset.php` - returns an asset object (not saved)*
`objects/category.php` - returns a category object (not saved)*
`objects/creator.php` - returns a creator object (not saved)*
`objects/keywords.php` - returns an array of keywords objects (not saved)*

`*` - objects are not saved in fixtures because we need them to be collected during integration tests to replicate the complex object save as it is. And for the sharing purpose - reuse them in other Adobe Stock Integration tests.

#### №1 Save asset keyword insert ignore test
This test covers the Keyword resource class functionality. Even though this class is not the @api, I decided to cover it with the integration test because it has a specific behaviour: insert ignore. This way to insert data is a part of the save asset object logic and seems to be good to have it tested to guaranty that keyword kept as expected.

#### №2 Save asset keyword along with the asset
As a first step, generate assets parts (category, creator, keywords) and assign them to the asset. Save an asset object. Save keywords. Get saved asset keywords by asset id and compare them with what provided initially with the asset object.

#### №3 Get keywords by asset
`asset.php` fixture provides saved asset data. By its id, the test method gets keywords and compare them with the one set to the asset before.

### Refactoring and additions
1. change the `asset.php` fixture to the `assets_for_api.php` - we will need a separate fixture for an asset object consists of different data rather than for API test.
1. provide asset fixture along with the keyword and others

#340 Cover with unit tests the keyword save and get the functionality

N/A


AdobeStockAsset/Test/Api/AssetRepository/SaveTest.php
```diff
@@ -137,7 +137,7 @@ private function saveAsset(array $data)
                 'operation' => self::SERVICE_NAME . self::SERVICE_OPERATION,
             ],
         ];
-        $requestData = ['asset' => $data];
+        $requestData = ['asset' => $data,];
```
sivaschenko, line 128: Such commas just do not look right. I think it's better to disable this rule in the IDE.

AdobeStockAsset/Test/Integration/GetAssetKeywordByAssetIdTest.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        $result = !array_diff($originalData, $savedData);
+        $this->assertTrue($result);
+    }
+
+    /**
+     * @return AssetInterface[]
+     */
+    public static function testGetAssetKeywordByIdDataProvider(): array
```
sivaschenko, line 69: Please do not use `test` prefix for data providers. It is reserved for tests by PHPUnit. As the provider is used to provide assets, the proper name would be `assetsProvider`
```suggestion
    public static function assetProvider(): array
```

AdobeStockAsset/Test/Integration/GetAssetKeywordByAssetIdTest.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        $originalData = [];
+        $originalKeywords = $asset->getKeywords();
+        /** @var KeywordInterface $original */
+        foreach ($originalKeywords as $original) {
+            $originalData[] = $original->getKeyword();
+        }
+
+        $result = !array_diff($originalData, $savedData);
```
sivaschenko, line 62: please use an assert to compare the expected and the actual result, also it would be good to use the Magento fixtures engine (`@magentoDataFixture` annotation) and static values as the expected result.

AdobeStockAsset/Test/Integration/SaveAssetKeywordInsertIgnoreTest.php
```diff
@@ -0,0 +1,91 @@
...
+    private $resourceConnection;
+
+    /**
+     * @inheritdoc
+     */
+    protected function setUp()
+    {
+        $this->keywordResourceModel = Bootstrap::getObjectManager()->get(Keyword::class);
+        $this->resourceConnection = Bootstrap::getObjectManager()->get(ResourceConnection::class);
+        $this->resourceConnection->getConnection()->beginTransaction();
```
sivaschenko, line 35: There is no need to use a transaction for rollback as the populated data is known.

Please use @magentoDataFixture to specify a fixture. The fixture file name + `_rollback` suffix will identify the file containing a rollback algorithm for corresponding fixture

AdobeStockAsset/Test/_files/api_assets.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
sivaschenko, line 1: I don't think we need this fixture as there is more advanced `AdobeStockAsset/Test/_files/asset.php` introduced

AdobeStockAsset/Test/_files/objects/category.php
```diff
@@ -0,0 +1,23 @@
+<?php
```
sivaschenko, line 1: I think that having additional `objects` folder is excessive here, can we get rid of it and have fixtures simply like `_files/category.php`, `_files/keywords.php` etc?

AdobeStockAsset/Test/_files/objects/asset.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
sivaschenko, line 1: I don't think we need this fixture as there is more advanced `AdobeStockAsset/Test/_files/asset.php` introduced

sivaschenko (changes requested):
Hi @coderimus thanks for the pull request! Sorry, for the late review.

Please take a look at my comments and see Travis test results.

sivaschenko:
@coderimus are there any updates on this pull request? Do you need any help to complete it? Can you please also sign the Adobe CLA.

coderimus:
Hi @sivaschenko,
I am working on finalising this PR: currently, I stack a little with the fixture generation for the asset save the test. If I will not fix this in 2 days, I will push changes to this PR and will request your assistance.

sivaschenko:
Depends on https://github.com/magento/adobe-stock-integration/pull/515

PR #367, opened 2019-08-16
Redirect URI pattern with escaped hostname

rogyar:
### Description (*)
Adobe I/O console requires to have the periods (.) in the hostname escaped for the "Redirect URl pattern".
 
image

Currently, for the "Redirect URL pattern" the system returns the same value as for the "Redirect URL". You need to modify the value manually by escaping the periods in the hostname when you paste the pattern value on the Adobe I/O console side. 

This PR brings a change that escapes the periods from hostname of the "Redirect URL pattern" automatically, so the user can easily copy the value from admin panel and paste to Adobe I/O without additional manual edit. 

Please note that only the hostname part must be escaped. 

N/A

1. Go to Admin -> Stores -> Configuration -> System -> Adobe Stock Integration
2. Make sure that values for Redirect URI and Pattern are not the same and the periods in the hostname in the Pattern are escaped.

image


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/adobe-stock-integration?pullRequest=367) <br/>All committers have signed the CLA.

AdobeStockAdminUi/Model/System/Config/Comment.php
```diff
@@ -89,6 +89,12 @@ private function getRedirectUrl(): string
      */
     private function getRedirectUrlPattern(): string
     {
-        return $this->getRedirectUrl();
+        $url = $this->getRedirectUrl();
+
+        /* Escape dots in the hostname */
+        $hostName = parse_url($url, PHP_URL_HOST);
+        $escapedHostName = str_replace('.', '\\\.', $hostName);
```
sivaschenko: I think all periods have to be escaped with `\\`, so there is a need to extract hostname before replacing. Also:
```suggestion
        $escapedHostName = str_replace('.', '\\.', $hostName);
```

sivaschenko (changes requested):
Welcome to the project and thanks for the fix @rogyar !
Please let me know what do you think regarding my comment.

AdobeStockAdminUi/Model/System/Config/Comment.php
rogyar: Basically, the console allows periods in other parts of the pattern (disallowed only for the hostname). But I believe if we escape everywhere, it should still work. I will double-check this point.
As for the suggestion, we need three backslashes there since we need to make the string processor know that one of our backslashes it's exact a backslash but not the escape character. I.e.
```
'that\'s a string`
`that\'s a string with the backslash -> \\'
```

sivaschenko: You are right. Yes, it's interesting how the string processor works, because both of backslashes are expected to be exactly backslashes in the result string. Still only one of them needs escaping.

sivaschenko (dismissed):

sivaschenko (approved):

PR #385, opened 2019-08-25
373: image related image series - base implementation

coderimus:
### Description (*)
This PR is the first initial implementation for the image series functionality. PR provides the back-end part along with the basic frontend implementation: ajax async call with data managing.

#### How it works?
The preview image block has a new element - <div> "More from this series". On its render on preview image activate the getImageSeries function executes and requested for the image related series. On success call of this request, the image list is returned and rendered.

The Image series is the list of assets filtered by the image id. To get it a new filter introduced - the Series. 
It handles the `serie_id` parameter. Serie id is the image adobe id.

#### Questions
1. Should we store the image series returned value per image in the observable array and return it on the second image preview attempt? Or request the image series every image preview section visit?
1. What value should we use as the default image preview item limit to render? 

N/A

N/A


AdobeStockClient/Model/SearchParametersProvider/Series.php
```diff
@@ -0,0 +1,36 @@
...
+namespace Magento\AdobeStockClient\Model\SearchParametersProvider;
+
+use AdobeStock\Api\Models\SearchParameters;
+use Magento\AdobeStockClientApi\Api\SearchParameterProviderInterface;
+use Magento\Framework\Api\SearchCriteriaInterface;
+
+/**
+ * Adobe image series filter.
+ */
+class Series implements SearchParameterProviderInterface
```
sivaschenko: Looks like there is no required custom logic in this parameter provider. Can series be included as part of simple filters?

AdobeStockImage/Model/GetImageSeries.php
```diff
@@ -0,0 +1,90 @@
...
+    public function execute(int $serieId): array
+    {
+        try {
+            $searchCriteria = $this->searchCriteriaBuilder->addFilter('serie_id', $serieId)
+                ->setSortOrders([])
+                ->create();
+            $items = $this->getImageList->execute($searchCriteria)->getItems();
+
+            $assets = [];
+            foreach ($items as $item) {
```
sivaschenko: Is it necessary to convert series items to asset interface? If not, it's ok to operate `DocumetInterface`s. I don't think there is any chance they will be saved to the database.

AdobeStockImage/Model/ImageSeriesSerialize.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    /**
+     * Serialize image series data.
+     *
+     * @param AssetInterface[] $series
+     *
+     * @return array
+     * @throws SerializationException
+     */
+    public function execute(array $series): array
```
sivaschenko: I'd propose to use the existing hydrator to extract data from `DocumentInterface`

sivaschenko (changes requested):
Hi @coderimus thanks for the pull request! Great work! The only thing I can suggest - is reducing the amount of code by reusing `SimpleFilters` class DI configuration and reducing the number of conversions by working with `DocumentInterface` instead of `AssetInterface`. Please see my review comments!

coderimus:
@sivaschenko thank you for your review and suggested improvements. I think I got the point: the image series is just a simple data which is needed only to rendered at the specific place. And it doesn't have any other role in the system. According to this, I made the next changes:

1. represent the series filter as a simple filter
1. allow the SimpleFilters set int type variables (previously, escaper transform the int into the string when calling the set filter value method)
1.  move image series serialise logic to the GetImageSeries model and make it works with Document instead of Asset

Please, review the latest comment to this branch.

sivaschenko (approved):

Issue #359, opened 2019-08-14
Locate saved preview image in Magento media gallery


sivaschenko:
### Story

https://github.com/magento/adobe-stock-integration/issues/312 [Story 24] User locates licensed and uploaded image inside Media Gallery

### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
3. Click on any image to expand the image preview
4. Click "Save Preview" button
5. Reopen the same image in the Adobe stock grid

### Expected result (*)
"Save Preview" button label is changed to "Locate Image".
Click on "Locate Image" button is closing the Adobe Stock slider panel, opens the directory containing the image in the media gallery and selects this image.

PR #406, opened 2019-08-31
#359: added Locate button for saved preview images

vovsky:
### Description (*)
This PR adds ability to locate downloaded image in a file browser

1. magento/adobe-stock-integration#359: Locate saved preview image in Magento media gallery


magento-cicd2:
[CLA assistant check](https://cla.dev.magento.com/magento/adobe-stock-integration?pullRequest=406) <br/>All committers have signed the CLA.

diazwatson:
Hi @vovsky thanks for your contribution!
Could you please fix the static tests?
```
FILE: ...r/magento/module-adobe-stock-asset/Model/AddIsDownloadedToSearchResult.php
--------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 46 | ERROR | [x] Missing short description
 47 | ERROR | [x] There must be exactly one blank line before tags
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------
```

AdobeStockAsset/Model/AddIsDownloadedToSearchResult.php
```diff
@@ -0,0 +1,80 @@
...
+
+            foreach ($items as $key => $item) {
+                $customAttributes = $item->getCustomAttributes();
+
+                $attribute = $this->attributeValueFactory->create();
+                $attribute->setAttributeCode(self::ATTRIBUTE_CODE_IS_DOWNLOADED);
+                $attribute->setValue((int)in_array($item->getId(), $downloadedIds));
+
+                $customAttributes[self::ATTRIBUTE_CODE_IS_DOWNLOADED] = $attribute;
+                $item->setCustomAttributes($customAttributes);
```
sivaschenko: It would be better to add the downloaded path as a custom attribute and use this attribute to open the corresponding folder in media gallery during `locate` operation.

AdobeStockAsset/Model/AddIsDownloadedToSearchResult.php
```diff
@@ -0,0 +1,80 @@
...
+namespace Magento\AdobeStockAsset\Model;
+
+use Magento\Framework\Api\AttributeValueFactory;
+use Magento\Framework\Api\Search\SearchResultInterface;
+use Magento\Framework\App\ResourceConnection;
+
+/**
+ * Class AddIsDownloadedToSearchResult
+ */
+class AddIsDownloadedToSearchResult
```
sivaschenko: I would be good to use generic name for this service like `AppendAttributes` as later it could be reused to add additional information like `is_licensed` etc

sivaschenko (changes requested):
Hi @vovsky , welcome to Adobe Stock Integration project and thank you for your pull request!

I have an improvement suggestion regarding the attribute passed from backend to frontend to identify that image is downloaded, please take a look at my review comments.

Feel free to contact me on slack for more details

AdobeStockAsset/Model/AddIsDownloadedToSearchResult.php
vovsky: @sivaschenko agree. I have made the changes, thank you. Locate feature works now

AdobeStockAsset/Model/AddIsDownloadedToSearchResult.php
vovsky: @sivaschenko agree. I have made the changes, thank you

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
```diff
@@ -328,6 +342,21 @@ define([
             });
         },
 
+        /**
+         * Locate downloaded image in media browser
+         *
+         * @param record
+         */
+        locate: function (record) {
```
sivaschenko, line 370: that implementation considers only the filename. Images can be saved to different directories in the media gallery.

So after modal is closed it's needed to:
 - Divide the `record.path` to directory path and filename
 - Use directory path to open corresponding folder in media gallery
 - Use filename to select corresponding image

AdobeStockAsset/Model/AppendAttributes.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->attributeValueFactory = $attributeValueFactory;
+    }
+
+    /**
+     * Add additional asset attributes
+     *
+     * @param SearchResultInterface $searchResult
+     * @return SearchResultInterface
+     */
+    public function execute(SearchResultInterface $searchResult): SearchResultInterface
```
sivaschenko, line 62: I'd extract a couple of private functions to make this class easier to read

sivaschenko (changes requested):
@vovsky thanks for updates! Looks great! However, there is a specific case that is not handled by the implemented locate functionality. Please take a look at my comments

sivaschenko:
Hi @vovsky can you please sign the Adobe CLA

filmaj:
FYI this can be done at opensource.adobe.com/cla.html

AdobeStockAsset/Model/AppendAttributes.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $items = $searchResult->getItems();
+        $itemIds = [];
+
+        foreach ($items as $key => $item) {
+            $itemIds[$key] = $item->getId();
+        }
+
+        if (count($itemIds)) {
+            $connection = $this->resourceConnection->getConnection();
```
sivaschenko: Can you please move this select from `adobe_stock_asset` table to a separate class to be able to reuse that, and place this class in the `ResourceModel` namespace in order to indicate the access to the database. I'd suggest moving to `\Magento\AdobeStockAsset\Model\ResourceModel\Asset\Load::execute(array $ids): array`

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
```diff
@@ -332,6 +348,52 @@ define([
...
+            $.ajaxSetup({async: false});
```
sivaschenko: Can you please extract this method to a separate JS file (`image-preview.js` is becoming a really long file and we are trying to limit it's responsibility to image preview rendering only).

I'd create `AdobeStockImageAdminUi/view/adminhtml/web/js/media-gallery.js` returning simple object:

```
define([
...
], function (...) {
        'use strict';

        return {
            jsTreeRootFolderName: 'Storage Root',
            jsTreeFolderNameMaxLength: 20,
           selectImage: function (path) {
              ...
           }
        }
    }
```

AdobeStockAsset/Model/AppendAttributes.php
sivaschenko: It would also be great if you could cover these files with unit tests

sivaschenko (changes requested):
Hi @vovsky , thanks for the updates! Please see my review comments.

sivaschenko (dismissed):

sivaschenko (approved):

Issue #397, opened 2019-08-30
"More from this model" image preview section


sivaschenko:
### Story

https://github.com/magento/adobe-stock-integration/issues/324 [Story 30] User sees images with the same model as the currently viewed one

### Preconditions
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Click on any image

### Expected result
Image preview is expanded, "More from this model" section with the same series images is displayed (possible to switch tabs between "More from this series" and "More from this model")

![image](https://user-images.githubusercontent.com/2028541/63356243-5b256e80-c35f-11e9-8ebc-29d1686e7da6.png)

### Technical Details

 - Update configuratioin of  `\Magento\AdobeStockClient\Model\SearchParametersProvider\SimpleFilters` to introduce the `search_parameters[model_id]` request parameter
 - I think `\Magento\AdobeStockImageAdminUi\Controller\Adminhtml\Preview\Series` controller can be reused to load same model images as well (do reduce number of requests to server)

konarshankar07:
@sivaschenko ,
Sorry, I mistakenly assign to me can you please put back in ready for development.
Thanks

konarshankar07:
Please ignore the above comment. I'll start working on this ticket
Thanks

PR #408, opened 2019-09-01
#397 : "More from this model" image preview section

konarshankar07:
### Description (*)

This PR is regarding addition of more from this model section in the image preview

1. magento/adobe-stock-integration#397: "More from this model" image preview section.

1. Login to admin panel
4. Click on any image
### Actual Result
image
image


AdobeStockImage/Model/GetRelatedImages.php
```diff
@@ -17,10 +17,20 @@
...
- * Class GetImageSeries
+ * Class GetRelatedImages
  */
-class GetImageSeries
+class GetRelatedImages
 {
+    /*
+     * Series ID.
+     */
+    const SERIE_ID = 'serie_id';
```
sivaschenko: To avoid code duplication, it's better to use the field as a method parameter or as a constructor parameter provided by DI and use cycles instead of similar code blocks

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
```diff
@@ -199,21 +201,45 @@ define([
...
+
         /**
          * Returns series to display under the image
          *
          * @param record
          * @returns {*[]}
          */
         getSeries: function (record) {
-            if (!record.series) {
+            if (!record.series || !record.model) {
```
sivaschenko: it's better to process series and same model images separately (i.e. use a loop)

sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request! Please take a look at the review comments

AdobeStockImage/Model/GetRelatedImages.php
konarshankar07: Hello @sivaschenko ,

We can avoid code duplication by replacing `Magento\Framework\Api\Search\SearchCriteriaBuilder` to `Magento\Framework\Api\SearchCriteriaBuilder` so that we can pass array of filter but we have already used `Magento\Framework\Api\Search\SearchCriteriaBuilder` in so many places that there are lot of chance some functionality might broke. Let me know if you want to move forward with this logic or any other logic from your side is always welcome

Thanks!

AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/image-preview.html
```diff
@@ -58,23 +58,44 @@ <h3 class="title" text="$col.getTitle($row())" if="$col.getTitle($row())"/>
...
             <button data-bind="visible: $col.canViewMoreKeywords($row()), click: function(){ $col.viewAllKeywords($row()) }">
                 <span translate="'View all'"/>
             </button>
         </div>
-        <div class="series">
-            <div class="title" translate="'More from this series'"></div>
-            <!-- ko foreach: $col.getSeries($row()) -->
-            <div class="thumbnail" data-bind="css: { 'hide': $index() >= 3}">
-                <img attr="src: thumbnail_url, alt: title">
+        <div id="adobe-stock-tabs" attr="'data-mage-init': getDataMageInit()">
```
sivaschenko: That is incorrect usage of `data-mage-init` attribute. I think `data-mage-init` attribute should be static in this case.

AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/image-preview.html
```diff
@@ -58,23 +58,44 @@ <h3 class="title" text="$col.getTitle($row())" if="$col.getTitle($row())"/>
...
+            <ul class="tabs-horiz">
+                <li>
+                    <a href="#series_content" id="series_tab">
```
sivaschenko: As there are multiple tabs per page, I'd suggest using unique ids (i.e. appending image id or index)

sivaschenko (changes requested):
Hi @konarshankar07 thanks for updates.

Please take a look at my commit with removed code duplication and see a couple of additional comments in this code review.

Also please pay attention that `GetImageSeriesTest` has to be adjusted

AdobeStockImage/Model/GetRelatedImages.php
```diff
@@ -81,22 +79,18 @@ public function __construct(
...
-            $modelSearchCriteria = $this->searchCriteriaBuilder
-                ->addFilter($modelFilter)
-                ->setPageSize($limit)
-                ->create();
-
-            return $this->serializeRelatedImages(
-                $this->getImageList->execute($serieSearchCriteria)->getItems(),
-                $this->getImageList->execute($modelSearchCriteria)->getItems()
-            );
+            foreach ($this->fields as $key => $field) {
```
konarshankar07, line 84: We can still refactor the code using the filter group for better searching. Please see [devdocs](https://devdocs.magento.com/guides/v2.3/extension-dev-guide/searching-with-repositories.html) for more information in order to do that we need to change `SearchCriteriaBuilder` interface which I mention in the previous comment. Correct me if I'm wrong.
Thanks

konarshankar07:
Hello @sivaschenko ,
Thanks for helping me out. Please check my comment on the "removed code duplication" commit. I'll continue to work on other requested changes

AdobeStockImage/Model/GetRelatedImages.php
sivaschenko: Thanks for the suggestion! Do you mean we might have been able to add both filters to a single filter group and perform a single request? I don't think that will be possible considering that it's required to keep the related image groups separate.

konarshankar07: Yes, I think you are right. I tried to implement and failed to achieve the expected result

konarshankar07:
Hello @sivaschenko ,
I'm not able to find the reference file for `GetImageSeriesTest`. Do I need to write the unit testing for `GetImageList` and remove the `GetImageSeriesTest`?
Thanks

sivaschenko:
@konarshankar07 here is the file https://github.com/magento/adobe-stock-integration/blob/develop/AdobeStockImage/Test/Unit/Model/GetImageSeriesTest.php

That test is causing build failure and has to be corrected

konarshankar07:
Hello @sivaschenko ,
I'm not able to check the status for this travis automated status for this PR. Can you please check why its not showing?
Here is the screeshot
image

Thanks

sivaschenko:
Hi @konarshankar07 I believe that is because of conflicts

sivaschenko (approved):

Issue #401, opened 2019-08-30
Licensing an image


sivaschenko:
### Story:

https://github.com/magento/adobe-stock-integration/issues/307: [Story 20] User licenses an image

### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
3. Have an Adobe Stock account with credits: contact @sivaschenko 

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Click on an image to expand the preview
5. Click on "License Image" button
6. Click "Confirm" in the quote confirmation popup

### Expected result (*)
Image is licensed, licensed version of image is saved

### Technical details:

Add a `\Magento\AdobeStockImageAdminUi\Controller\Adminhtml\License\License` backend controller (use client for license API request, use `\Magento\AdobeStockImageApi\Api\SaveImagePreviewInterface::execute` for saving the image file)
Add `licenseImage` method to Client
Use `/Rest/Libraries/1/Content/License?content_id=<media_id>&license=Standard` Adobe Stock API endpoint for retrieving information.
The frontend functionality should be added to the `/js/components/grid/column/image-preview.js.licenseAndSave`

PR #411, opened 2019-09-02
License and save

rogyar:
### Description (*)
This PR introduces image license and save functionality. This PR also contains a set of refactorings since the approach that has been implemented for saving the preview images is not extensible as it is. 

1. #401 : Licensing an image

4. Choose an image
6. Click "Ok" in the confirmation window

The image should be saved to the Magento storage in its original size.


rogyar:
I'm still working on some minor improvements

AdobeStockImage/Model/SaveAsset.php
```diff
@@ -0,0 +1,131 @@
...
+use Magento\AdobeStockAssetApi\Api\Data\AssetInterface;
+use Magento\Framework\Exception\AlreadyExistsException;
+use Magento\Framework\Exception\NoSuchEntityException;
+
+/**
+ * Class SaveAsset
+ *
+ * Service for saving asset
+ */
+class SaveAsset
```
sivaschenko: I think this class can be moved to AdobeStockAsset module

AdobeStockImage/Model/SaveImage.php
```diff
@@ -0,0 +1,85 @@
...
+
+use Magento\AdobeStockImageApi\Api\SaveImageInterface;
+use Magento\Framework\Exception\CouldNotSaveException;
+use Magento\AdobeStockClientApi\Api\ClientInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Class SaveImage
+ */
+class SaveImage implements SaveImageInterface
```
sivaschenko, line 21: Looks like this service is applicable only to licensed images. I would indicate that in the naming. However, if it will accept URL instead of adobeId - this service can be applicable for both licensed and preview images

AdobeStockImageApi/Api/SaveImageInterface.php
```diff
@@ -0,0 +1,34 @@
...
+     *
+     * @param int    $adobeId
+     * @param string $destinationPath
+     *
+     * @return void
+     * @throws CouldNotSaveException
+     * @throws NotFoundException
+     * @throws LocalizedException
+     */
+    public function execute(int $adobeId, string $destinationPath): void;
```
sivaschenko: I think the service accepting URL instead of adobeId could be more universal.

AdobeStockImageAdminUi/Controller/Adminhtml/License/License.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->logger = $logger;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute()
+    {
+        try {
+            $params = $params = $this->getRequest()->getParams();
```
sivaschenko:
```suggestion
            $params = $this->getRequest()->getParams();
```

sivaschenko (changes requested):
Hi @rogyar , thanks for working on this functionality. The implementation looks great so far.

Please see my review comments

AdobeStockImageApi/Api/SaveImageInterface.php
rogyar: That's a great catch. However, we still need adobeID for getting/saving the asset. I decided to pass the `AssetInterface` with URL instead.
In that way, we can eliminate two interfaces and implementations for saving image/preview and use the unified one. 
Please, take a look at the recent commit.

AdobeStockAsset/Model/SaveAsset.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Save asset
+     *
+     * @param AssetInterface $asset
+     * @throws AlreadyExistsException
+     */
+    public function execute(AssetInterface $asset): void
+    {
+        $category = $asset->getCategory();
+        if ($category !== null && !$this->isCategorySaved($category->getId())) {
```
sivaschenko: isSaved checks can be skipped for all entities to speed up the processing (current resource models are using `INSERT IGNORE` operation)

AdobeStockClient/Model/Client.php
```diff
@@ -227,6 +240,64 @@ public function getQuotaConfirmationMessage(int $contentId): string
...
+    }
+
+    /**
+     * Convert a stock file object to a document object
+     *
+     * @param StockFile $file
+     * @return DocumentInterface
+     * @throws IntegrationException
+     */
+    private function convertStockFileToDocument(StockFile $file): DocumentInterface
```
sivaschenko: `\Magento\AdobeStockClient\Model\StockFileToDocument::convert` should be used instead

AdobeStockClientApi/Api/ClientInterface.php
```diff
@@ -56,4 +56,19 @@ public function getFullEntitlementQuota(): UserQuotaInterface;
      * @return bool
      */
     public function testConnection(string $apiKey = null): bool;
+
+    /**
+     * Invokes licensing image operation via Adobe Stock API
+     *
+     * @param int $contentId
+     */
+    public function licenseImage(int $contentId);
```
sivaschenko: return type is missing. It's 'License' in the implementation, however, I think it should be void

AdobeStockImage/Model/GetImageByAdobeId.php
```diff
@@ -0,0 +1,90 @@
...
+use Magento\Framework\Api\FilterBuilder;
+use Magento\Framework\Api\Search\SearchCriteriaBuilder;
+use Magento\Framework\Exception\NotFoundException;
+
+/**
+ * Class GetImageByAdobeId
+ *
+ * Service for getting image by content id
+ */
+class GetImageByAdobeId
```
sivaschenko: This class implementation is not specific to images. It can be moved to AdobeStockAsset module and renamed to `GetAssetById` (adobe_id equals to internal id)

In this case `Magento\AdobeStockAssetApi\Api\GetAssetListInterface`  can be used insgead of `GetImageListInterface`

AdobeStockAsset/Model/SaveAsset.php
```diff
@@ -0,0 +1,126 @@
...
+        }
+        $asset->setCategoryId($category->getId());
+
+        $creator = $asset->getCreator();
+        if ($creator !== null && !$this->isCreatorSaved($creator->getId())) {
+            $creator = $this->creatorRepository->save($creator);
+        }
+        $asset->setCreatorId($creator->getId());
+
+        $this->assetRepository->save($asset);
```
sivaschenko, line 76: Please set `is_licensed` asset attribute to `true`

sivaschenko (changes requested):
Thanks for updates @rogyar . I have several improvement suggestions. Please see my review comments.

AdobeStockAsset/Model/SaveAsset.php
rogyar: Thank you for the suggestion. 
Since the `SaveAsset` service is common for both preview and image, I've set the flag in the corresponding controller once the image is licensed

https://github.com/magento/adobe-stock-integration/pull/411/commits/18ee4077973b4b0c7e41977ca53e9abe0d21f978

sivaschenko (approved):

PR #418, opened 2019-09-05
Added loading animation for license popup

rogyar:
### Description (*)
The license popup requires some time to be loaded because of an additional AJAX request for retrieving the license quota. As a result, there's a delay between clicking the "License and save" button and showing the popup. By adding the loading animation here we let the user now that something is happening and prevent from clicking the button once again

N/A

4. Choose an image

You should see the loading animation before the popup appears.

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
@@ -401,6 +401,7 @@ define([
          */
         showLicenseConfirmation: function (record) {
             var licenseAndSave = this.licenseAndSave;
+            $(this.adobeStockModalSelector).trigger('processStart');
             $.ajax(
                 {
                     type: 'POST',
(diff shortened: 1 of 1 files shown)


sidolov (approved):

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
```diff
@@ -422,7 +424,7 @@ define([
                                     licenseAndSave(record);
                                 }
                             }
-                        });
+                        })
                     },
 
                     error: function (response) {
```
sidolov, line 430: I believe we should stop spinner in case of error too

sidolov (changes requested):

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
rogyar: Hi @sidolov. Absolutely, we already have this part 
https://github.com/magento/adobe-stock-integration/blob/develop/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js#L429

sidolov (approved):

Issue #434, opened 2019-09-18
User quota is not updated during authorization


sivaschenko:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
3. Have an Adobe Stock account with credits: contact @sivaschenko 

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Click on "Sign in" link in top right corner and login with the Adobe Stock account with credits
5. Click on the logged-in user name (that appears instead of "Sign in" link) to show user profile popup

### Expected result (*)
Credits/Images are showing the default value "0", they are updated only after the page is refreshed

### Expected result (*)
Credits/Images should display correct values instantly after login

PR #459, opened 2019-09-23
User quota is not updated during authorization

Nazar65:

1. magento/adobe-stock-integration#434: User quota is not updated during authorization
2. ...

1. Login to admin panel
4. Click on "Sign in" link in top right corner and login with the Adobe Stock account with credits
5. Click on the logged-in user name (that appears instead of "Sign in" link) to show user profile popup

### Expected result (*)
Credits/Images are showing the default value "0", they are updated only after the page is refreshed


AdobeIms/Controller/Adminhtml/User/Profile.php
```diff
@@ -12,6 +12,7 @@
 use Magento\Backend\App\Action;
 use Magento\Framework\Controller\ResultFactory;
 use Psr\Log\LoggerInterface;
+use Magento\AdobeStockClientApi\Api\ClientInterface;
```
sivaschenko: Unfortunately, we cannot use adobe stock modules from `AdobeIms` as it has to be a separate independent package.

I think the solution can be: 
 - moving User data generation out of Controller to a dedicated model and applying a plugin to this model from AdobeStockAsset module.
OR
 - updating the quota via additional asynchronous request after login (this is more preferable than plugin) - additional request should not affect user experience as it should be performed only once after login

sivaschenko (changes requested):
Hi @Nazar65 thanks for the fix. Please take a look at my review comment!

AdobeIms/Controller/Adminhtml/User/Profile.php
Nazar65: ok, i think second way is better. i will try to implement it.

Nazar65:
hi @sivaschenko could you review new changes ?

sivaschenko (approved):

Issue #471, opened 2019-09-24
Move default premium filter to di.xml configuration


sivaschenko:
By default premium filter is set to `All` in `\Magento\AdobeStockClient\Model\SearchParametersProvider\Premium::apply`
Instead of setting the default filter in Client, it should be set by adding the filter to `Magento\AdobeStockImage\Model\GetImageList` in  `AdobeStockImage/etc/di.xml`

PR #492, opened 2019-09-27
Move default premium filter to di.xml configuration

Nazar65:
### Description (*)
By default premium filter is set to All in \Magento\AdobeStockClient\Model\SearchParametersProvider\Premium::apply

Instead of setting the default filter in Client, it now sets by adding the filter to Magento\AdobeStockImage\Model\GetImageList in AdobeStockImage/etc/di.xml

1. magento/adobe-stock-integration#471: Move default premium filter to di.xml configuration
2. ...

1. Configure Adobe Stock integration
2. Open Adobe Stock search
3. Verify premium filter all applied
4. Chose in grid filters filter premium
5. Verify filter applied.


AdobeStockImage/etc/di.xml
```diff
@@ -27,6 +27,11 @@
                     <item name="condition" xsi:type="string">or</item>
                     <item name="field" xsi:type="string">photo</item>
                 </item>
+                <item name="premium_filter" xsi:type="array">
+                    <item name="type" xsi:type="string">premium_price_filter</item>
+                    <item name="condition" xsi:type="string">eq</item>
+                    <item name="field" xsi:type="string">all</item>
```
sivaschenko: I think it's worth to try to use `Magento\AdobeStockClient\Model\SearchParametersProvider\SimpleFilters` for Premium filter. It should be possible after accepting the following suggestion:
```suggestion
                    <item name="field" xsi:type="string">ALL</item>
```

sivaschenko (changes requested):
Thanks for the pull request @Nazar65 ! Please see my code review comments

Nazar65:
@sivaschenko done :heavy_check_mark:

sivaschenko (approved):

Issue #485, opened 2019-09-26
Licensing image. Out of credits message.


sivaschenko:
### Story

https://github.com/magento/adobe-stock-integration/issues/307 [Story #20] User licenses an image

### Preconditions
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Click on any image to expand image preview
5. Click on "License" button
6. Login to adobe stock account that does not have credits

### Expected result
A popup displaying the information about insufficient credit should be shown. The "Buy Credits" button should open https://stock.adobe.com/ in a new tab

![image](https://user-images.githubusercontent.com/2028541/65724637-85362300-e0b9-11e9-8b3c-bba43d36e919.png)

PR #496, opened 2019-09-27
Show alternative license popup view on zero credit

rogyar:
### Description (*)
This PR introduces an alternative popup view for licensing image in case there are no credits left. In this case, user will get a popup that recommends buying more credits and points the user to the corresponding URL

1. #485 : Licensing image. Out of credits message.

4. Choose an image
6. Log into the Adobe Stock account that has no credits
7. You should see a popup that recommends buying more credits

image

8. By clicking on the "Buy Credits" a new tab is being appeared that goes to https://stock.adobe.com/


AdobeStockImageAdminUi/Block/Adminhtml/Panel.php
```diff
@@ -71,4 +71,14 @@ public function getUserQuota(): UserQuotaInterface
             return $userQuota;
         }
     }
+
+    /**
+     * Get URL for buying more credits
+     *
+     * @return string
+     */
+    public function getBuyCreditsUrl(): string
```
rogyar, line 80: I wasn't sure where to put this value

sivaschenko: I think the best place to put it is `AdobeStockAsset/etc/config.xml` (looks like that will require the creation of Config.php and ConfigInterface for AdobeStockAsset/AdobeStockAssetApi module)

AdobeStockClient/Model/Client.php
```diff
@@ -235,9 +235,15 @@ public function getFullEntitlementQuota(): UserQuotaInterface
...
      */
-    public function getQuotaConfirmationMessage(int $contentId): string
+    public function getQuotaInformation(int $contentId): array
     {
-        return $this->getLicenseInfo($contentId)->getPurchaseOptions()->getMessage();
+        $licenseInfo = $this->getLicenseInfo($contentId);
+        $quota = $licenseInfo->getEntitlement()->getQuota();
+
+        return [
+            'message' => $licenseInfo->getPurchaseOptions()->getMessage(),
```
sivaschenko, line 244: It would be good to introduce a `QuotaInterface` for quota and I think we can merge the `Client::getQuotaInformation` method with `ClientInterface::getFullEntitlementQuota` and `ClientInterface::getQuota ` having that data interface in place

sivaschenko (changes requested):
Thanks for the pull request @rogyar ! Please see my improvement suggestions

AdobeStockClient/Model/Client.php
sivaschenko: will be done in https://github.com/magento/adobe-stock-integration/pull/459

sivaschenko (approved):

PR #515, opened 2019-10-07
315: Adobe Media Gallery

coderimus:
### Description (*)
This PR provides a new module `AdobeMediaGallery`. This module is responsible for storing the media data such as media asset (path, title, content type), media asset keywords and their link to the media asset.

Currently, it is under construction, and its structure is possible to be changed.

What is currently done:
* 3 new tables added: `adobe_media_gallery`, `adobe_media_gallery_asset_keyword`, `adobe_media_gallery_keyword`
* Commands for media asset and keyword introduced: I added them to the API module under the `Model` folder to keep them defined at the API module and not implementation.
* Keep AdobeStockAsset keyword model and asset methods (such as getContentType, getWidth etc) to handle needed adobe stock asset data.
* Modify the save asset repository action with handling the media asset save action.

Next steps:
* Proceed work on media gallery module keywords part
* Finalize media asset logic
* Cover with tests

N/A

N/A


AdobeMediaGallery/Model/Asset/Command/GetById.php
```diff
@@ -0,0 +1,85 @@
...
+
+        if (empty($data)) {
+            $message = __('There is no such media asset with "%1"', $assetId);
+            throw new NoSuchEntityException($message);
+        }
+
+        foreach ($data as $assetData) {
+            /** @var AssetInterface $asset */
+            $asset = $this->assetFactory->create();
+            $asset->setId($assetData[AssetInterface::ID]);
```
sivaschenko: `\Magento\Framework\EntityManager\HydratorInterface::hydrate`can be used here to simplify the logic and make it more flexible.

The only thing that is needed to be able to hydrate entity is adding the short configuration to di.xm i.e.:

    <type name="Magento\Framework\EntityManager\MetadataPool">
        <arguments>
            <argument name="metadata" xsi:type="array">
                <item name="Magento\AdobeStockAssetApi\Api\Data\AssetInterface" xsi:type="array">
                    <item name="entityTableName" xsi:type="string">adobe_stock_asset</item>
                    <item name="identifierField" xsi:type="string">id</item>
                </item>
            </argument>
        </arguments>
    </type>

AdobeMediaGallery/etc/db_schema.xml
```diff
@@ -0,0 +1,59 @@
...
+        <column xsi:type="varchar" name="title" length="255" nullable="true" comment="Title"/>
+        <column xsi:type="varchar" name="content_type" length="255" nullable="true" comment="Content Type"/>
+        <column xsi:type="int" name="width" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Width"/>
+        <column xsi:type="int" name="height" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Height"/>
+        <column xsi:type="timestamp" name="created_at" on_update="false" nullable="false" default="CURRENT_TIMESTAMP" comment="Created At"/>
+        <column xsi:type="timestamp" name="updated_at" on_update="true" nullable="false" default="CURRENT_TIMESTAMP" comment="Updated At"/>
+        <index referenceId="ADOBE_MEDIA_GALLERY_ID" indexType="btree">
+            <column name="id"/>
+        </index>
+        <constraint xsi:type="unique" referenceId="ADOBE_MEDIA_GALLERY_ID_PATH_TITLE_CONTENT_TYPE_WIDTH_HEIGHT">
```
sivaschenko: Only `path` field should be unique, and `id` should be primary

AdobeMediaGallery/etc/db_schema.xml
```diff
@@ -0,0 +1,59 @@
...
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd">
+    <table name="adobe_media_gallery" resource="default" engine="innodb" comment="Adobe Media Gallery">
+        <column xsi:type="int" name="id" padding="10" unsigned="true" nullable="false" identity="true" comment="Entity ID"/>
+        <column xsi:type="varchar" name="path" length="255" nullable="true" comment="Path"/>
```
sivaschenko, line 11: please add a "source" field of nullable varchar 255 type

AdobeMediaGalleryApi/Api/Data/AssetInterface.php
```diff
@@ -0,0 +1,170 @@
...
+use Magento\Framework\Api\ExtensibleDataInterface;
+
+/**
+ * Asset Interface
+ *
+ * @api
+ */
+interface AssetInterface extends ExtensibleDataInterface
+{
+    const ID = 'id';
```
sivaschenko: Once hydrator is applied to `getById` command, these constants are going to be used only inside the Asset model, so it's better to move them to the model and make private

AdobeMediaGalleryApi/Api/Data/KeywordInterface.php
```diff
@@ -0,0 +1,66 @@
...
+
+use Magento\Framework\Api\ExtensibleDataInterface;
+
+/**
+ * Interface KeywordInterface
+ * @api
+ */
+interface KeywordInterface extends ExtensibleDataInterface
+{
+    const ID = 'id';
```
sivaschenko: Same as for asset contants

AdobeMediaGalleryApi/Model/Asset/Command/GetByIdInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+declare(strict_types=1);
+
+namespace Magento\AdobeMediaGalleryApi\Model\Asset\Command;
```
sivaschenko, line 9: Please move the service interfaces to the `Api` namespace of the module

AdobeStockAsset/Model/AssetRepository.php
```diff
@@ -121,28 +177,34 @@ public function delete(AssetInterface $asset): void
...
+                $collection,
+                AssetInterface::class
+            );
+
+            $this->collectionProcessor->process($searchCriteria, $collection);
+
+            $items = [];
+            /** @var AssetInterface $item */
+            foreach ($collection->getItems() as $item) {
+                $item->setKeywords($this->getAssetKeywords->execute($item->getId()));
```
sivaschenko: There is no need to include properties of Media Gallery Asset with Adobe Stock Asset.
For example to get keywords from asset: `$mediaGalleryGetById->execute($asset->getMediaGalleryId())->getKeywords()`

_metapackage/composer.json
```diff
@@ -13,6 +13,8 @@
         "magento/module-adobe-stock-admin-ui": "*",
         "magento/module-adobe-ims": "*",
         "magento/module-adobe-ui": "*",
-        "magento/module-adobe-ims-api": "*"
+        "magento/module-adobe-ims-api": "*",
```
sivaschenko, line 16: Can you please also update the global modman file

sivaschenko (changes requested):
Hi @coderimus , thanks for working on this implementation. It looks really good. See my review comments

AdobeMediaGalleryApi/Model/Asset/Command/GetByIdInterface.php
coderimus: I think that in this case, we can place commands under the Magento\AdobeMediaGallertApi\Model namespace. Those commands interfaces are not the WEB API and according to the [6.4. Service Contracts (Application) layer](https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html#64-service-contracts-application-layer): `6.4.1.3. All other APIs, including explicit extension points such as Chain or Composite implementations, MUST be placed under MyModuleApi/Model`. What do you think about this?

AdobeMediaGallery/etc/db_schema.xml
coderimus: But is it an applicable minimum for the case when we are trying to save the same image at the same path?

AdobeMediaGallery/Model/Asset/Command/GetById.php
coderimus: Good point! But I do not see the `@api` notation for the `Magento\Framework\EntityManager\HydratorInterface`. Are we in safety to use it for this new media gallery module?

AdobeMediaGallery/Model/Asset.php
```diff
@@ -0,0 +1,178 @@
...
+     */
+    public function setId($value): void
+    {
+        $this->setData(self::ID, $value);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function getPath(): ?string
```
YevSent: In which cases it might return `null` if the method implementation has `string` typecasting.

AdobeMediaGallery/Model/Asset/Command/GetById.php
YevSent: Also, a factory can be used to create an asset object. It seems like, the asset interface doesn't require setter methods.

AdobeMediaGallery/Model/Asset/Command/Save.php
```diff
@@ -0,0 +1,171 @@
...
+            ->from(['amg' => self::TABLE_ADOBE_MEDIA_GALLERY])
+            ->where('amg.path = ?', $asset->getPath())
+            ->where('amg.title = ?', $asset->getTitle())
+            ->where('amg.content_type = ?', $asset->getContentType())
+            ->where('amg.width = ?', $asset->getWidth())
+            ->where('amg.height = ?', $asset->getHeight());
+
+        $data = $connection->query($select)->fetchAll();
+
+        return 1;
```
YevSent: Could you explain why the method returns a hardcoded value?

AdobeMediaGallery/Model/Keyword/Command/GetAssetKeywords.php
```diff
@@ -0,0 +1,82 @@
...
+
+            $select = $connection->select()
+                ->from(['k' => self::TABLE_KEYWORD])
+                ->join(['ak' => self::TABLE_ASSET_KEYWORD], 'k.id = ak.keyword_id')
+                ->where('ak.asset_id = ?', $assetId);
+            $data = $connection->query($select)->fetchAll();
+
+            $keywords = [];
+            foreach ($data as $keywordData) {
+                $keyword = $this->keywordFactory->create();
```
YevSent: The `data` argument can be provided for the factory to create `keyword` object. Also, it seems like, the setter methods for `KeywordInterface` can be omitted.

AdobeMediaGallery/Model/ResourceModel/Asset.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace  Magento\AdobeMediaGallery\Model\ResourceModel;
+
+use Magento\AdobeMediaGalleryApi\Api\Data\AssetInterface;
+
+/**
+ * Keyword resource model
+ */
+class Asset
```
YevSent: Could you describe the purpose of this class and possible scenarios? Are there any reasons to don't use the `AssetInterface::ID` constant in the client code?

AdobeMediaGalleryApi/Model/Asset/Command/GetByIdInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+interface GetByIdInterface
+{
+    /**
+     * Get media asset by id.
+     *
+     * @param int $assetId
+     *
+     * @return \Magento\AdobeMediaGalleryApi\Api\Data\AssetInterface
+     * @throws \Zend_Db_Statement_Exception
```
YevSent: It's better to catch all Zend Exceptions in the implementation and don't expose them to the API level. Also, it adds an additional dependency to ZF1 to the API.

AdobeMediaGalleryApi/Model/Keyword/Command/SaveAssetLinksInterface.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Interface SaveAssetLinksInterface
+ * @api
+ */
+interface SaveAssetLinksInterface
+{
+    /**
+     * Save asset keywords links
+     *
+     * @param int                $assetId
```
YevSent: Formatting is broken.

AdobeStockAsset/Model/AssetRepository.php
```diff
@@ -63,21 +68,39 @@ class AssetRepository implements AssetRepositoryInterface
...
 
     /**
      * AssetRepository constructor.
      *
-     * @param ResourceModel $resource
-     * @param AssetCollectionFactory $collectionFactory
-     * @param AssetFactory $factory
-     * @param JoinProcessorInterface $joinProcessor
-     * @param CollectionProcessorInterface $collectionProcessor
+     * @param ResourceModel                      $resource
```
YevSent: Please, avoid using tab formatting as developers always will have to change it during constructor refactoring.

AdobeStockAsset/Model/AssetRepository.php
```diff
@@ -96,16 +122,46 @@ public function __construct(
...
+     *
+     * @param AssetInterface $asset
+     *
+     * @return int
+     * @throws \Magento\Framework\Exception\CouldNotSaveException
+     */
+    private function saveMediaGalleryAsset(AssetInterface $asset): int
+    {
+        /** @var MediaAssetInterface $mediaAsset */
+        $mediaAsset = $this->mediaAssetInterfaceFactory->create();
```
YevSent: You can use a factory to create `MediaAssetInterface` object instead of using setter methods.

AdobeStockAsset/Test/Unit/Model/DocumentToAssetTest.php
```diff
@@ -69,9 +69,9 @@ public function setUp(): void
         $this->categoryFactory = $this->getMockBuilder(CategoryInterfaceFactory::class)
             ->disableOriginalConstructor()
             ->getMock();
-        $this->keywordFactory = $this->getMockBuilder(KeywordInterfaceFactory::class)
-            ->disableOriginalConstructor()
-            ->getMock();
+       // $this->keywordFactory = $this->getMockBuilder(KeywordInterfaceFactory::class)
```
YevSent: Please, do not commit the commented out code.

AdobeStockAssetApi/Api/Data/AssetInterface.php
```diff
@@ -78,6 +78,11 @@ interface AssetInterface extends \Magento\Framework\Api\ExtensibleDataInterface
      */
     const MEDIA_TYPE_ID = 'media_type_id';
 
+    /**
+     * Media gallery id
+     */
+    const MEDIA_GALLERY_ID = 'media_gallery_id';
```
YevSent: All constants should have visibility scope. Also, after refactoring to use factories instead of setter methods, probably it makes sense to remove such constants at all or move to the implementation and make them private.

AdobeMediaGallery/Model/Asset/Command/Save.php
coderimus: This line along with the commented code is just a temporary thing. I made it and commit this state of implementation just to have a discussion about the command and how we operate with them. We had it together with you and Sergii Sergii Ivashchenko :) With the next update I will replace it with the proper implementation.

AdobeMediaGallery/Model/Asset.php
coderimus: Good point. Will change this part.

AdobeMediaGallery/Model/ResourceModel/Asset.php
coderimus: This class was a test try. Thre is no need to have it. It was removed in the scope of the last commit. :)

coderimus:
@sivaschenko @joni-jones 
Thank you for your previous review notes. I tried to implement them all to follow the best solutions we decided to make. Please, review the current state of this PR. But before let me clarify the some aspects:

In the scope of this PR I provided the next changes: 

1. Create a new module AdobeMediaGalley which handles media data information such as path to the media, content type, media type etc
1. In the scope of this PR I moved asset keywords logic to the AdobeMediaGalley module along with the link asset to keyword
1. Introduce the command approach for the newly added module

The next step should be:

1. Exclude AdobeMedia related object properties from the Asset object and create a new DocumentToMediaAsset provider. Remove from the asset object Adobe Media Gallery data interfaces with their implementations
1. Work on the extension attributes: add needed data to the asset object through them

AdobeMediaGallery/Model/Asset.php
```diff
@@ -0,0 +1,121 @@
...
+use Magento\AdobeMediaGalleryApi\Api\Data\AssetExtensionInterface;
+use Magento\AdobeMediaGalleryApi\Api\Data\AssetInterface;
+use Magento\Framework\Model\AbstractExtensibleModel;
+
+/**
+ * Adobe Stock Asset
+ */
+class Asset extends AbstractExtensibleModel implements AssetInterface
+{
+    private const ID = 'id';
```
sivaschenko, line 20: `private const` in a class rather than an interface :+1:

AdobeMediaGallery/Model/Asset.php
```diff
@@ -0,0 +1,121 @@
...
+
+    private const PATH = 'path';
+
+    private const TITLE = 'title';
+
+    private const CONTENT_TYPE = 'content_type';
+
+    private const WIDTH = 'width';
+
+    private const HEIGHT = 'height';
```
sivaschenko, line 26: source field is missing (can be added in a separate PR)

AdobeMediaGallery/Model/Asset/Command/GetById.php
```diff
@@ -0,0 +1,73 @@
...
+            ->where('amg.id = ?', $assetId);
+        $data = $connection->query($select)->fetchAll();
+
+        if (empty($data)) {
+            $message = __('There is no such media asset with "%1"', $assetId);
+            throw new NoSuchEntityException($message);
+        }
+
+        /** @var AssetInterface $asset */
+        $mediaAsset = $this->assetFactory->create(reset($data));
```
sivaschenko: It's better to use `fetch` (fetches one row) instead of `fetchAll` to avoid unnecessary `reset` call here

AdobeMediaGallery/Model/Asset/Command/Save.php
```diff
@@ -0,0 +1,113 @@
...
+
+            $insertData = [
+                self::PATH => $asset->getPath(),
+                self::TITLE => $asset->getTitle(),
+                self::CONTENT_TYPE => $asset->getContentType(),
+                self::WIDTH => $asset->getWidth(),
+                self::HEIGHT => $asset->getHeight(),
+            ];
+
+            $connection->insertOnDuplicate($tableName, $insertData, $onDuplicateFields);
```
sivaschenko: No need to pass `$onDuplicateFields` if they are the same as passed columns to instert
```suggestion
            $connection->insertOnDuplicate($tableName, $insertData);
```

AdobeMediaGallery/Model/Asset/Command/Save.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Get saved asset id.
+     *
+     * @param AssetInterface $asset
+     *
+     * @return int
+     */
+    private function getMediaAssetId(AssetInterface $asset): int
```
sivaschenko: All these fields are not unique and we cannot rely on identifying asset using them. It's better to use `lastInsertId` method of connection even if it's not in the interface.

Looks like it's the only way to retrieve the `id` of created entity (it's currently used in both `ResourceModel` and `EntityManager`) There is alternative `sequence` approach, but we wouldn't like to use it for assets)

AdobeMediaGallery/Model/Keyword/Command/GetAssetKeywords.php
```diff
@@ -0,0 +1,85 @@
...
+            $data = $connection->query($select)->fetchAll();
+
+            $keywords = [];
+            foreach ($data as $keywordData) {
+                $insertData = [
+                    'data' => [
+                        $keywordData,
+                    ],
+                ];
+                $assetKeyword = $this->assetKeywordFactory->create($insertData);
```
sivaschenko: `$insertData` and `$assetKeyword` variabless are unnecessary.
```suggestion
                $keywords[] = $this->assetKeywordFactory->create(['data' => $keywordData]);
```

AdobeMediaGallery/Model/Asset/Command/GetById.php
```diff
@@ -0,0 +1,73 @@
...
+use Magento\AdobeMediaGalleryApi\Model\Asset\Command\GetByIdInterface;
+use Magento\Framework\App\ResourceConnection;
+use Magento\Framework\Exception\NoSuchEntityException;
+
+/**
+ * Class GetById
+ */
+class GetById implements GetByIdInterface
+{
+    private const TABLE_ADOBE_MEDIA_GALLERY = 'adobe_media_gallery';
```
sivaschenko: References to `Adobe` should be avoided in this module (except of temporary module name) as it's going to be part of Magento core
```suggestion
    private const TABLE_MEDIA_GALLERY_ASSET = 'media_gallery_asset';
```

AdobeMediaGallery/Model/Keyword/Command/SaveAssetKeywords.php
```diff
@@ -0,0 +1,92 @@
...
+use Magento\Framework\App\ResourceConnection;
+use Magento\Framework\DB\Adapter\AdapterInterface;
+use Magento\Framework\Exception\CouldNotSaveException;
+
+/**
+ * Class SaveAssetKeywords
+ */
+class SaveAssetKeywords implements SaveAssetKeywordsInterface
+{
+    private const TABLE_KEYWORD = 'adobe_media_gallery_keyword';
```
sivaschenko:
```suggestion
    private const TABLE_KEYWORD = 'media_gallery_keyword';
```

AdobeMediaGallery/Model/Keyword/Command/SaveAssetLinks.php
```diff
@@ -0,0 +1,74 @@
...
+use Magento\Framework\App\ResourceConnection;
+use Magento\Framework\DB\Adapter\AdapterInterface;
+use Magento\Framework\Exception\CouldNotSaveException;
+
+/**
+ * Class SaveAssetLinks
+ */
+class SaveAssetLinks implements SaveAssetLinksInterface
+{
+    private const TABLE_ASSET_KEYWORD = 'adobe_media_gallery_asset_keyword';
```
sivaschenko:
```suggestion
    private const TABLE_ASSET_KEYWORD = 'media_gallery_asset_keyword';
```

AdobeMediaGallery/Model/Keyword/Command/SaveAssetLinks.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public function execute(int $assetId, array $keywordIds): void
+    {
+        try {
+            $values = [];
+            foreach ($keywordIds as $keywordId) {
+                $data[] = $assetId;
+                $data[] = $keywordId;
+                array_push($values, $data);
+                unset($data);
```
sivaschenko:
```suggestion
                $values[] = [$assetId, $keywordId];
```

AdobeMediaGallery/Model/Asset.php
```diff
@@ -0,0 +1,121 @@
...
+
+    private const CREATED_AT = 'created_at';
+
+    private const UPDATED_AT = 'updated_at';
```
sivaschenko, line 28: Please add a `source` filed

AdobeMediaGallery/etc/db_schema.xml
sivaschenko: You are right, let's keep this constraint

AdobeMediaGallery/etc/db_schema.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0"?>
+<!--
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+-->
+<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd">
+    <table name="adobe_media_gallery" resource="default" engine="innodb" comment="Adobe Media Gallery">
```
sivaschenko:
```suggestion
    <table name="media_gallery_asset" resource="default" engine="innodb" comment="Media Gallery Asset">
```

AdobeMediaGallery/etc/db_schema.xml
```diff
@@ -0,0 +1,61 @@
...
+        <column xsi:type="varchar" name="title" length="255" nullable="true" comment="Title"/>
+        <column xsi:type="varchar" name="content_type" length="255" nullable="true" comment="Content Type"/>
+        <column xsi:type="int" name="width" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Width"/>
+        <column xsi:type="int" name="height" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Height"/>
+        <column xsi:type="timestamp" name="created_at" on_update="false" nullable="false" default="CURRENT_TIMESTAMP" comment="Created At"/>
+        <column xsi:type="timestamp" name="updated_at" on_update="true" nullable="false" default="CURRENT_TIMESTAMP" comment="Updated At"/>
+        <constraint xsi:type="primary" referenceId="PRIMARY">
+            <column name="id"/>
+        </constraint>
+        <index referenceId="ADOBE_MEDIA_GALLERY_ID" indexType="btree">
```
sivaschenko:
```suggestion
        <index referenceId="MEDIA_GALLERY_ASSET_ID" indexType="btree">
```

AdobeMediaGallery/etc/db_schema.xml
```diff
@@ -0,0 +1,61 @@
...
+            <column name="id"/>
+        </index>
+        <constraint xsi:type="unique" referenceId="ADOBE_MEDIA_GALLERY_ID_PATH_TITLE_CONTENT_TYPE_WIDTH_HEIGHT">
+            <column name="path"/>
+        </constraint>
+    </table>
+    <table name="adobe_media_gallery_keyword" resource="default" engine="innodb" comment="Adobe Media Gallery Keyword">
```
sivaschenko:
```suggestion
    <table name="media_gallery_keyword" resource="default" engine="innodb" comment="Media Gallery Keyword">
```

AdobeMediaGallery/etc/db_schema.xml
```diff
@@ -0,0 +1,61 @@
...
+            <column name="id"/>
+        </index>
+        <constraint xsi:type="unique" referenceId="ADOBE_MEDIA_GALLERY_KEYWORD_KEYWORD_UNIQUE">
+            <column name="keyword"/>
+        </constraint>
+        <constraint xsi:type="primary" referenceId="PRIMARY">
+            <column name="id"/>
+        </constraint>
+    </table>
+    <table name="adobe_media_gallery_asset_keyword" resource="default" engine="innodb" comment="Adobe Media Gallery Asset Keyword">
```
sivaschenko:
```suggestion
    <table name="media_gallery_asset_keyword" resource="default" engine="innodb" comment="Media Gallery Asset Keyword">
```

AdobeMediaGallery/etc/db_schema_whitelist.json
```diff
@@ -0,0 +1,49 @@
+{
+    "adobe_media_gallery": {
```
sivaschenko:
```suggestion
    "media_gallery_asset": {
```

AdobeMediaGalleryApi/Api/Data/AssetInterface.php
```diff
@@ -0,0 +1,90 @@
...
+ * @api
+ */
+interface AssetInterface extends ExtensibleDataInterface
+{
+    /**
+     * Get ID
+     *
+     * @return int|null
+     */
+    public function getId(): ?int;
```
sivaschenko, line 25: Please add `getSource` method

AdobeMediaGalleryApi/Api/Data/AssetInterface.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public function getExtensionAttributes(): AssetExtensionInterface;
+
+    /**
+     * Set an extension attributes object.
+     *
+     * @param \Magento\AdobeStockAssetApi\Api\Data\AssetExtensionInterface $extensionAttributes
+     * @return void
+     */
+    public function setExtensionAttributes(AssetExtensionInterface $extensionAttributes): void;
```
sivaschenko: I don't think that we should keep the `setExtensionAttributes` method as the extension attributes can be set using `$data` constructor parameter using `\Magento\Framework\Api\ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY` key

AdobeMediaGalleryApi/Api/Data/KeywordInterface.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public function getExtensionAttributes(): KeywordExtensionInterface;
+
+    /**
+     * Set extension attributes
+     *
+     * @param \Magento\AdobeMediaGalleryApi\Api\Data\KeywordExtensionInterface $extensionAttributes
+     * @return void
+     */
+    public function setExtensionAttributes(KeywordExtensionInterface $extensionAttributes): void;
```
sivaschenko: I don't think that we should keep the `setExtensionAttributes` method as the extension attributes can be set using `$data` constructor parameter using `\Magento\Framework\Api\ExtensibleDataInterface::EXTENSION_ATTRIBUTES_KEY` key

AdobeMediaGalleryApi/Model/Keyword/Command/GetAssetKeywordsInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+declare(strict_types=1);
+
+namespace Magento\AdobeMediaGalleryApi\Model\Keyword\Command;
+
+/**
+ * Interface GetAssetKeywordsInterface
+ * @api
+ */
+interface GetAssetKeywordsInterface
```
sivaschenko: I'd name it `GetByIdInterface` to be consistent with Asset commands

AdobeMediaGalleryApi/Model/Keyword/Command/GetAssetKeywordsInterface.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * Get asset related keywords.
+     *
+     * @param int $assetId
+     *
+     * @return \Magento\AdobeMediaGalleryApi\Api\Data\KeywordInterface[]
+     */
+    public function execute(int $assetId): array;
```
sivaschenko: I find it better to keep function parameter names generic (makes it easier to copy-paste))
```suggestion
    public function execute(int $id): array;
```

AdobeMediaGalleryApi/README.md
```diff
@@ -0,0 +1,9 @@
+# Magento_AdobeMediaGallery Module
```
sivaschenko: Please use the new common format for the readme. Can be taken from any other module in Adobe Stock Integration, i.e.: https://github.com/magento/adobe-stock-integration/blob/develop/AdobeIms/README.md

AdobeStockAsset/Model/Asset.php
```diff
@@ -402,6 +402,22 @@ public function setContentType(string $contentType): void
...
+     */
+    public function getMediaGalleryId(): ?int
+    {
+        return $this->getData(self::MEDIA_GALLERY_ID);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function setMediaGalleryId(int $mediaGalleryId): void
```
sivaschenko, line 184: All the methods moved to media asset should be removed from Adobe Stock asset

AdobeStockAsset/Model/AssetRepository.php
```diff
@@ -96,16 +130,52 @@ public function __construct(
...
-     * @inheritdoc
+     * Save asset
+     *
+     * @param AssetInterface $asset
+     * @return void
+     * @throws \Magento\Framework\Exception\CouldNotSaveException
      */
     public function save(AssetInterface $asset): void
     {
+        $mediaGalleryAssetId = $this->saveMediaGalleryAsset($asset);
```
sivaschenko: Adobe Stock Asset Repository should be exclusively used to save Adobe Stock Asset. Please move the media gallery asset saving to `Magento\AdobeStockAsset\Model\SaveAsset` class

AdobeStockAsset/Model/AssetRepository.php
sivaschenko: Keywords are not Adobe Stock asset's property and should be set to it

AdobeStockAsset/etc/db_schema.xml
```diff
@@ -9,14 +9,10 @@
...
         <column xsi:type="int" name="category_id" padding="10" unsigned="true" nullable="true" identity="false" comment="Category ID"/>
         <column xsi:type="int" name="creator_id" padding="10" unsigned="true" nullable="true" identity="false" comment="Creator ID"/>
-        <column xsi:type="varchar" length="255" name="path" nullable="true" comment="Path"/>
         <column xsi:type="int" name="is_licensed" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Is Licensed"/>
-        <column xsi:type="varchar" length="255" name="title" nullable="true" comment="Title"/>
-        <column xsi:type="int" name="width" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Width"/>
-        <column xsi:type="int" name="height" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Height"/>
-        <column xsi:type="varchar" length="255" name="content_type" nullable="true" comment="Content Type"/>
         <column xsi:type="varchar" length="255" name="creation_date" nullable="true" comment="Creation Date"/>
```
sivaschenko: Please remove `creation_date field` as well

AdobeStockAsset/etc/db_schema.xml
```diff
@@ -9,14 +9,10 @@
     <table name="adobe_stock_asset" resource="default" engine="innodb" comment="Adobe Stock Asset">
         <column xsi:type="int" name="id" padding="10" unsigned="true" nullable="false" identity="true" comment="Entity ID"/>
         <column xsi:type="int" name="media_type_id" padding="10" unsigned="true" nullable="true" identity="false" comment="Media Type ID"/>
```
sivaschenko: Please remove `media_type_id` filed. Looks like it's duplicate to `content_type` (`content_type` is more detailed)

AdobeStockAsset/etc/db_schema.xml
```diff
@@ -37,6 +33,7 @@
         </constraint>
         <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_CATEGORY_ID_ADOBE_STOCK_CATEGORY_ID" table="adobe_stock_asset" column="category_id" referenceTable="adobe_stock_category" referenceColumn="id" onDelete="SET NULL"/>
         <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_CREATOR_ID_ADOBE_STOCK_CREATOR_ID" table="adobe_stock_asset" column="creator_id" referenceTable="adobe_stock_creator" referenceColumn="id" onDelete="SET NULL"/>
+        <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_MEDIA_GALLERY_ID_ADOBE_STOCK_CREATOR_ID" table="adobe_stock_asset" column="media_gallery_id" referenceTable="adobe_media_gallery" referenceColumn="id" onDelete="SET NULL"/>
```
sivaschenko:
```suggestion
        <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_MEDIA_GALLERY_ID_MEDIA_GALLERY_ASSET_ID" table="adobe_stock_asset" column="media_gallery_id" referenceTable="media_gallery_asset" referenceColumn="id" onDelete="SET NULL"/>
```

AdobeStockAssetApi/Api/AssetRepositoryInterface.php
```diff
@@ -40,7 +40,9 @@ public function delete(AssetInterface $item): void;
      * Get a list of assets
      *
      * @param \Magento\Framework\Api\SearchCriteriaInterface $searchCriteria
+     *
      * @return \Magento\AdobeStockAssetApi\Api\Data\AssetSearchResultsInterface
+     * @throws \Magento\Framework\Exception\IntegrationException
      */
     public function getList(SearchCriteriaInterface $searchCriteria) : AssetSearchResultsInterface;
```
sivaschenko, line 37: Retrieving assets should be performed from the database and should not result in calls to Adobe Stock API, so there should not be a chance of `\Magento\Framework\Exception\IntegrationException` being thrown

AdobeStockAssetApi/Api/Data/AssetInterface.php
```diff
@@ -15,52 +15,124 @@
  */
 interface AssetInterface extends \Magento\Framework\Api\ExtensibleDataInterface
 {
+    /**
+     * Asset Id
+     */
     const ID = 'id';
```
sivaschenko: Let's move these constants to implementation class and make them private

AdobeStockAssetApi/Api/Data/AssetInterface.php
```diff
@@ -15,52 +15,124 @@
...
 
+    /**
+     * Asset save path
+     */
     const PATH = 'path';
```
sivaschenko: Fields moved to media gallery asset should be removed

sivaschenko (changes requested):
Hi @coderimus thanks for extracting a media gallery module from `AdobeStockAsset` ! The `AdobeMediaGallery` module is very well organized!

Please see my review comments

sivaschenko (approved):

Issue #526, opened 2019-10-10
Position of user profile block is dependent on browser window size


sivaschenko:
### Preconditions
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Sign In to Adobe using the link in the top right corner
5. Resize browser window

### Expected result
User profile block shifts left in smaller browser windows and it shifts right as the window become wider (see screenshots below)

Small window:
![image](https://user-images.githubusercontent.com/2028541/66593539-1e236e80-eb8f-11e9-9ba7-9893c3504259.png)

Wide window:

![image](https://user-images.githubusercontent.com/2028541/66593614-3bf0d380-eb8f-11e9-848c-88255986a492.png)

PR #533, opened 2019-10-11
Position of user profile block is dependent on browser window size

Nazar65:

2. magento/adobe-stock-integration#526 Position of user profile block is dependent on browser window size

Same as in issue.

1  https://github.com/magento/adobe-stock-integration/issues/526


sivaschenko (changes requested):
Hi @Nazar65 thanks for the pull request! I have tested the functionality and while the user profile position is perfect, the issue with extra margin still seems to be present

sivaschenko (approved):

Issue #489, opened 2019-09-26
[Unit tests] Cover \Magento\AdobeStockImage\Model\GetImageList with a unit test


sivaschenko:
Cover \Magento\AdobeStockImage\Model\GetImageList with a unit test

sivaschenko:
Hi @karyna-tsymbal-atwix  , do you work on this issue? are there any updates?

PR #538, opened 2019-10-13
#489 Cover \Magento\AdobeStockImage\Model\GetImageList with a unit test

karyna-t:
### Description (*)

Unit test for \Magento\AdobeStockImage\Model\GetImageList.

Issue: https://github.com/magento/adobe-stock-integration/issues/489


sivaschenko (changes requested):
Thanks for your pull request @karyna-tsymbal-atwix , however, the test doe not cover major functionality of GetImageList class. 

Can you please add a test(s) that is also covering situation when default filters and applied filters are not empty. I'd suggest using data provider to be able to test several default/applied filters combinations.

Thank you!

sivaschenko (approved):

PR #574, opened 2019-10-24
FIX: keywords save during the image save

coderimus:
### Description (*)
The main part of this hotfix PR is to fix the keywords save. Previously if image didn't have keywords the get keywords action return an empty array and, as a result, no keywords saved. Now we check: if there are no keywords it means that we need to take them from the media gallery document and save.

And as additional, I just rearranged some included class :)

N/A

N/A


coderimus:
Hello @sivaschenko ,
Please, review the provided changes. I changed some logic for the save image process at the execute method. I'd rather simplify it with removing some logical part and also use direct comparison and `isset` instead of `empty`.

sivaschenko (approved):

Issue #514, opened 2019-10-07
[Bug] Serie Id and Model Id filters do not indicate applied values after relogin to admin panel


sivaschenko:
### Preconditions
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> Advanced -> System -> Adobe Stock Integration fieldset

### Steps to reproduce
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Click on any image to expand image preview
5. Click on "See more" link in the related images section
6. Logout from admin panel
7. Login back to admin panel from another/incognito browser
8. Reopen the Adobe Stock grid

### Actual result
Step 5: Related images filter (Serie Id/Model Id) filter is applied to the grid
Step 8: The filter is applied but the applied value is not indicated in the applied filters section (see screen shot below)

### Expected result
The filter applied value should be indicated in the applied filters section

![image](https://user-images.githubusercontent.com/2028541/66319622-5a4b9a80-e915-11e9-92c8-cb009f637ff1.png)

konarshankar07:
Hello @sivaschenko ,
I am not able to reproduce this issue as per the steps you have added in this ticket. Can you please confirm?
Thanks

PR #580, opened 2019-10-26
See more link is visible even if there are less than 4 related images

konarshankar07:
### Description (*)

This issue was found whilst investigating https://github.com/magento/adobe-stock-integration/issues/514. See more link is visible for the merchant/user if there are less than 4 related images for the record.

N/A

1. Login to admin panel
4. Search the keyword with text`'298323255`
5. Press enter
6. Click on first image to expand image preview
7. Click on more from the series tab in related image section
8. Check for the see more link

### Actual result
There is a see more link for the record

### Expected result
There should not be see more link for the record because there is only three related series image


AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -70,6 +70,16 @@ define([
             return record.series;
         },
 
+        /**
+         * Count number of images in Related Serie
+         *
+         * @param {Object} record
+         * @returns boolean
+         */
+        countSeriesImages: function(record) {
```
sivaschenko: Please change the name and comment for the function to correctly explain the function purpose. It's not returning the number of images but if the number of images in greater than 4

```suggestion
        canShowMore: function(record) {
```

Same for the `model`

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -70,6 +70,16 @@ define([
...
+            return (parseInt(record.series().length, 10) >= 4);
```
sivaschenko: Radix parameter is 10 by default, so it's possible to skip passing this optional parameter to `parseInt`. Also, braces are not necessary.

```suggestion
            return parseInt(record.series().length) >= 4;
```

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -80,6 +90,16 @@ define([
...
         },
 
+        /**
+         * Count number of images in Related Model
+         *
+         * @param {Object} record
+         * @returns boolean
+         */
+        countModelImages: function(record) {
+            return (parseInt(record.model().length, 10) >= 4);
```
sivaschenko: let's move "4" to `tabImagesLimit` ui component option.

sivaschenko (changes requested):
Thanks for the pull request @konarshankar07 ! Please see my review comments

sivaschenko (approved):

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
```diff
@@ -150,7 +152,7 @@ define([
                 showLoader: true,
                 data: {
                     'image_id': record.id,
-                    'limit': 4
+                    'limit': self.tabImagesLimit
```
sivaschenko: `this` should work here,  it is used a few lines above for setting the URL. Otherwise, it's always better to specify `context: this` or `.bind(this)` instead of declaring a variable.

sivaschenko (changes requested):
Hi @konarshankar07 thanks for updates! See another one small comment

sivaschenko (approved):

Issue #587, opened 2019-10-26
Cover GetAssetKeywords class with a unit test


sivaschenko:
Cover `\Magento\AdobeMediaGallery\Model\Keyword\Command\GetAssetKeywords` with a unit test

PR #592, opened 2019-10-27
Add unit test for 'get asset keywords' command

valentinboyanov:
### Description (*)
Add unit test for 'get asset keywords' command implemented here `\Magento\AdobeMediaGallery\Model\Keyword\Command\GetAssetKeywords`

### Fixed Issues
1. magento/adobe-stock-integration#587: Cover GetAssetKeywords class with a unit tes

### DIFF
--- a/AdobeMediaGallery/Test/Unit/Model/Keyword/Command/GetAssetKeywordsTest.php
+++ b/AdobeMediaGallery/Test/Unit/Model/Keyword/Command/GetAssetKeywordsTest.php
@@ -51,40 +51,28 @@ class GetAssetKeywordsTest extends TestCase
         $this->assertInstanceOf( GetAssetKeywords::class, $this->sut);
     }
 
-    public function testFindOne()
+    /**
+     * @dataProvider cases()
+     */
+    public function testFind($databaseQueryResult, $expectedNumberOfFoundKeywords)
     {
-        $assetIdWithOneKeyword = 1;
-        $oneKeywordQueryResult = ['keywordRawData'];
-        $this->configureResourceConnectionStub($oneKeywordQueryResult);
+        $randomAssetId = 12345;
+        $this->configureResourceConnectionStub($databaseQueryResult);
         $this->configureAssetKeywordFactoryStub();
 
-        $keywords = $this->sut->execute($assetIdWithOneKeyword);
+        /** @var KeywordInterface[] $keywords */
+        $keywords = $this->sut->execute($randomAssetId);
 
-        $this->assertCount(1, $keywords);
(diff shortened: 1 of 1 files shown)


AdobeMediaGallery/Test/Unit/Model/Keyword/Command/GetAssetKeywordsTest.php
```diff
@@ -0,0 +1,140 @@
...
+            $this->assetKeywordFactoryStub
+        );
+    }
+
+    public function testCanBeCreated()
+    {
+        $this->assertInstanceOf( GetAssetKeywords::class, $this->sut);
+    }
+
+    public function testFindOne()
```
sivaschenko: It would be good to use a data provider providing query result instead of separate test methods for one, two and zero results.

AdobeMediaGallery/Test/Unit/Model/Keyword/Command/GetAssetKeywordsTest.php
```diff
@@ -0,0 +1,140 @@
...
+    {
+        $assetIdWithOneKeyword = 1;
+        $oneKeywordQueryResult = ['keywordRawData'];
+        $this->configureResourceConnectionStub($oneKeywordQueryResult);
+        $this->configureAssetKeywordFactoryStub();
+
+        $keywords = $this->sut->execute($assetIdWithOneKeyword);
+
+        $this->assertCount(1, $keywords);
```
sivaschenko: I think it's better to perform more strict assertion here:
```suggestion
        $this->asserEquals($oneKeywordQueryResult, $keywords);
```

sivaschenko (changes requested):
Hi @valentinboyanov thanks for the pull request and welcome back to the project! Please see my review comments.

If this PR is made at the Meet Magento Spain contribution day, can you please add a label "mm19es" to the PR.

Thanks!

Please ignore Static tests build - it's not currently stabilized

AdobeMediaGallery/Test/Unit/Model/Keyword/Command/GetAssetKeywordsTest.php
valentinboyanov: I thought about this, but I have finally decided to test only the responsibility of the command. In this case, check the content of the array would be test something that is moked `\Magento\AdobeMediaGallery\Test\Unit\Model\Keyword\Command\GetAssetKeywordsTest::configureAssetKeywordFactoryStub` and this makes the test fragile.

sivaschenko (dismissed):

sivaschenko (approved):

Issue #582, opened 2019-10-26
Remove database records when image is deleted from media gallery


sivaschenko:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Click on any image to expand preview section
5. Save Preview version of the image
6. Delete the saved image in the media gallery

### Expected result (*)
`media_gallery_asset` and `adobe_stock_asset` tabels do not contain deleted image information

### Technical Details:

There is a plugin for adobe stock asset `\Magento\AdobeStockImage\Plugin\Product\Gallery\Processor` the plugin should be revisited and corresponding plugin should be created for the media gallery asset (`AdobeMediaGallery` module)

jporter10:
@magento I am working on it

sivaschenko:
Hi @jporter10 are you working on this issue?

PR #597, opened 2019-10-29
Remove database records when image is deleted from media gallery

coderimus:
### Description
#582 Remove database records when an image is deleted from media gallery.

This PR provides delete functionality for the media asset and related data. Also, the asset object got the CASCADE on delete instruction when the media object was triggered to be deleted.

Provides some refactoring and code improvements like logging, logical improvements in the code execution etc

### Fixed Issues
N/A

### Manual testing scenarios
1. Login to admin panel
1. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
1. Click "Search Adobe Stock" button to open images grid
1. Click on any image to expand preview section
1. Save Preview version of the image
1. Delete the saved image in the media gallery

### Acceptance criteria
`media_gallery_asset` and a`dobe_stock_asset` tables do not contain deleted image information


coderimus:
Hello @sivaschenko ,
Please create an issue for coverage with unit tests the DeleteById class and assign on me. Thank you :)

coderimus:
@sivaschenko please, ignore my previous message about the new issue for cover the delete by id functionality. I added this test in scope of this PR.

2 moments to highlight:

1. The DeleteById unit test covers 2 scenarios: when the delete function executed successfully and when some error occurred.
 1. You will see this line `$connection->delete($tableName, 'id = ' . $mediaAsset->getId());` I choose concatenation instead of sending `where` query parameters as an array because after the deep investigation I found that this approach with an array makes us completely depended on the Zend implementation - it accept mixed types for where condition and convert array to string at the low level. I would prefer to send a string to the delete method as a where parameter.

AdobeMediaGallery/Model/Asset/Command/DeleteById.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public function execute(int $mediaAssetId): void
+    {
+        try {
+            /** @var AssetInterface $mediaAsset */
+            $mediaAsset = $this->getMediaAssetById->execute($mediaAssetId);
+            /** @var AdapterInterface $connection */
+            $connection = $this->resourceConnection->getConnection();
+            $tableName = $this->resourceConnection->getTableName(self::TABLE_MEDIA_GALLERY_ASSET);
+            $connection->delete($tableName, 'id = ' . $mediaAsset->getId());
```
sivaschenko: I saw you comment, but I think it's still necessary to pass the argument as an array as it is a good practice to quote all variables used for Mysql queries. To be sure to be sure :)

AdobeMediaGallery/Model/Asset/Command/DeleteById.php
```diff
@@ -0,0 +1,83 @@
...
+        } catch (\Exception $exception) {
+            $message = __(
```
sivaschenko: Can you please check why static tests are failing for this construction:

    inconsistency the number of arguments and placeholders were discovered: 

    Phrase: 'Could not delete media asset with id %id: %error' 
    File: /var/www/html/app/code/Magento/AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php 
    Line: 151

AdobeMediaGallery/Model/Asset/Command/DeleteById.php
sivaschenko: Also, ideally `id` should be a constant as well

AdobeMediaGallery/Model/Asset/Command/GetListByPath.php
```diff
@@ -0,0 +1,85 @@
...
+     * @return array
+     * @throws IntegrationException
+     */
+    public function execute(string $mediaFilePath): array
+    {
+        try {
+            $connection = $this->resourceConnection->getConnection();
+            $select = $connection->select()
+                ->from(self::TABLE_MEDIA_GALLERY_ASSET)
+                ->where('path in (?)', $mediaFilePath);
```
sivaschenko: I'd move `path` to a constant

AdobeMediaGallery/Model/Asset/Command/GetListByPath.php
```diff
@@ -0,0 +1,85 @@
...
+            $data = $connection->fetchAssoc($select);
+
+            $mediaAssets = [];
+            foreach ($data as $id => $assetData) {
+                $mediaAssets[$id] = $this->mediaAssetFactory->create(['data' => $assetData]);
```
sivaschenko: Cool approach ;)

AdobeMediaGallery/Model/Asset/Command/GetListByPath.php
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Return media asset asset list
+     *
+     * @param string $mediaFilePath
+     *
+     * @return array
+     * @throws IntegrationException
+     */
+    public function execute(string $mediaFilePath): array
```
sivaschenko: Considering that path is unique, I'd rename the command to `GetByPath` and return `AssetInterface`

AdobeMediaGallery/Plugin/Product/Gallery/Processor.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param ProcessorSubject $subject
+     * @param $result
+     * @param Product $product
+     * @param $file
+     *
+     * @return mixed
+     */
+    public function afterRemoveImage(ProcessorSubject $subject, $result, Product $product, $file)
+    {
+        if (!is_string($file)) {
```
sivaschenko, line 57: I like this attention to details 👍

AdobeMediaGallery/Plugin/Product/Gallery/Processor.php
```diff
@@ -0,0 +1,85 @@
...
+
+        try {
+            $mediaAssetList = $this->getMediaListByPath->execute($file);
+        } catch (NoSuchEntityException $exception) {
+            return $result;
+        }
+
+        /** @var AssetInterface $mediaAsse */
+        foreach ($mediaAssetList as $mediaAsset) {
+            $this->deleteMediaAssetById->execute($mediaAsset->getId());
```
sivaschenko: I was thinking that `DeleteByPath` command can replace both `DeleteById` and `GetListByPath`

AdobeMediaGallery/Plugin/Wysiwyg/Images/Storage.php
```diff
@@ -0,0 +1,104 @@
...
+     */
+    public function afterDeleteFile(StorageSubject $subject, $result, $target)
+    {
+        if (!is_string($target)) {
+            return $result;
+        }
+
+        $relativePath = $this->filesystem->getDirectoryRead(DirectoryList::MEDIA)->getRelativePath($target);
+        $mediaAssetList = $this->getMediaListByPath->execute($relativePath);
+        if (!isset($mediaAssetList)) {
```
sivaschenko: It's impossible for these conditions to result in `true` considering the current implementation

AdobeMediaGallery/Plugin/Wysiwyg/Images/Storage.php
```diff
@@ -0,0 +1,104 @@
...
+            return $result;
+        }
+
+        try {
+            /** @var AssetInterface $mediaAsse */
+            foreach ($mediaAssetList as $mediaAsset) {
+                $this->deleteMediaAssetById->execute($mediaAsset->getId());
+            }
+        } catch (\Exception $exception) {
```
sivaschenko, line 87: It will make sense to try-catch wrap the delete operation in the Processor plugin as well

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+            [
+                'resourceConnection' => $this->resourceConnection,
+                'getMediaAssetById' => $this->getMediaAssetById,
+                'mediaAssetFactory' =>  $this->mediaAssetFactory,
+                'logger' =>  $this->logger,
+            ]
+        );
+
+        $this->mediaAsset = $this->createMock(AssetInterface::class);
+        $this->adapter = $this->createMock(AdapterInterface::class);
```
sivaschenko: If the `adaptor` is created in `setUp` - it's a good idea to configure the `resourceConnection` mock to return this adaptor in `setUp` method as well.

In this case we don't need a property for `resourceConnection` and will work only with `adaptor` property further

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Test delete media asset by id action
+     *
+     * @param int $mediaAssetId
+     * @dataProvider deleteMediaAssetTestScopes
+     */
+    public function testSuccessfulDeleteByIdExecution(int $mediaAssetId)
+    {
+        $this->getMediaAssetById->expects($this->at(0))
```
sivaschenko:
```suggestion
        $this->getMediaAssetById->expects($this->once())
```

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+            ->method('execute')
+            ->with($mediaAssetId)
+            ->willReturn($this->mediaAsset);
+
+        $this->resourceConnection->expects($this->at(0))
```
sivaschenko:
```suggestion
        $this->resourceConnection->expects($this->once())
```

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+            ->method('getConnection')
+            ->willReturn($this->adapter);
+
+        $tableName = 'media_gallery_asset';
+        $this->resourceConnection->expects($this->at(1))
```
sivaschenko: I don't think it's necessary to test that `getTableName` is called strictly after `getConnection`. While that sequence of calls is not required, asserting the sequence makes the test more fragile
```suggestion
        $this->resourceConnection->expects($this->once())
```

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+            ->method('getTableName')
+            ->with($tableName)
+            ->willReturn($tableName);
```
sivaschenko: I'd use different table names to ensure that the correct one is used further

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+        $this->mediaAsset->expects($this->once())
+            ->method('getId')
+            ->willReturn($mediaAssetId);
+
+        $this->adapter->expects($this->at(0))
```
sivaschenko: There is only one call to `adapter::delete` method expected, no need to verify the sequence
```suggestion
        $this->adapter->expects($this->once())
```

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+            DeleteById::class,
+            [
+                'resourceConnection' => $this->resourceConnection,
+                'getMediaAssetById' => $this->getMediaAssetById,
+                'mediaAssetFactory' =>  $this->mediaAssetFactory,
+                'logger' =>  $this->logger,
+            ]
+        );
+
+        $this->mediaAsset = $this->createMock(AssetInterface::class);
```
sivaschenko: I'd create the asset mock directly in the test as this is a dynamic entity (theoretically more than one asset can be used)

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Assume that delete action will thrown an Exception
+     *
+     * @param int $mediaAssetId
+     * @dataProvider deleteMediaAssetTestScopes
+     */
+    public function testExceptionOnDeleteExecution(int $mediaAssetId)
+    {
+        $this->getMediaAssetById->expects($this->at(0))
```
sivaschenko: Please use `at` method only when you need to verify the sequence of calls (or when the same method is called several times with either different arguments or different return values)

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+            [
+                'id' => $mediaAssetId,
+                'error' => $testErrorText,
+            ]
+        )->render();
+        $this->logger->expects($this->once())
+            ->method('critical')
+            ->with($testMessage)
+            ->willReturnSelf();
+        $this->expectExceptionMessage($testMessage);
```
sivaschenko: I wouldn't verify the test message. It will make the test fail even if small correction to exception message phrasing is applied

AdobeMediaGallery/Test/Unit/Model/Asset/Command/DeleteByIdTest.php
```diff
@@ -0,0 +1,177 @@
...
+        $this->deleteMediaAssetById->execute($mediaAssetId);
+    }
+
+    /**
+     * @return array
+     */
+    public function deleteMediaAssetTestScopes(): array
```
sivaschenko: Data provider is not necessary to providing id's in my opinion, the behaviour is pretty much the same for any `id`.

Also, (note for the future unit tests) data providers are usually used to provide at least two arguments for the test:
test data and expected result (or flags telling the test how the verification should change depending on test data provided) i.e. `$input, $output` or `$input, $isExceptionExpected`

AdobeStockAsset/etc/db_schema.xml
```diff
@@ -27,7 +27,7 @@
         </constraint>
         <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_CATEGORY_ID_ADOBE_STOCK_CATEGORY_ID" table="adobe_stock_asset" column="category_id" referenceTable="adobe_stock_category" referenceColumn="id" onDelete="SET NULL"/>
         <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_CREATOR_ID_ADOBE_STOCK_CREATOR_ID" table="adobe_stock_asset" column="creator_id" referenceTable="adobe_stock_creator" referenceColumn="id" onDelete="SET NULL"/>
-        <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_MEDIA_GALLERY_ID_ADOBE_STOCK_CREATOR_ID" table="adobe_stock_asset" column="media_gallery_id" referenceTable="media_gallery_asset" referenceColumn="id" onDelete="SET NULL"/>
+        <constraint xsi:type="foreign" referenceId="ADOBE_STOCK_ASSET_MEDIA_GALLERY_ID_ADOBE_STOCK_CREATOR_ID" table="adobe_stock_asset" column="media_gallery_id" referenceTable="media_gallery_asset" referenceColumn="id" onDelete="CASCADE"/>
```
sivaschenko: I think `CREATOR_ID` was added to the constraint name by mistake before, it would be good to correct the name

AdobeMediaGalleryApi/Model/Asset/Command/SaveInterface.php
```diff
@@ -11,19 +11,18 @@
 use Magento\AdobeMediaGalleryApi\Api\Data\AssetInterface;
 
 /**
- * Save media asset.
- *
+ * Save media asset
  * @api
```
sivaschenko: Could you please remove all the `@api` annotation from `AdobeMediaGalleryApi` module. We cannot add API to the patch release, but we can mark these interfaces back with `@api` later in 2.4 release.

I missed this point initially, but thankfully that was brought up by @joni-jones during the public meeting yesterday

sivaschenko (changes requested):
@coderimus Thanks for the pull request! Looks great!

As per your request, I did a very detailed and critical review of the unit tests :)

Please see my review comments

AdobeMediaGallery/Plugin/Wysiwyg/Images/Storage.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * @var DeleteByIdInterface
+     */
+    private $deleteMediaAssetById;
+
+    /**
+     * @var Filesystem
+     */
+    protected $filesystem;
```
sivaschenko: Magento Health Index test results helped to identify this unnecessary protected scope
```suggestion
    private $filesystem;
```

coderimus:
@sivaschenko thank you for your detail and constructive review. I implemented the requested changes, but there are two questions:

1. You wrote about [those two lines of the test class](https://github.com/magento/adobe-stock-integration/pull/597/files#diff-b0139084384152b31004be7e37c8dae7R83-R84): "I'd use different table names to ensure that the correct one is used further" Sorry, but I didn't catch the idea of this suggestion. If I used different table names, I would get an error. Please, describe a little bit detail what you mean :)

1. I fixed some static tests for error message parameters inconsistent, but there are some more: AdobeMediaGallery has soft dependencies of Magento_CMS and Magento_Catalog. I added the sequence to the module, but there are still errors in the static tests. Can you please check the static test report? Maybe we can ignore those messages or turnoff them for media module?

AdobeMediaGallery/Model/Asset/Command/GetByPath.php
```diff
@@ -0,0 +1,90 @@
...
+     * @return AssetInterface
+     * @throws IntegrationException
+     */
+    public function execute(string $mediaFilePath): AssetInterface
+    {
+        try {
+            $connection = $this->resourceConnection->getConnection();
+            $select = $connection->select()
+                ->from(self::TABLE_MEDIA_GALLERY_ASSET)
+                ->where(self::MEDIA_GALLERY_ASSET_PATH . ' in (?)', $mediaFilePath);
```
sivaschenko:
```suggestion
                ->where(self::MEDIA_GALLERY_ASSET_PATH . ' = ?', $mediaFilePath);
```

AdobeMediaGallery/Model/Keyword/Command/SaveAssetKeywords.php
```diff
@@ -47,7 +47,11 @@ public function __construct(
     }
 
     /**
-     * @inheritdoc
+     * Save asset keywords.
+     *
+     * @param KeywordInterface[] $keywords
+     * @param int $assetId
+     * @throws CouldNotSaveException
```
sivaschenko: I'd use `@inheritdoc` here

AdobeMediaGallery/etc/module.xml
```diff
@@ -6,5 +6,10 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Magento_AdobeMediaGallery" />
+    <module name="Magento_AdobeMediaGallery">
+        <sequence>
+            <module name="Magento_Catalog"/>
```
sivaschenko: moule.xml is used to define installation dependencies. `AdobeMediaGallery` depends on Catalog and CMS modules, but they do not necessarily need to be installed before `AdobeMediaGallery` (as there are not database schema dependency). So these dependencies should only be added to composer.json

sivaschenko (changes requested):

AdobeMediaGallery/Model/Keyword/Command/SaveAssetKeywords.php
coderimus: I removed it because of [this rule in the dockbloc standards](https://devdocs.magento.com/guides/v2.3/coding-standards/docblock-standard-general.html#inheritdoc) "The @inheritdoc tag SHOULD NOT be used. If a child class method requires a long description to explain its purpose, it may use @inheritdoc to indicate the new description is intended as an addition to the parent method description. In general such method overrides are a code smell and should be used as an incentive to make the code more self-documenting if possible."

AdobeMediaGallery/etc/module.xml
coderimus: @sivaschenko thank you for an explanation! I got the point now with this soft/hard dependencies in our case :)

sivaschenko (approved):

Issue #588, opened 2019-10-26
Cover SaveAssetKeywords class with a unit test


sivaschenko:
Cover `\Magento\AdobeMediaGallery\Model\Keyword\Command\SaveAssetKeywords` with a unit test

PR #606, opened 2019-10-30
[Unit test coverage] Covering the SaveAssetKeywords by Unit Test

eduard13:
### Description (*)
This PR introduces unit test coverage for `\Magento\AdobeMediaGallery\Model\Keyword\Command\SaveAssetKeywords` model class.

1. magento/adobe-stock-integration#588: Cover SaveAssetKeywords class with a unit test

N/A


AdobeMediaGallery/Test/Unit/Model/Keyword/Command/SaveAssetKeywordsTest.php
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Providing asset keywords
+     *
+     * @return array
+     */
+    public function assetKeywordsDataProvider(): array
+    {
+        return [
+            [
+                [],
```
sivaschenko, line 142: Hm, looks like the test actually discovered an issue here!)

Would you be able to adjust the code of `SaveAssetKeywords ` so that `insertArray` is not called if there is nothing to insert?

AdobeMediaGallery/Test/Unit/Model/Keyword/Command/SaveAssetKeywordsTest.php
```diff
@@ -0,0 +1,144 @@
...
+        $this->sut->execute([], 1);
+    }
+
+    /**
+     * Preparing the resource connection
+     */
+    private function prepareResourceConnection(): void
+    {
+        $selectStub = $this->createMock(Select::class);
+        $selectStub->method('from')->willReturnSelf();
```
sivaschenko: All those methods are not called in the class under test. It is necessary to keep only `getConnection` method call expectation and ideally move it to the `setUp` method

sivaschenko (changes requested):
Thanks for the test @eduard13 , please take a look at the review comments

AdobeMediaGallery/Test/Unit/Model/Keyword/Command/SaveAssetKeywordsTest.php
eduard13: Actually we have the `getKeywordIds` private method, that fetches the keyword ids. The reason of why I moved configuring `resourceConnection` separately, is that if an Exception is thrown, then we don't need `connection` at all.

eduard13:
Hi @sivaschenko, thanks for reviewing it. The model and test for it were adjusted. Could you please check the changes?

sivaschenko (approved):

Issue #208, opened 2019-07-11
[MFTF]: User cannot access stock images from Magento without token


sidolov:
### Description
 Implement MFTF test to cover scenario: "User cannot access stock images from Magento without token"
### Test steps
1. Before - Save Adobe Stock API credentials in Magento Admin
2. Before - LoginAsAdmin
3. Action Disable Adobe Stock Integration
4. Navigate to */cms/page/
5. User clicks Add New Page button
6. Save Configuration
7. Go to Magento Media Gallery (i.e. CMS -> Page -> Add new page -> Insert Image)
8. Click `Search Adobe Stock` button
9. Assert that correct error message is shown "Failed to authenticate to Adobe Stock API" in the grid body
10. After - Remove Adobe Stock API credentials in Magento Admin
11. After - logout

PR #607, opened 2019-10-30
[MFTF coverage] Implementing the MFTF for "User cannot access stock images ...

eduard13:
...from Magento without token" scenario

### Description (*)
This PR covers the following scenario by MFTF:

- Implement MFTF test to cover scenario: "User cannot access stock images from Magento without token"

1. magento/adobe-stock-integration#208: [MFTF]: User cannot access stock images from Magento without token

N/A

### Possible issue

While running any of MFTF tests, I get the following JS error:
```text
browser.js:95 Uncaught TypeError: Cannot read property 'targetElementId' of undefined
```
So checking that line, we have the following:
```javascript
}).done(function (data) {
    self.modal.html(data).trigger('contentUpdated');
    self.modalLoaded = true;

    // Uncommenting the following condition, fixes the issue
    // if (typeof options !== 'undefined') {
        self.targetElementId = options.targetElementId; // The error is thrown on this line
    // }
});
```
ℹ️ I'd like to confirm that there is no error on opening the `Insert file` modal manually.


sivaschenko:
Hi @eduard13 , the JS issue does not seem to be related to Adobe Stock Integration. Is it reproducible on vanilla Magento?

AdobeStockAdminUi/Test/Mftf/ActionGroup/AdminAssertFailedAuthenticationMessageActionGroup.xml
```diff
@@ -0,0 +1,17 @@
...
+<!--
+ /**
+  * Copyright © Magento, Inc. All rights reserved.
+  * See COPYING.txt for license details.
+  */
+-->
+
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
+    <actionGroup name="AdminAssertFailedAuthenticationMessageActionGroup">
```
rogyar: Please, rename the action group to `AssertAdminFailedAuthenticationMessageActionGroup` (see 11.3.5 in https://devdocs.magento.com/guides/v2.3/coding-standards/technical-guidelines.html)

rogyar (changes requested):

AdobeStockAdminUi/Test/Mftf/ActionGroup/AdminAssertFailedAuthenticationMessageActionGroup.xml
eduard13: Hi @rogyar, the ActionGroup has been renamed.
Thank you.

eduard13:
> Hi @eduard13 , the JS issue does not seem to be related to Adobe Stock Integration. Is it reproducible on vanilla Magento?

@sivaschenko I couldn't reproduce it there, so let's skip it then.

AdobeStockAdminUi/Test/Mftf/ActionGroup/AssertAdminFailedAuthenticationMessageActionGroup.xml
```diff
@@ -0,0 +1,17 @@
...
+  * Copyright © Magento, Inc. All rights reserved.
+  * See COPYING.txt for license details.
+  */
+-->
+
+<actionGroups xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+              xsi:noNamespaceSchemaLocation="urn:magento:mftf:Test/etc/actionGroupSchema.xsd">
+    <actionGroup name="AssertAdminFailedAuthenticationMessageActionGroup">
+        <arguments>
+            <argument name="message" type="string" defaultValue="Failed to authenticate to Adobe Stock API. Please correct the API credentials"/>
```
hws47a: Can we use a data entity to store this text? As I know data should be separated from the logic.
Maybe this file can be used: https://github.com/magento/adobe-stock-integration/blob/develop/AdobeStockImageAdminUi/Test/Mftf/Data/AdobeStockModalData.xml

AdobeStockAdminUi/Test/Mftf/ActionGroup/AssertAdminFailedAuthenticationMessageActionGroup.xml
```diff
@@ -0,0 +1,17 @@
...
+        </arguments>
+        <see selector="{{AdobeStockSection.modal}}" userInput="{{message}}" stepKey="seeFailedAuthenticationMessage"/>
+    </actionGroup>
+</actionGroups>
```
hws47a: An empty line should be at the end of any file.

AdobeStockAdminUi/Test/Mftf/Test/AdminCannotAccessStockImagesWithWrongCredentialsTest.xml
```diff
@@ -0,0 +1,35 @@
...
+            <actionGroup ref="LoginActionGroup" stepKey="login"/>
+            <actionGroup ref="AdminAdobeStockSetConfigActionGroup" stepKey="setWrongCredentials">
+                <argument name="enabled" value="1"/>
+                <argument name="apiKey" value="wrong-api-key"/>
+                <argument name="privateKey" value="wrong-private-key"/>
+            </actionGroup>
+        </before>
+        <after>
+            <actionGroup ref="logout" stepKey="logout"/>
+        </after>
```
hws47a: After every test state should be returned back. I think in this case module should be disabled and keys are empty.

hws47a (changes requested):
In total, good job!

AdobeStockAdminUi/Test/Mftf/Test/AdminCannotAccessStockImagesWithWrongCredentialsTest.xml
sivaschenko: Thanks for the comment @hws47a 

Just some addition from my side:

Considering that module is enabled by default, I think it's ok to leave it enabled (or even remove enabling from the preconditions).

Test suite is responsible for preparing the configuration for all our MFTF tests (`AdobeStockImageAdminUi/Test/Mftf/Suite/adobe-stock-integration.xml`) so to return the test to inital state it will make sense to apply:
`<actionGroup ref="AdminAdobeStockSetConfigActionGroup" stepKey="setCorrectModuleConfig"/>` before the logout.

AdobeStockAdminUi/Test/Mftf/ActionGroup/AssertAdminFailedAuthenticationMessageActionGroup.xml
eduard13: Agree with this one 👍

AdobeStockAdminUi/Test/Mftf/Test/AdminCannotAccessStockImagesWithWrongCredentialsTest.xml
eduard13: Thank you @hws47a, I've added the reverting credentials step after all steps. 
Please note that I've left enabling the module in preconditions just to make sure that the module is enabled (as there can be other tests that disables it, and do not enables it back 🤔).

sivaschenko (approved):

PR #609, opened 2019-10-30
Enhance AdobeMediaGallery exception tests

peter279k:
N/A

The Travis CI build will be passed.

# Changed log
- Add `assertExceptionMessage` to assert exception message is same as result.


sivaschenko:
Hey @peter279k why do you think it's necessary to check for exception message?

In my opinion, it's enough to check for the correct exception class was thrown should be enough, not to make the unit test too fragile.

peter279k:
Hi @sivaschenko, thanks for your reply.

It seems to make sense after your explanation.

Closing this PR :).

Issue #603, opened 2019-10-29
Curated gallery filter is applied when accessing Adobe Stock listing the first time 


sivaschenko:
### Story:

https://github.com/magento/adobe-stock-integration/issues/344: [Stories #28] User sees the curated images when open Stock

### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)

### Expected result:

"Gallery Id: Kqy5H45lnKzkC5I8JGFXYVwr2ltNwMe7" filter is applied by default.
Images from https://stock.adobe.com/collections/Kqy5H45lnKzkC5I8JGFXYVwr2ltNwMe7 are displayed.

### Technical Details:

 - Add `SimpleFilters` configuration in `AdobeStockClient/etc/di.xml`
 - Apply the filter in `\Magento\AdobeStockImage\Model\GetImageList`, do not use `defaultFilters` configuration as this is non-permanent one-time filter

PR #619, opened 2019-11-01
Curated gallery filter is applied when open adobe stock first time

Nazar65:
### Description (*)

1.     Install Magento with Adobe Stock Integration
2.     Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

1. magento/adobe-stock-integration#603: Curated gallery filter is applied when open adobe stock first time

1. Open Adobe stock first time without filter verify images in grid -> https://stock.adobe.com/collections/Kqy5H45lnKzkC5I8JGFXYVwr2ltNwMe7
2. Chose other filters Apply this filters
3. Verify result  there is no gallery_id filter
4. Open adobe stock verify filters applied and gallery_id filter missed


AdobeStockImage/etc/di.xml
```diff
@@ -33,8 +33,9 @@
                     <item name="condition" xsi:type="string">eq</item>
                     <item name="field" xsi:type="string">ALL</item>
                 </item>
-            </argument>
-        </arguments>
+	  </argument>
+	  <argument name="defaultGalleryId" xsi:type="string">Kqy5H45lnKzkC5I8JGFXYVwr2ltNwMe7</argument>
```
sivaschenko: Please move this configuration to config.xml in AdobeStockImage module

AdobeStockImage/Model/GetImageList.php
```diff
@@ -40,42 +40,73 @@ class GetImageList implements GetImageListInterface
...
         return $this->getAssetList->execute($searchCriteria);
     }
 
+    /**
+     * Set Curated filter if not set other filters.
+     *
+     * @param SearchCriteriaInterface $searchCriteria
+     * @return SearchCriteriaInterface
+     */
+    private function setCuratedFilter(SearchCriteriaInterface $searchCriteria): SearchCriteriaInterface
```
sivaschenko: It's better to rename this method to match technical naming
```suggestion
    private function setDefaultGallery(SearchCriteriaInterface $searchCriteria): SearchCriteriaInterface
```

AdobeStockClient/etc/di.xml
```diff
@@ -63,7 +63,8 @@
                 <item name="colors_filter" xsi:type="string">setFilterColors</item>
                 <item name="serie_id" xsi:type="string">setSerieId</item>
                 <item name="model_id" xsi:type="string">setModelId</item>
-                <item name="premium_filter" xsi:type="string">setFilterPremium</item>
+	       	<item name="premium_filter" xsi:type="string">setFilterPremium</item>
```
sivaschenko: Indents have to be corrected

sivaschenko (changes requested):
Hi @Nazar65 thanks for working on this functionality! I have added some code review comments.

AdobeStockImage/etc/di.xml
Nazar65: Ok

AdobeStockImage/Model/GetImageList.php
Nazar65: Ok

AdobeStockClient/etc/di.xml
Nazar65: Ok

AdobeStockImage/Model/GetImageList.php
```diff
@@ -40,42 +47,83 @@ class GetImageList implements GetImageListInterface
...
      * @param GetAssetListInterface $getAssetList
      * @param FilterBuilder $filterBuilder
+     * @param ScopeConfigInterface
      * @param array $defaultFilters
      */
     public function __construct(
         FilterGroupBuilder $filterGroupBuilder,
         GetAssetListInterface $getAssetList,
         FilterBuilder $filterBuilder,
+        ScopeConfigInterface $scopeConfigInterface,
```
sivaschenko: Please create an `\Magento\AdobeStockImage\Model\Config` model and `\Magento\AdobeStockImageApi\Api\ConfigInterface` to handle image related configuration

AdobeStockImage/Model/GetImageList.php
```diff
@@ -40,42 +47,83 @@ class GetImageList implements GetImageListInterface
...
+                if ($filter->getField() === 'words') {
+                    $isSetWordsFilter = true;
+                }
+            }
+        }
+
+        if (!$isSetWordsFilter) {
+            $galleryFilter[] = $this->filterBuilder
+                ->setField('gallery_id')
+                ->setConditionType('like')
```
sivaschenko:
```suggestion
                ->setConditionType('eq')
``

**Review by sivaschenko — CHANGES_REQUESTED** [review]
Hi @Nazar65 thanks for the pull request! Please take a look at the review comments. Also could you please fix static test failures.

File: AdobeStockImage/Model/GetImageList.php (hunk 4 above)
Comment (Nazar65) [reply AdobeStockImage/Model/GetImageList.php]: ok

File: AdobeStockImageApi/Api/ConfigInterface.php [hunk 6]
```diff
@@ -0,0 +1,23 @@
...
+ * @api
+ */
+interface ConfigInterface
+{
+    /**
+     * Retrieve default gallery id filter value (gallery_id)
+     *
+     * @return string|null
+     */
+    public function getDefaultGalleryId():? string;
```
sivaschenko:
```suggestion
    public function getDefaultGalleryId(): ?string;
```

AdobeStockImage/Model/GetImageList.php
```diff
@@ -40,42 +41,83 @@ class GetImageList implements GetImageListInterface
...
+                    $isSetWordsFilter = true;
+                }
+            }
+        }
+
+        if (!$isSetWordsFilter) {
+            $galleryFilter[] = $this->filterBuilder
+                ->setField('gallery_id')
+                ->setConditionType('eq')
+                ->setValue($this->config->getDefaultGalleryId())
```
sivaschenko: Please check if default gallery id is set with `!empty($this->config->getDefaultGalleryId())` before applying the filter

AdobeStockImage/Model/Config.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * @var ScopeConfigInterface
+     */
+    private $scopeConfig;
+
+    /**
+     * @var UrlInterface
+     */
+    private $url;
```
sivaschenko: unnecessary property

AdobeStockImage/Model/Config.php
```diff
@@ -0,0 +1,52 @@
...
+
+use Magento\AdobeStockImageApi\Api\ConfigInterface;
+use Magento\Framework\App\Config\ScopeConfigInterface;
+
+/**
+ * Class Config
+ */
+class Config implements ConfigInterface
+{
+
```
sivaschenko: extra blank line

AdobeStockImage/Model/GetImageList.php
coderimus: Hi @sivaschenko @Nazar65 

I have an opinion that having such a model `Config`  maybe not so good because if it will handle the imaged related configuration it will become an ordinary helper which is not allowed. if it will store the one method only like today it has inconsistency name Config while being a service class it provides a config value.

I would advise 2 variants:
1. Rename this class according to its role - GetDefaultMediaId with one method execute and keep it under the API namespace
2. Rename this class according to is role - GetDefaultMediaId and place it under the Model\COnfig namespace. With this, we will define the config namespace and will be able to put any new configuration related implementation to it. Also, this covers a case when you consider this class Config as not a service API interface.

Some notes from standards:
`6.4.2.3. Each service interface SHOULD declare a single public method. An interface name SHOULD reflect the task or action to be performed. For example, Magento\InventoryApi\Api\StockSourceLinksDeleteInterface::execute(array $links). The only exception is a Repository API, which MAY be added for convenience and MUST be limited to singular CRUD operations and getList($searchCriteria).`

`6.4.1.1. Service contract interfaces SHOULD be placed in separate API modules, except when an existing module already contains Service Contracts in the Api folder. Other modules will depend on the API module, and implementations could be easily swapped via di.xml. API module names must end with the Api suffix. For example, if a module is named MyModule, its APIs SHOULD be declared in a module named MyModuleApi.
`

What do you think about this?

sivaschenko (changes requested):
Thanks for the updates! Please see the recent review comments

sivaschenko (approved):

Issue #555, opened 2019-10-19
Allow to save/download already licensed images to media gallery


sivaschenko:
### Story:

https://github.com/magento/adobe-stock-integration/issues/310: [Story 23] User saves previously licensed image into Magento Media Gallery

### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Login to Adobe Stock (use Sign In link in top right corner)
5. Find an image that is licensed for the Adobe Stock account but not licensed on actual Magento instance
6. Click on the image to expand the preview

### Expected result (*)
Only "Save" button is available for such image
Clicking "Save" button will bring up a popup to specify the image name (if the image preview was not saved before)
After confirming the image name a full licensed image is saved to the media gallery

### Additional Info
See https://github.com/magento/adobe-stock-integration/pull/552 PR - it is adding `is_licensed` filed to the search result

sivaschenko:
Hi @serhiyzhovnir do you have any updates regarding this issue?

drpayyne:
@sivaschenko the bot is auto assigning me and thinking PR is created for this issue as I referenced it in my PR as a dependency. But obviously, that isn't the case.

Issue #443, opened 2019-09-23
[MFTF] User saves licensed image into Magento Media Gallery


filmaj:
# Description
Implement MFTF test to cover scenario: "User saves licensed image into Magento Media Gallery" (story 23, #310)

#  Preconditions

1. Admin User is logged into Admin Panel
2. User signs into Adobe IMS

# Test steps

3. Open Adobe Stock Images Grid 
4. User expands image preview of first already-licensed image
5. User clicks ~Download~ Save button
6. Open Media Gallery
7. Verify that licensed image is displayed in the Media Gallery

Edit: ~Download~ Save

Issue #442, opened 2019-09-23
[MFTF] User views licensed images in the grid


filmaj:
# Description
Implement MFTF test to cover scenario: "User views licensed images in the grid" (story 22, #309)

#  Preconditions

1. Admin User is logged into Admin Panel
2. User signs into Adobe IMS

# Test steps

3. Open Adobe Stock Images Grid 
4. User expands image preview of first image
5. User clicks License and Save button
6. User clicks OK in the confirmation dialog
7. Open Adobe Stock Images Grid
8. Verify that first image has a licensed label

drpayyne:
Waiting on @sivaschenko to get back on how to obtain a new image every time. 

Also, two scenarios are considered here. One for preview saved already, and one for brand new image.

filmaj:
We shouldn't be licensing images during the tests, so with that in mind, I think we can consider this issue in its current form a duplicate of #440.

Perhaps we could modify this test such that:

- assume the Adobe IMS account used for the test has a pre-licensed image
- assume the Media Gallery does not have any saved images
- open the Stock search grid and search for the pre-licensed image
- validate the "licensed" label exists on the image
- validate the button for the image says "Save"
- click the Save button
- ensure the image is visible in the Media Gallery with the "licensed" label.

Thoughts?

drpayyne:
@filmaj one of the tests in #443 does the ‘saving’ functionality test for a prelicensed image. 

Actually, both. One is for preview saved and other, without.

filmaj:
OK, excellent. In that case I would say we close this issue in favour of #440 and proceed as you have detailed in your PRs.

drpayyne:
@filmaj, please proceed on closing this. Thank you.

sivaschenko:
@filmaj @drpayyne this task can be closed after the corresponding verification will be merged. If you'd like to add this verification to the test introduced in https://github.com/magento/adobe-stock-integration/pull/637 - please add this issue to the list of resolved issues in that PR

filmaj:
Actually, looking at this again, I think we need a dedicated test for:

1. loading the adobe stock grid as-is
2. assuming the test always uses a specific Adobe IMS account with a specific pre-licensed image in the account
3. searching for the pre-licensed image in the search bar via its file id
4. asserting the licensed label is there

filmaj:
@sivaschenko do you know which test / PR  closed this issue? I do not think this task is resolved.

filmaj:
[...]

Issue #446, opened 2019-09-23
[MFTF] User locates preview image in the Media Gallery


filmaj:
# Description
Implement MFTF test to cover scenario: "User locates preview image in the Media Gallery" (story 27, #329)

#  Preconditions

1. Admin User is logged into Admin Panel
2. User signs into Adobe IMS

# Test steps

3. Open Adobe Stock Images Grid 
4. User expands image preview of first already-preview-saved image
5. User clicks ~View Preview~ Locate button in Media Gallery button
6. Verify that preview image is displayed in the Media Gallery

drpayyne:
@filmaj @sivaschenko, I understand the story is different, and functionality too. But I feel the test for this and https://github.com/magento/adobe-stock-integration/issues/444 is the same. Both would be testing the _Locate_ button. And this button is irrespective of whether an image is licensed or not. If there's a local copy of image (whether it's a preview or a fully sized licensed), _Locate_ button appears and I feel the logic of the test is identical. 

I've already created a PR for https://github.com/magento/adobe-stock-integration/issues/444 at https://github.com/magento/adobe-stock-integration/pull/639. Please refer the test and let me know.

Thanks.

drpayyne:
With the generous help of @sivaschenko, I have come to this understanding of which button appears when and with what text in the image preview section. Also linked the MFTF test issues. So from the below understanding, I feel this is a duplicate of #444.

New image #440(in Media Gallery), #442(with label in Stock Grid)
Preview button text: Save Preview
License button text: License and Save

Already saved preview #440(in Media Gallery), #442(with label in Stock Grid)
Preview button text: Locate
License button text: License and Save

Already licensed on adobe stock (not saved locally) #443
No grey button
License button text: Save

Already saved preview and licensed in Adobe Stock (no licensed image locally) #443
Preview button text: Locate
License button text: Save

Licensed image locally #444
Preview button text: Locate
No red button

filmaj:
Hey @drpayyne, sorry, I'm a bit confused. Given your breakdown above, seems like  there are several interactions at play and each changes the expected button text as well as which buttons are displayed. In that sense, I feel there is value in having separate tests for these various scenarios.

Or am I misunderstanding  you?

drpayyne:
@filmaj Yes, indeed. A separate test for each scenario would be better, in my opinion. At the moment, only #440 and #442 would need to have two tests each. Already developed two tests for #443 and created a PR.

[...]

PR #637, opened 2019-11-04
MFTF automation for: User saves licensed image into Magento Media Gallery

drpayyne:
### Description

> **~On hold until https://github.com/magento/adobe-stock-integration/issues/555 is done. This is a TDD approach as suggested by @sivaschenko.~**

-  `AdminAdobeStockPreviewedLicensedImageSaveTest`: tests saving a previously licensed image **which already has a preview saved**.
-  `AdminAdobeStockNotPreviewedLicensedImageSaveTest`: tests saving a previously licensed image **which doesn't already have a preview saved**
-  Added `name` attribute for Locate button to obtain previously saved preview image's name in KO template
-  Two tests, since generated image name can be obtained from either prompt field (preview not saved) or Locate button's `name` attribute (preview saved already, so no prompt field while saving licensed image)
-  Both tests delete the licensed image after to return to the state before the test began
-  Fixed a couple of `stepKey` typos

### Fixed Issues
-  magento/adobe-stock-integration#443: [MFTF] User saves licensed image into Magento Media Gallery
-  magento/adobe-stock-integration#442: [MFTF] User views licensed images in the grid

### Manual testing scenarios

-  Add `magento/ADOBE_STOCK_LICENSED_IMAGE` credential with the image ID of a licensed image
-  `$ vendor/bin/mftf run:test AdminAdobeStockNotPreviewedLicensedImageSaveTest --remove`
-  `$ vendor/bin/mftf run:test AdminAdobeStockPreviewedLicensedImageSaveTest --remove`


AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockPreviewNotSavedLicensedImageSaveTest.xml
```diff
@@ -0,0 +1,46 @@
...
+            <actionGroup ref="AdminOpenMediaGalleryForPageNoEditorActionGroup" stepKey="openMediaGalleryForPage"/>
+            <actionGroup ref="AdminAdobeStockOpenPanelFromMediaGalleryActionGroup" stepKey="openAdobeStockPanel"/>
+            <actionGroup ref="AdminAdobeStockClickSignInActionGroup" stepKey="clickOnSignIn"/>
+            <actionGroup ref="AdminAdobeStockImsPopupSignInFillUserDataActionGroup" stepKey="fillUserCredentials"/>
+            <actionGroup ref="AdminAdobeStockImsPopupClickSignInActionGroup" stepKey="clickSignInImsPopup"/>
+            <actionGroup ref="AdminAdobeStockAssertUserLoggedActionGroup" stepKey="assertUserLoggedIn"/>
+        </before>
+        <after>
+            <actionGroup ref="AdminMediaGalleryDeleteImage" stepKey="deleteImageActionGroup">
+                <argument name="generatedImageName" value="//NEED TO PASS VALUE HERE"/>
```
drpayyne: Hi @sivaschenko, I need to pass the generated image name here to delete. And this test is when the preview image is not previously saved. So the generated image name is in another action group (AdminAssertAdobeStockImageSavedActionGroup). I don't think I can accommodate the image delete in that ActionGroup since it is used in multiple places. So do I create a new ActionGroup which will have the same functionality as the one mentioned above, but deletes the image after asserting that it is present in the Media Gallery

AdobeStockImageAdminUi/Test/Mftf/Section/AdobeStockSection.xml
```diff
@@ -21,7 +21,8 @@
         <element name="modalError" type="text" selector="//aside[contains(@class, 'modal-popup')]//div[@class='modal-content']//div[text()='Something went wrong.']"/>
         <element name="countGridImages" type="text" selector="//div[@class='masonry-image-grid']/div[contains(.,data-repeat-index)]"/>
         <element name="recordsFound" type="text" selector="//*[@id='adobe-stock-images-search-modal']//*[text()='records found']/parent::div"/>
-        <element name="selectedImagePreview" type="text" selector="//img[contains(@alt,'{{imageName}}')]" parameterized="true"/>
+        <element name="mediaGalleryImage" type="text" selector="//img[contains(@alt,'{{imageName}}')]" parameterized="true"/>
+        <element name="mediaGalleryDeleteButton" type="button" selector="#delete_files"/>
```
drpayyne, line 25: @sivaschenko, I think these _can_ be moved to another Section since these two are not really related to Adobe Stock, but rather the Magento Media Gallery.

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
```diff
@@ -351,6 +351,15 @@ define([
...
             return this.isDownloaded() ?  $.mage.__('License') : $.mage.__('License and Save');
+        },
+
+        /**
+         * Returns image name of previously saved preview image
+         *
+         * @returns {String}
+         */
+        getImageName: function () {
+            return this.preview().displayedRecord()['path'].split('/').pop();
```
drpayyne: @sivaschenko, this logic is based on an assumption and precaution that images can be nested in folders, so will have multiple directory slashes in the path. I can modify it to `return this.preview().displayedRecord()['path'].substring(1);` to remove the leading slash if necessary.

drpayyne:
For `AdminAdobeStockPreviewSavedLicensedImageSaveTest`, as discussed with @sivaschenko: https://github.com/magento/adobe-stock-integration/issues/446#issuecomment-549685367, _Locate_ button should be visible when image preview is saved and image is licensed in Adobe Stock (or anywhere else).

But this PR https://github.com/magento/adobe-stock-integration/pull/646/files#diff-d2bf191f817963fae506090c774bdd35R21 changes the visibility of the _Locate_ button to be visible only if image is downloaded AND NOT licensed.

This test is for saved preview image which is licensed elsewhere and according to @sivaschenko and myself, _Locate_ button should be present here. 

So, I'm needing some guidance here. Thanks!

AdobeStockImageAdminUi/Test/Mftf/Section/AdobeStockSection.xml
sivaschenko: @drpayyne that's a really good suggestion! Ideally, these selectors can be contributed to the CMS module of magento2 repository, but currently we have to keep them in our module. Moving to a separate section would be good.

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
sivaschenko: @drpayyne Images really can be nested in folders. Image are saved to Media Gallery folder from which Adobe Stock Integration panel is opened.

sivaschenko:
"But this PR https://github.com/magento/adobe-stock-integration/pull/646/files#diff-d2bf191f817963fae506090c774bdd35R21 changes the visibility of the Locate button to be visible only if the image is downloaded AND NOT licensed." - @drpayyne  good catch, thank you, this issue is fixed by https://github.com/magento/adobe-stock-integration/pull/649

diazwatson:
Git conflicts are now fixed

diazwatson:
I am getting the following error when running the test
```
Time: 31.07 seconds, Memory: 6.00MB

There was 1 error:

---------
1) AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest: Admin adobe stock preview not saved licensed image save test
 Test  tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:AdminAdobeStockPreviewNotSavedLicensedImageSaveTest

  [MalformedLocatorException] CSS or XPath locator is malformed: {{AdobeStockSection.mediaGalleryImage({{generatedImageName}})}}


Scenario Steps:

 18. $I->seeElement("{{AdobeStockSection.mediaGalleryImage({{generatedImage...") at tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:82
 17. // Entering Action Group [deleteImageActionGroup] AdminMediaGalleryDeleteImage
 16. $I->saveScreenshot() at tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:119
 15. $I->waitForElementVisible(".scalable.action-add-image.plugin",10) at tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:43
 14. $I->click("div[data-index=content]") at tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:42
 13. $I->waitForPageLoad(30) at tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:41

#1  Codeception\Module\WebDriver->seeElement
#2  /Users/rdiaz/Sites/adobestock/vendor/magento/magento2-functional-testing-framework/src/Magento/FunctionalTestingFramework/_generated/AcceptanceTesterActions.php:1932
#3  /Users/rdiaz/Sites/adobestock/dev/tests/acceptance/tests/functional/Magento/FunctionalTest/_generated/default/AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest.php:82
#4  Magento\AcceptanceTest\_default\Backend\AdminAdobeStockPreviewNotSavedLicensedImageSaveTestCest->_after

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

filmaj:
One question that needs tackling in here (as well as in #639) is the preconditions around the Adobe Stock account and its (pre)-licensed images. That is, what kind of assumptions can we make about the Adobe Stock account we will be using during testing? Will it come "seeded" with previously-licensed images? If not, will we have the ability to license images (and possibly incur costs on the account) at runtime on every test execution? It seems like this PR (and #639) expect the former situation, and require us hard-coding the licensed image ID ahead of time in an environment variable. This is possibly something we need to bring up with the Adobe Stock team.

drpayyne:
@filmaj True, and it is actually a block in two issues I'm assigned to. #440 and #442. Need some guidance on how to obtain a new image for every test run.

drpayyne:
Hi @diazwatson. Sorry about that. I overlooked that mistake. I've fixed it now. Will test locally and get back soon. Thank you.

AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockPreviewNotSavedLicensedImageSaveTest.xml
drpayyne: @filmaj could I get some guidance here, please? Thank you.

drpayyne:
Hey @sivaschenko and @diazwatson (@filmaj too if you're interested), couple of thoughts on these two tests regarding maintaining the Magento state.
`AdminAdobeStockPreviewNotSavedLicensedImageSaveTest`: After the test, will be deleting the saved image, among other things like logging out. And the state will be maintained since there was no image saved prior to the test (neither preview image nor licensed image).
`AdminAdobeStockPreviewSavedLicensedImageSaveTest`: Same steps as above to run after the test. Although deleting the saved licensed image, would also delete the preview image. Before the test, a preview image exists but after the test, no image exists. And doesn't fulfill the consistent state of Magento since test will fail on the subsequent run and this (functionally) transforms into the previous test mentioned above which has no preview image saved before the test.

Any suggestions on how to tackle this scenario?

Apologies for the long explanation, hope I was clear. If not, do let me know, I could demo it. Also note, I've changed test names and environment variable names for clarity from the previous commits. 

Thanks.

AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockNotPreviewedLicensedImageSaveTest.xml
```diff
@@ -0,0 +1,51 @@
...
+        <actionGroup ref="AdminSearchImagesOnModalActionGroup" stepKey="searchForNotPreviewedLicensedImageToSave">
+            <argument name="query" value="{{AdobeStockConfigDataNotPreviewedLicensedNotSavedImage.value}}"/>
+        </actionGroup>
+        <actionGroup ref="AdminAdobeStockExpandImagePreviewActionGroup" stepKey="expandLicensedImage"/>
+        <click selector="{{AdobeStockImagePreviewSection.saveLicensedImage}}" stepKey="clickSave"/>
+        <waitForPageLoad stepKey="waitForPromptModal" />
+        <grabValueFrom selector="{{AdobeStockImagePreviewSection.generatedImageName}}" stepKey="grabImageFileName" />
+        <click selector="{{AdobeStockImagePreviewSection.confirm}}" stepKey="clickOnPopupConfirm"/>
+        <waitForPageLoad stepKey="waitForMediaGalleryOpen" />
+        <seeElement selector="{{AdobeStockSection.mediaGalleryImage({$grabImageFileName})}}" stepKey="assertSavedImage" />
```
drpayyne: @diazwatson I used direct code here instead of the action group: `AdminAssertAdobeStockImageSavedActionGroup` as this test and the other test in this PR have the same selected actions, but in a slightly different order. So I'm wondering if I can decouple that action group, and reorder as necessary.

filmaj:
@drpayyne I think all tests should assume that there are no saved images in the media gallery, before and after test runs. So with that in mind, tests should ensure saving/deleting images as needed in the `before`/`after` blocks. What do you think?

drpayyne:
@filmaj sounds perfect. I'll implement that. Thank you!

diazwatson:
Hi @drpayyne how are you doing with the requested changes, is there anything we can help with?

drpayyne:
Hi @diazwatson, thank you. I'm going to proceed with what @filmaj suggested. For `AdminAdobeStockNotPreviewedLicensedImageSaveTest`, I'll go ahead as usual and for `AdminAdobeStockPreviewedLicensedImageSaveTest`, I'll use the same licensed image identifier, but will perform the _Save Preview_ flow in the `<before/>` node, so once the test starts, the preview is saved. And for both these tests, I'll delete the image file in the `<after/>` node. So that the state is maintained and consistent across multiple test runs.

drpayyne:
I've updated the code according to my comment above, @diazwatson. A thought: https://github.com/magento/adobe-stock-integration/pull/637#discussion_r343476418

AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockNotPreviewedLicensedImageSaveTest.xml
drpayyne: This is still an issue regarding the DRY principle. I can refactor `AdminAssertAdobeStockImageSavedActionGroup` into two action groups, and reuse them here, IF it is possible to pass `stepKey` data from one action group to another (need to pass image file name from the first action group to the other). So the above code would become:
```
<!-- The decoupled action groups -->
<actionGroup ref="AdminAdobeStockGrabImageNameActionGroup" name="grabImageName"/>
<actionGroup ref="AdminAssertAdobeStockImageSavedActionGroup" name="assertImageSave"/>
```
instead of the original coupled flow
```
<!-- The coupled action groups -->
<actionGroup ref="AdminAssertAdobeStockImageSavedActionGroup" name="assertImageSave"/>
```

AdobeStockImageAdminUi/Test/Mftf/Section/AdobeStockSection.xml
filmaj: @sivaschenko @drpayyne should I create an issue for moving the selectors out of this project and into magento2? Possibly as part of Milestone 3 (after the next 2.3.4 release)?

filmaj:
I'm trying this test w/ the latest develop branch and I'm not seeing images that I _think_ should be recognized as licensed showing up as licensed in the Stock search grid. Data returned from the Stock API in my case has empty data for the  `is_licensed` property.

Screen Shot 2019-11-07 at 1 07 41 PM

So tough for me to test this at this moment 😞

drpayyne:
@filmaj Can you check if it is licensed in the License History in stock.adobe.com?

AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockNotPreviewedLicensedImageSaveTest.xml
rogyar: I would suggest making the mentioned actions. Instead of using the direct actions in the test, it's better to create two action groups:
- `AdminAdobeStockGrabImageNameForSaveActionGroup`
- `AdminAssertImageInMediaGalleryActionGroup`

Please note the name of the second action group. It's not something related to AdobeStock, but the standard media gallery. There's a chance that something similar already exists, I would recommend to double-check it.

drpayyne: Thanks, @rogyar! I shall implement these changes.

rogyar:
Hi @drpayyne. Just a suggestion. Please, try to use only action groups in your tests and avoid direct operations call. It's one of the key points of MFTF Coverage Best practices. In that way, we create reusable components that might be helpful for other tests as well

drpayyne:
@rogyar yes, of course. I merely wanted to ask before decoupling an existing action group into two action groups for this particular scenario. Thanks!

filmaj:
I'm having the exact same issue as in #639 when running the test:

```
[searchForNotPreviewedLicensedImageToSave] AdminSearchImagesOnModalActionGroup
  [fillQueryInput] fill field "input.data-grid-search-control","193210330"
  [applySearch] click "#words + button.action-submit"
  [waitForSearchForQuery] wait for loading mask to disappear
[expandLicensedImage] AdminAdobeStockExpandImagePreviewActionGroup
  [clickOnThumbnail] click "aside.adobe-stock-modal [data-role=grid-wrapper] img[data-role=thumbnail]"
  [waitForImagePreviewToExpand] wait for loading mask to disappear
[clickSave] click "//button[@class='action-default primary']//span[text()='Save']"
[saveScreenshot] save screenshot
```

AdobeStockImageAdminUi/Test/Mftf/Section/AdobeStockImagePreviewSection.xml
```diff
@@ -8,7 +8,9 @@
 
 <sections xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:mftf:Page/etc/SectionObject.xsd">
     <section name="AdobeStockImagePreviewSection">
-        <element name="save" type="block" selector="//button[@class='action-secondary']/span[contains(.,'Save Preview')]"/>
+        <element name="save" type="block" selector="//button[@class='action-secondary']//span[text()='Save Preview']"/>
+        <element name="locateImage" type="block" selector="//button[@class='action-secondary']//span[text()='Locate')]"/>
+        <element name="saveLicensedImage" type="button" selector="//div[@class='actions']/descendant::span[text()='Save']"/>
```
drpayyne, line 13: Hi @filmaj, please refer: https://github.com/magento/adobe-stock-integration/pull/639#discussion_r346739377 as it is the same issue and fix.

drpayyne:
Although there is one problem which I just realized. Since we shouldn't be licensing new images for every test run (https://github.com/magento/adobe-stock-integration/issues/440#issuecomment-550007076), I use a prelicensed image in the credentials.
-  `AdminAdobeStockNotPreviewedLicensedImageSaveTest`: test is successful as the test should ensure a licensed but not previewed image _can_ be saved.
-  `AdminAdobeStockPreviewedLicensedImageSaveTest`: test is unsuccessful and I think not possible as, a licensed image will have just one button: _Save_. It isn't possible to save a preview of an already licensed image in the `<before/>` node. So this flow cannot be tested unless we take a random unlicensed image, save the preview, and license the image in the `<before/>` node, and then test if we can save this image.

filmaj:
We need to make sure the appropriate environment variable is set in travis' settings.

filmaj:
I have added the env var to travis settings, FYI

PR #647, opened 2019-11-04
DRY up the MFTF tests a little

filmaj:
### Description (*)

- removed duplicate action group for opening adobe stock search panel (`AdminAdobeStockOpenSearchModalActionGroup`, `AssertAdminUserHaveNoAccessToAdobeStockActionGroup`)
- make opening the media gallery action wait TWICE - once to load the gallery, the other time to load images (the media gallery shows two loading modals).
- the open media gallery action group now also supports handling CMS page with WYSIWYG editor disabled (by removing requirement to wait for show/hide button). this should support Magento instances with both WYSIWYG editor enabled or disabled.

This is the first in a series of PRs I will send as I am trying to get all of the MFTF tests passing locally. I appreciate any feedback!


AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockACLTest.xml
```diff
@@ -66,7 +66,8 @@
         </actionGroup>
 
         <!-- verify that user can access to adobe stock image -->
-        <actionGroup ref="AdminAdobeStockOpenSearchModalActionGroup" stepKey="openAdobeStockPopup"/>
+        <actionGroup ref="AdminOpenMediaGalleryForPageNoEditorActionGroup" stepKey="openMediaGalleryForPageWithAccess"/>
+        <actionGroup ref="AdminAdobeStockOpenPanelFromMediaGalleryActionGroup" stepKey="openAdobeStockPanelWithAccess"/>
```
filmaj, line 70: It would be nice if you could nest `actionGroup`s within an `actionGroup`, alas, I asked on the #mftf Slack channel and was told that is not possible. I think that would keep this much DRYer. Perhaps it is worth combining + adding into one new `actionGroup` which is just a copy-paste of both of the above `actionGroup`s, so that from any test, you could call e.g. `AdminAdobeStockOpenPanel` and it would open the add-new-CMS-page, expand content, click insert image, wait for the media gallery to load, and finally load up Adobe Stock search grid - but all in one action.

diazwatson (approved):

AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockLocalizationTest.xml
```diff
@@ -32,11 +32,13 @@
...
         </after>
         <!-- verify that user can search adobe stock image in ru_RU locale -->
-        <actionGroup ref="AdminAdobeStockOpenSearchModalActionGroup" stepKey="openSearchModal"/>
+        <actionGroup ref="AdminOpenMediaGalleryForPageNoEditorActionGroup" stepKey="openMediaGalleryForPage"/>
+        <actionGroup ref="AdminAdobeStockOpenPanelFromMediaGalleryActionGroup" stepKey="openAdobeStockPanel"/>
         <actionGroup ref="AdminSearchImagesOnModalActionGroup" stepKey="searchForPopular">
             <argument name="query" value="Автомобили"/>
         </actionGroup>
         <actionGroup ref="AdminAdobeStockExpandImagePreviewActionGroup" stepKey="expandImagePreview"/>
+        <!-- TODO: the following assertion/requirement on the specific keyword we expect wont work as we have no guarantee on the exact keyword to use as the image could be random -->
```
drpayyne, line 41: Hi @filmaj. My understanding is that you may not get the same image every time, right? But would it be the case even though you're searching for a specific keyword?  Maybe searching for an image ID stored in `.ENV` would help getting the same image every time? Maybe I'm understanding your comment wrong. Thanks!

filmaj: Hey @drpayyne, yes that  is correct. At least, that  was my experience when running the test locally - the images that came up with the search term would have various keywords and  did not match the keyword expectation in this test.

Maybe there is a better method of honouring the spirit of this test with a different approach. Ultimately what we want to achieve is that we want to validate that we receive results in the correct language/locale. Is there a different way we could go about it?

Maybe this is worth spinning out into its own issue?

PR #653, opened 2019-11-05
MFTF test tweaks, part two

filmaj:
### Description (*)

- Fixing up ImagePreviewAttributesTest: fixed xpath selector for image preview navigation buttons, removed inspecting dimensions and category from test (as consecutive images may actually end up being same size and category). Instead only assert on File #, which is guaranteed to be unique. Also added a wait for load after using navigation element.
- update stepkey to reflect actual search term used
- fixing AdminAdobeStockImagesPreviewNavigationTest: the image preview navigation controls are buttons, so the elements should be defined as buttons. also wait for the loading mask to disappear after using next/previous navigation in image preview section.
- while were at it, might as well ensure we wait for the loading mask to disappear every time we interact with image preview navigation controls

### DIFF
--- a/AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockImagePreviewAttributesTest.xml
+++ b/AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockImagePreviewAttributesTest.xml
@@ -41,20 +41,12 @@
         <actionGroup ref="AssertVisibleImagePreviewAttributesActionGroup" stepKey="verifyAttributeFile">
             <argument name="attributeName" value="File #"/>
         </actionGroup>
-        <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('Dimensions')}}" stepKey="getDimensionAttributeValue"/>
-        <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('File type')}}" stepKey="getFileTypeAttributeValue"/>
-        <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('Category')}}" stepKey="getCategoryAttributeValue"/>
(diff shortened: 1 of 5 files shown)


AdobeStockImageAdminUi/Test/Mftf/Test/AdminAdobeStockImagePreviewAttributesTest.xml
```diff
@@ -41,20 +41,12 @@
         <actionGroup ref="AssertVisibleImagePreviewAttributesActionGroup" stepKey="verifyAttributeFile">
             <argument name="attributeName" value="File #"/>
         </actionGroup>
-        <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('Dimensions')}}" stepKey="getDimensionAttributeValue"/>
-        <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('File type')}}" stepKey="getFileTypeAttributeValue"/>
-        <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('Category')}}" stepKey="getCategoryAttributeValue"/>
         <grabTextFrom selector="{{AdobeStockImagePreviewSection.attribute('File #')}}" stepKey="getFileAttributeValue"/>
         <click selector="{{AdobeStockImagePreviewSection.navigation('next')}}" stepKey="navigateToNextImage"/>
+        <waitForLoadingMaskToDisappear stepKey="waitForNextImageToLoad"/>
```
filmaj, line 46: The above two lines may be worth wrapping into an actiongroup, maybe? Since whenever we click on the image preview section's navigation controls, the loading mask will appear, so clicking  on those navigation controls should always be paired with waiting on the loading mask to disappear, IMO.

Nazar65 (approved):

drpayyne (approved):

sivaschenko (approved):

PR #655, opened 2019-11-06
Made test's <after/> node stateless

drpayyne:
### Description

Made `<after/>` node stateless 

### Manual testing scenarios

No matter where `AdminAdobeStockSignedInCreditsVisibleTest` fails, the `<after/>` node won't fail as it doesn't depend on where the test failed.


filmaj (approved):
LGTM!

PR #668, opened 2019-11-08
Removed core modules

sivaschenko:
Removed core modules

Merge only after modules will be introduced to core


sivaschenko:
@magento run all tests

Issue #641, opened 2019-11-04
The licensed label is shown under the image if the row of images has small height


sivaschenko:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
3. Wide image is licensed

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Log in to Adobe Stock using Sign in link
4. Find the wide licensed image

### Actual result

The licensed label is shown under the image if the row of images has small height

![image](https://user-images.githubusercontent.com/2028541/68123731-aa625080-ff05-11e9-959d-a71e18fb1fb1.png)

drpayyne:
Hi @sivaschenko. Where should the label be placed? stock.adobe.com places it on the top left:
![image](https://user-images.githubusercontent.com/22623301/68145156-2aff6c00-ff5b-11e9-92ca-5578e4334552.png)

sivaschenko:
@drpayyne the label should be placed according to the mockup:
![image](https://user-images.githubusercontent.com/2028541/68236906-f7295280-fffd-11e9-92b2-f9d7e05d3732.png)

drpayyne:
@sivaschenko Thanks, will do it.

sivaschenko:
@drpayyne the PR was reverted. The solution causes a bug that makes license labels for the row to disappear when the image preview is opened.

drpayyne:
Hi @sivaschenko. Sorry about that. It's my fault that I overlooked that scenario. Have fixed it now in PRs: https://github.com/magento/adobe-stock-integration/pull/670 and https://github.com/magento/magento2/pull/25525

PR #670, opened 2019-11-08
Resolved issue 641

drpayyne:
### Description 
This PR solves the below mentioned issue by programmatically calculating the top displacement from the respective image height. Prevents the previously caused bug by using `top` instead of `bottom` as image container height increases to accommodate opened preview. Also, inspect overlay label container to notice `top` displacement vary according to respective image height.

Two more required files are present in magento/magento2 repository's PR: https://github.com/magento/magento2/pull/25525

### Fixed Issues
1. magento/adobe-stock-integration#641: The licensed label is shown under the image if the row of images has small height

### Manual testing scenarios
1. Open wide licensed images in Adobe Stock grid

### DIFF
--- a/AdobeStockImageAdminUi/etc/module.xml
+++ b/AdobeStockImageAdminUi/etc/module.xml
@@ -6,5 +6,9 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Magento_AdobeStockImageAdminUi" />
+    <module name="Magento_AdobeStockImageAdminUi">
+        <sequence>
+            <module name="Magento_Ui"/>
+        </sequence>
+    </module>
 </config>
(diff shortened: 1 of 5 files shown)


sivaschenko:
Hi @drpayyne , thanks for the pull request! I think it's better to extend the overlay component from magento2 core and and implement `getStyles` method in AdobeStockImageAdminUi module as these styles are specific to Adobe Stock Integration grid

drpayyne:
Hi @sivaschenko, have done the required changes. Thanks!

AdobeStockImageAdminUi/etc/module.xml
```diff
@@ -6,5 +6,9 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Magento_AdobeStockImageAdminUi" />
+    <module name="Magento_AdobeStockImageAdminUi">
```
sivaschenko: I don't think there is need to add `Magento_Ui` module to sequence, there is no hard installation or configuration dependency. Soft dependency should be enough (composer.jsoon)

drpayyne: @sivaschenko, I have fixed it. Thanks!

sivaschenko (approved):

AdobeStockImageAdminUi/etc/module.xml
sivaschenko: Thank you!

Issue #669, opened 2019-11-08
Licensed overlay label is missing for just licensed images


sivaschenko:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. License an Image
5. Reopen the grid

### Expected result

Licensed image has Lisensed overlay label

### Actual result

Licensed image does not have Lisensed overlay label

PR #674, opened 2019-11-08
#669 : Licensed overlay label is missing for just licensed images

konarshankar07:
### Description (*)

This PR will update the license label if the image is just licensed and also fixed small css issue with save preview button

1. magento/adobe-stock-integration#669: Licensed overlay label is missing for just licensed images

1. Login to admin panel
4. License an Image
5. Reopen the grid
### Expected Result
License label will be visible for the recently licensed label

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
@@ -29,7 +29,8 @@ define([
             messageDelay: 5,
             modules: {
                 login: '${ $.loginProvider }',
-                preview: '${ $.parentName }.preview'
+                preview: '${ $.parentName }.preview',
+                source: '${ $.provider }'
             }
         },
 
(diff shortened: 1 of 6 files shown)


AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
```diff
@@ -142,7 +143,8 @@ define([
                         displayedRecord['is_licensed'] = 1;
                         displayedRecord['is_licensed_locally'] = 1;
                     }
-
+                    this.source().set('params.t ', Date.now());
+                    this.preview().hide();
```
sivaschenko: The preview should not be hidden after saving the image

sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request! There is only one comment

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
konarshankar07: If we don't close the preview then we'll get the `record.series is not a function` error. So what should be the best solution to overcome this issue?

PR #680, opened 2019-11-10
Fix for issues found with PHPStan (Level 2)

hws47a:
### Description (*)
I've run PHPStan (Level 2) check and found these errors.

```
------ ---------------------------------------------------------------------------- 
  Line   module-adobe-ims-api/Api/Data/UserProfileInterface.php                      
 ------ ---------------------------------------------------------------------------- 
  84     PHPDoc tag @return with type void|null is not subtype of native type void.  
 ------ ---------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------------------- 
  Line   module-adobe-ims/Controller/Adminhtml/OAuth/Callback.php                                      
 ------ ---------------------------------------------------------------------------------------------- 
  110    Call to an undefined method Magento\Backend\Model\Auth\Credential\StorageInterface::getId().  
  155    Call to an undefined method Magento\Backend\Model\Auth\Credential\StorageInterface::getId().  
 ------ ---------------------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------- 
  Line   module-adobe-ims/Controller/Adminhtml/User/Logout.php                              
 ------ ----------------------------------------------------------------------------------- 
  52     Call to an undefined method Magento\Framework\HTTP\Client\CurlFactory::execute().  
 ------ ----------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------- 
  Line   module-adobe-ims/Model/GetToken.php                                                         
 ------ -------------------------------------------------------------------------------------------- 
  88     Call to method getAccessToken() on an unknown class Magento\AdobeIms\Model\TokenResponse.   
  88     Call to method getRefreshToken() on an unknown class Magento\AdobeIms\Model\TokenResponse.  
  90     Call to method getError() on an unknown class Magento\AdobeIms\Model\TokenResponse.         
 ------ -------------------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------- 
  Line   module-adobe-ims/Test/Unit/Model/GetImageTest.php                
 ------ ----------------------------------------------------------------- 
  111    Call to an undefined method Psr\Log\LoggerInterface::expects().  
 ------ ----------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------- 
  Line   module-adobe-ims/Test/Unit/Model/UserProfileRepositoryTest.php                         
 ------ --------------------------------------------------------------------------------------- 
[...]

PR #679, opened 2019-11-10
Fixed issue found by PHPStan level 1

hws47a:
### Description (*)
I've analysed the code with PHPStan (level 1) and found the following issues:
```
 ------ ------------------------------------------------------------------ 
  Line   module-adobe-stock-asset/Test/Api/AssetRepository/DeleteTest.php  
 ------ ------------------------------------------------------------------ 
  55     Constant TESTS_WEB_API_ADAPTER not found.                         
  75     Constant TESTS_WEB_API_ADAPTER not found.                         
  77     Variable $notExistedAssetId might not be defined.                 
  79     Constant TESTS_WEB_API_ADAPTER not found.                         
  106    Constant TESTS_WEB_API_ADAPTER not found.                         
 ------ ------------------------------------------------------------------ 

 ------ ------------------------------------------------------------------- 
  Line   module-adobe-stock-asset/Test/Api/AssetRepository/GetByIdTest.php  
 ------ ------------------------------------------------------------------- 
  73     Constant TESTS_WEB_API_ADAPTER not found.                          
  80     Constant TESTS_WEB_API_ADAPTER not found.                          
  85     Constant TESTS_WEB_API_ADAPTER not found.                          
  115    Constant TESTS_WEB_API_ADAPTER not found.                          
 ------ ------------------------------------------------------------------- 
```

Issue with 'TESTS_WEB_API_ADAPTER' happens because constant is defined using xml file. I've wrapped it in `constant()` call.

Another issue fixed with a possibility of having an undefined variable.

### Manual testing scenarios (*)
Run API Functional tests

PR #681, opened 2019-11-10
Add PHPStan (level 2) to travis CI

hws47a:
### Description (*)
After improvements made by PHPStan checks it could good to keep quality and add PHPStan checks into travis CI.

Before this one, these two pull requests must be merged before:
- #679 
- #680

### DIFF
--- a/AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
+++ b/AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
@@ -52,7 +52,7 @@ class DeleteTest extends WebapiAbstract
     {
         $response = $this->deleteAsset($this->getAssetId());
 
-        if (TESTS_WEB_API_ADAPTER === self::ADAPTER_REST) {
+        if (constant('TESTS_WEB_API_ADAPTER') === self::ADAPTER_REST) {
             $this->assertSame([], $response);
         } else {
             $this->assertNull($response);
@@ -72,11 +72,11 @@ class DeleteTest extends WebapiAbstract
             $this->deleteAsset($notExistedAssetId);
             $this->fail('Expected throwing exception');
         } catch (\Exception $e) {
-            if (TESTS_WEB_API_ADAPTER === self::ADAPTER_REST) {
+            if (constant('TESTS_WEB_API_ADAPTER') === self::ADAPTER_REST) {
                 $errorData = $this->processRestExceptionResult($e);
                 self::assertEquals($notExistedAssetId, $errorData['parameters'][0]);
                 self::assertEquals(Exception::HTTP_NOT_FOUND, $e->getCode());
-            } elseif (TESTS_WEB_API_ADAPTER === self::ADAPTER_SOAP) {
+            } elseif (constant('TESTS_WEB_API_ADAPTER') === self::ADAPTER_SOAP) {
                 $this->assertInstanceOf('SoapFault', $e);
             } else {
                 throw $e;
@@ -103,7 +103,7 @@ class DeleteTest extends WebapiAbstract
             ],
         ];
 
-        return (TESTS_WEB_API_ADAPTER === self::ADAPTER_SOAP)
+        return (constant('TESTS_WEB_API_ADAPTER') === self::ADAPTER_SOAP)
             ? $this->_webApiCall($serviceInfo, ['id' => $assetId])
             : $this->_webApiCall($serviceInfo);
     }
--- a/AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
+++ b/AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
@@ -30,7 +30,7 @@ class DeleteTest extends WebapiAbstract
     private $objectManager;
 
     /**
-     * @var Collection
+     * @var CollectionFactory
      */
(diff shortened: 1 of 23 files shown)


AdobeStockImage/Test/Unit/Model/SaveImageTest.php
```diff
@@ -180,9 +181,9 @@ public function assetProvider(): array
      * Get document
      *
      * @param string|null $path
-     * @return Document|MockObject
+     * @return MockObject
```
sivaschenko, line 184: I think `Document|MockObject` is correct param type here

AdobeStockImageAdminUi/Test/Unit/Model/Listing/DataProviderTest.php
```diff
@@ -143,9 +143,9 @@ public function testGetDataException(): void
 
     /**
      * @param array $itemsData
-     * @return SearchResultInterface|MockObject
+     * @return MockObject
```
sivaschenko, line 146:
```suggestion
     * @return SearchResultInterface|MockObject
```

AdobeStockImageAdminUi/Test/Unit/Ui/Component/Listing/Filter/ColorTest.php
```diff
@@ -177,7 +177,7 @@ public function testPrepare(?string $colorPickerMode, string $appliedValue): voi
         $color->prepare();
     }
 
-    private function verifyApplyFilter(string $appliedValue, Filter $filter, ContextInterface $context): void
+    private function verifyApplyFilter(string $appliedValue, Filter $filter, MockObject $context): void
```
sivaschenko, line 180:
```suggestion
    private function verifyApplyFilter(string $appliedValue, Filter $filter, ContextInterface $context): void
```
Almost every object in the test is `MockObject`, `ContextInterface` is more specific type

sivaschenko (changes requested):
Hi @hws47a , thanks for the pull request! Looks nice, however, I am not sure about `Type -> MockObject` changes, what do you think?

AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
```diff
@@ -67,16 +67,16 @@ public function testDelete(): void
...
     public function testDeleteWithException(): void
     {
+        $notExistedAssetId = -1;
         try {
-            $notExistedAssetId = -1;
             $this->deleteAsset($notExistedAssetId);
             $this->fail('Expected throwing exception');
         } catch (\Exception $e) {
-            if (TESTS_WEB_API_ADAPTER === self::ADAPTER_REST) {
+            if (constant('TESTS_WEB_API_ADAPTER') === self::ADAPTER_REST) {
```
sivaschenko, line 75: I don't think that constant('TESTS_WEB_API_ADAPTER') is better than TESTS_WEB_API_ADAPTER.

Is there any way to point phpstan to the constant declaration?

sivaschenko:
@hws47a I think it would be better to merge:
https://github.com/magento/adobe-stock-integration/pull/679
https://github.com/magento/adobe-stock-integration/pull/680
https://github.com/magento/adobe-stock-integration/pull/681

into a single PR

AdobeStockImage/Test/Unit/Model/SaveImageTest.php
hws47a: PHPStan validation is failing here. 
Basically return type hint should be the same as @return tag in DocBlock.
So I had to choose between MockObject and Document. I decided to use MockObject because MockObject's interface is used after getting `getDocument` result.

AdobeStockImageAdminUi/Test/Unit/Ui/Component/Listing/Filter/ColorTest.php
hws47a: It's not 100% correct. ContextInterface is not inheriting MockObject and vice versa. 
So you have to choose one. Actually "MockObject" methods are used inside this method, that's why it's more correct to use MockObject here.
`ContextInterface` doesn't have `expects` method

AdobeStockAsset/Test/Api/AssetRepository/DeleteTest.php
hws47a: There is no TESTS_WEB_API_ADAPTER declaration in PHP.
So, in any case, PHPStan wouldn't find it. PHPStorm validates it as undefined as well. Actually wrap it in `constant()` is a good idea because if it would be actually undefined `constant()` call will return `null` while raw `TESTS_WEB_API_ADAPTER` will throw php warning or error.

hws47a:
@sivaschenko no problem with merging together. We can just close two other MR's

AdobeStockImageAdminUi/Test/Unit/Model/Listing/DataProviderTest.php
hws47a: Probably same as above or under

hws47a:
@sivaschenko I've closed two other CR's. Anyway, they have a description of all issues found which can be used for the review.

sivaschenko (approved):

PR #701, opened 2019-11-12
#704 Add check for non strict_types=1 files

hws47a:
Ticket #704 

### Description
Core Magento LiveCodeTest.php file tests files for having `strict_types=1` in changed files. But there is no way to check all codebase or provided folder. I've created a script to find all PHP files in our repository and check if they have these lines.

After setting up check it was found that 17 files didn't have the required declaration and they were fixed in this pull request.

Example of a job fail when files without strict_types=1 found: https://travis-ci.org/magento/adobe-stock-integration/jobs/613269038

Example of a success job once all files fixed: https://travis-ci.org/magento/adobe-stock-integration/jobs/613269754


diazwatson:
Hi @hws47a thanks for working to improve the quality of the code
Could you update this PR with those 17 files fixed?

hws47a:
Hi @diazwatson, I need these files to be unchanged, so I can test Travis CI pipeline.
My idea is to fix them as soon as the pipeline will work fine.

As you see now we have broken files but the pipeline is not failing.

sivaschenko (approved):
Great job!

Issue #644, opened 2019-11-04
Overflowing text in sort dropdown


drpayyne:
### Preconditions
1.  Open Adobe Stock panel

### Steps to reproduce
1. View sorting dropdown

### Expected result (*)
Properly rendered text

### Actual result (*)
![image](https://user-images.githubusercontent.com/22623301/68144947-b5939b80-ff5a-11e9-952a-7a3371c1fb35.png)
![image](https://user-images.githubusercontent.com/22623301/68144980-c3492100-ff5a-11e9-8d90-ca9e9b6463e0.png)

###  Expected Result:
![image](https://user-images.githubusercontent.com/2028541/68715378-c4510280-0599-11ea-83d1-a560258f0b27.png)

drpayyne:
I'm working on this, thanks.

sivaschenko:
@drpayyne Expected Result:
![image](https://user-images.githubusercontent.com/2028541/68715378-c4510280-0599-11ea-83d1-a560258f0b27.png)

drpayyne:
Closing this, as the PR revert in magento2 repository is merged to fix this issue. Thanks!

PR #707, opened 2019-11-13
Resolved issue 644

drpayyne:
### Description
- Added extra padding to the right and increased container size by one point to accomodate extra padding. This doesn't lead to any visible container reorganization. Just extra padding

### Fixed Issues
-  magento/adobe-stock-integration#644: Overflowing text in sort dropdown

### Manual testing scenarios
1.  Open Adobe Stock panel
2.  Verify sort options text don't overflow for all options


sivaschenko:
Hi @drpayyne thanks for the pull request! Can you please open the pull request from your fork instead of creating branches in https://github.com/magento/adobe-stock-integration repository.

Also, pay attention to the expected result in the issue: the actual issue is not the overflow but the border and background colour of dropdown arrow.

PR #721, opened 2019-11-15
First pass at running MFTF tests in travis!

### LINKED ISSUE #321: Add MFTF tests to Travis build
Add MFTF tests to Travis build

# TODO

- [x] fix pathvalidator issue; try not using the `ext/` dir and instead place module code directly in `app/code/Magento/.`
- [x] need to store credentials in MFTF appropriately, filed in #698
- [x] enable https on the travis magento instance
- [x] make sure the adobe.io integration api + private key used in travis has specified the correct redirect uri (matches domain used in travis) - this will likely only be an issue for the signinsignout and signincreditsvisible tests (since those log into adobe ims)
- [x] MFTF preparedness of magento2 instance needs to be done
- tests run a long time, how to optimize?
  - [x] split up test runs across different builds using a build matrix?
  - get more detailed info on timing for each test - filed issue in https://github.com/magento/magento2-functional-testing-framework/issues/508
  - can we save any time on startup? ~5 mins are spent just setting the magento instance up.
    - [x] cache any downloads, i.e. chromedriver?
    - [x] to use caching, ensure 'build pushed branches' setting on travis is ON (see note in top of docs [here](https://docs.travis-ci.com/user/caching/))

filmaj:
### Description (*)

Lots in here:

- stands up a magento instance running over https on travis, with the instance prepped for MFTF test runs
- to fit under the travis maximum time limit for a job of 50mins, I had to create several new MFTF Suites and assign the tests out to different suits, roughly bucketing about  6 tests per suite, in order for each suite to not run over the 50 min limit on travis.
- I set for the MFTF tests to only run on PHP 7.3, hopefully that is a good assumption

Still TODO before we merge this:

- [x] ensure the correct env vars are in place in the travis settings for this repo
- [x] ensure the api key/private key we use for the console.adobe.io integration has the correct travis-specific redirect URI set (the magento instance running in travis uses https://magento2.travis)

This PR fixes #321


.travis.yml
```diff
@@ -8,51 +8,104 @@ addons:
...
+  - mv Adobe* magento2/app/code/Magento/.
+  - pushd magento2
   - composer install
-  - composer config minimum-stability dev
-  - composer config repositories.ext path "../ext/*/*/*"
-  - composer require magento/adobe-stock-integration
+  - composer require astock/stock-api-libphp
+  - popd
 before_script:
-  - cp ../ext/magento/adobe-stock-integration/dev/tests/unit/phpunit.xml dev/tests/unit/phpunit.xml
```
sivaschenko, line 50: It would be better to copy our custom `phpunit.xml` to the default location for easier unit tests launch

.travis.yml
```diff
@@ -8,51 +8,104 @@ addons:
...
-  - cp -f ../ext/magento/adobe-stock-integration/dev/travis/before_script.sh dev/travis/before_script.sh
-  - ./dev/travis/before_script.sh
+  - bash -x dev/travis/before_script.sh
```
sivaschenko, line 91: `-x` 👍

.travis.yml
```diff
@@ -8,51 +8,104 @@ addons:
...
-  - if [ $TEST_SUITE == 'static' ]; then phpcs --standard=dev/tests/static/framework/Magento/ ../ext ; fi
+  - pushd magento2
+  - if [ $TEST_SUITE == 'unit' ]; then vendor/bin/phpunit --configuration dev/tests/unit/phpunit.xml; fi
+  - if [ $TEST_SUITE == 'phpstan' ]; then composer require --dev phpstan/phpstan fooman/phpstan-magento2-magic-methods && vendor/bin/phpstan analyse -l 2 ../ext -a dev/tests/api-functional/framework/autoload.php ; fi
+  - if [ $TEST_SUITE == 'static' ]; then vendor/bin/phpcs --standard=dev/tests/static/framework/Magento/ ../ext ; fi
+  - if [ $TEST_SUITE == 'functional' ]; then echo "Running MFTF suite ${MFTF_SUITE}" && vendor/bin/mftf run:group $MFTF_SUITE --remove -vvv; fi
 after_success:
   - if [ $TEST_SUITE == 'unit' ]; then cd ../ && travis_retry coveralls; fi
+after_failure:
+  - if [ $TEST_SUITE == 'functional' ]; then cat "${HOME}/drivers/*.log"
```
sivaschenko: I think it would be good to also output `<magento>/var/log/exception.log`, `<magento>/var/log/system.log` and anything in `<magento>/var/reports/*` after failure

dev/travis/config/apache_virtual_host
```diff
@@ -0,0 +1,43 @@
...
+    SSLCertificateFile %TRAVIS_BUILD_DIR%/%MAGENTO_HOST_NAME%.cert
+    SSLCertificateKeyFile %TRAVIS_BUILD_DIR%/%MAGENTO_HOST_NAME%.key
+
+    <Directory "%MAGENTO_ROOT%">
+        Options FollowSymLinks MultiViews ExecCGI
+        AllowOverride All
+        Require all granted
+    </Directory>
+
+    <IfModule mod_fastcgi.c>
```
sivaschenko, line 34: Why do we need fastcgi?

dev/travis/before_install.sh
```diff
@@ -23,4 +23,56 @@ phpenv rehash;
...
+    sudo service apache2 restart
+
+    # Install chromedriver to power chrome
+    if [ ! -d "${HOME}/drivers" ]; then
+        mkdir -p "${HOME}/drivers"
+    fi
+    if [ ! -f "${HOME}/drivers/chromedriver" ]; then
+        pushd "${HOME}/drivers"
+        # TODO: factor out the version to travis env var
+        wget http://chromedriver.storage.googleapis.com/78.0.3904.70/chromedriver_linux64.zip
```
sivaschenko: We are using the latest stable chrome. Is there a dynamic link to the latest stable chromedriver?

.travis.yml
```diff
@@ -8,51 +8,104 @@ addons:
...
+    - php: '7.1'
+      env: TEST_SUITE=functional MFTF_SUITE=adobe_stock_integration_suite_configuration
+    - php: '7.1'
+      env: TEST_SUITE=functional MFTF_SUITE=adobe_stock_integration_suite_keywords
+    - php: '7.1'
+      env: TEST_SUITE=functional MFTF_SUITE=adobe_stock_integration_suite_ims
+    - php: '7.1'
+      env: TEST_SUITE=functional MFTF_SUITE=adobe_stock_integration_suite_see_more
+    - php: '7.2'
+      env: TEST_SUITE=functional MFTF_SUITE=adobe_stock_integration_suite_filters_part_one
```
sivaschenko: The fix of the issue slowing down the tests is now merged to `2.3-develop`. Could you check if we could get rid of `part_*` test suites on the latest code base?

AdobeStockImageAdminUi/Test/Mftf/Suite/adobe-stock-integration-configuration.xml
```diff
@@ -0,0 +1,22 @@
...
+  * See COPYING.txt for license details.
+  */
+-->
+
+<suites xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="../../../../../../vendor/magento/magento2-functional-testing-framework/src/Magento/FunctionalTestingFramework/Suite/etc/suiteSchema.xsd">
+    <suite name="adobe_stock_integration_suite_configuration">
+        <before>
+            <actionGroup ref="LoginAsAdmin" stepKey="loginAsAdmin"/>
+            <actionGroup ref="AdminAdobeStockSetConfigActionGroup" stepKey="setCorrectModuleConfig"/>
```
sivaschenko: I think that configuration tests suite should not have the configuration as a precondition.

sivaschenko (changes requested):

sivaschenko:
@filmaj well done! See my review comments. Also, can you please update paths for unit/static tests to pass

.travis.yml
```diff
@@ -8,51 +8,104 @@ addons:
...
 language: php
 php:
   - '7.1'
   - '7.2'
   - '7.3'
 env:
   global:
-    - COMPOSER_BIN_DIR=~/bin
+    # we should revisit the nodejs version. Long Term Stable (LTS) is version 12 now.
     - NODE_JS_VERSION=8
```
filmaj: @sivaschenko what do you think of bumping the nodejs version to 12? What do we use node for, anyways?

dev/travis/before_install.sh
```diff
@@ -23,4 +23,56 @@ phpenv rehash;
 test -n "$GITHUB_TOKEN" && composer config github-oauth.github.com "$GITHUB_TOKEN" || true
 
 composer config --global http-basic.repo.magento.com "$MAGENTO_USERNAME" "$MAGENTO_PASSWORD"
-git clone https://github.com/magento/magento2
+# TODO: we should specify a particular SHA/branch here
+git clone --depth 1 https://github.com/magento/magento2
```
filmaj, line 27: @sivaschenko I think we need to narrow down which version/SHA of `magento2` we clone/use for functional tests. Do you have any opinion on this?

dev/travis/config/apache_virtual_host
filmaj: I just copied the travis virtual host config from the main magento2 repo, but expanded it to add a virtual host for use over HTTPS  https://github.com/magento/magento2/blob/2.3-develop/dev/travis/config/apache_virtual_host

filmaj: As for "why do we need fastci?" - I'll be honest, I have no idea! If there's a better option I am all ears 😄

.travis.yml
sivaschenko: I think we used it for eslint tests, as we are running them on Jenkins now, I think we can remove node dependency

dev/travis/before_install.sh
sivaschenko: I think we should use the latest one

dev/travis/config/apache_virtual_host
sivaschenko: I think this block can be removed, am I wrong?

filmaj: Will test it now!

filmaj: Confirmed that without this module, PHP does not get processed and HTTP(s) requests  to the server return raw PHP code instead of HTML. So, I think it is necessary, at least in this current setup 🤷‍♂ I am not an expert in Apache or PHP and don't know what a "typical" PHP setup looks like so I don't trust  myself in this question nor should anyone else trust me!

filmaj:
@sivaschenko I think this is ready for another review. One outstanding issue is that unit tests  report green even though they fail due to a whitelist/configuration issue in PHPunit.

filmaj:
Going to close this and open from the origin so we can take a look at  the test runs with proper encrypted env vars in place.

Issue #678, opened 2019-11-10
Image file name is not visible if the title is in other than English language


konarshankar07:
### Preconditions (*)

1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)

1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Enter text `246811446` in search by keywords
5. Click on the image to open the image preview
6. Click on Save Preview button

### Expected result (*)

File id is used as the suggested filename

### Actual result (*)

![image](https://user-images.githubusercontent.com/39480008/68548195-22ea7500-0410-11ea-8f38-91a1ea636091.png)

sivaschenko:
Let's add `File id` suffix to the end of the generated image name and remove `-` if there are no characters in result image name without suffix.

So Expected result:
`File id` is used as the suggested filename

@tkacheva what do you think?

PR #722, opened 2019-11-16
[WIP] Use image ID as a proposed filename for non-latin titles

rogyar:
### Description (*)
This PR introduces a slightly changed logic for media gallery file name generation. In case if an image from Adobe Stock does not have latin characters in its title, the image ID will be proposed as a file name. 

1. #678 : Image file name is not visible if the title is in other than English language

1. Login to admin panel
4. Enter text `246811446` in search by keywords
5. Click on the image to open the image preview
6. Click on Save Preview button

The image ID should be proposed as the file name.

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
@@ -177,10 +177,13 @@ define([
          * @return string
          */
         generateImageName: function (record) {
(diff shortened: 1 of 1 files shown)


sivaschenko:
@rogyar thank you for the pull request, it looks good! Is the PR WIP? Would you like to add anything to the PR?

The solution is very specific, but that's whats we come up with and I don't see the better approach at the moment to handle file names considering current Magento (media gallery) implementation.

sivaschenko (approved):

Issue #720, opened 2019-11-15
Save a licensed image, then delete, then open grid, 'locate' button is shown on preview


filmaj:
### Preconditions (*)

1. Load up the Adobe Stock grid and sign into an Adobe IMS account with a licensed image

### Steps to reproduce (*)

1. Expand the preview for a licensed image
2. Save the licensed image to the media gallery
3. Select the saved image and delete it from the media gallery
4. Open the adobe stock grid once more, the preview for the previously-saved image should already be open

### Expected result (*)

1. In the expanded preview of the saved-and-deleted-image, button should have 'Save' text

### Actual result (*)

1. In the expanded preview of the saved-and-deleted-image, button has 'Locate' text
2. If you click this button, you get an error: 'cannot find image in media gallery'

PR #727, opened 2019-11-18
[WIP] Add ability to get the image asset data by request

coderimus:
### Description (*)

The original problem, requested at the #720 , caused by next: we didn't update image asset information when open/switch between images. As a suggestion, I provided the next change: on image preview change we send a request with the image id and get an image asset object back. It will have all related information with which we can update image presentation. 'is_downloaded' as well :) By updating the record we can dynamically affect its representation.

The problem I faced: I didn't find a good solution for the latest case - open the grid widget. I do not know at this moment how to fire a needed request on its opening. Other cases (press buttons, switch between series) work fine.

**What's left:**
- solve the modal open issue
- add tests

**Additional changes:**
- refactoring
- add a new class with serialization

#720

### DIFF
--- a/AdobeStockAsset/Model/GetAssetById.php
+++ b/AdobeStockAsset/Model/GetAssetById.php
@@ -21,6 +21,8 @@ use Magento\AdobeStockAssetApi\Api\GetAssetByIdInterface;
  */
 class GetAssetById implements GetAssetByIdInterface
 {
+    private const MEDIA_ID = 'media_id';
+
     /**
      * @var SearchCriteriaBuilder
      */
@@ -56,7 +58,7 @@ class GetAssetById implements GetAssetByIdInterface
      */
     public function execute(int $adobeId): Document
     {
-        $mediaIdFilter = $this->filterBuilder->setField('media_id')
+        $mediaIdFilter = $this->filterBuilder->setField(self::MEDIA_ID)
             ->setValue($adobeId)
             ->create();
         $searchCriteria = $this->searchCriteriaBuilder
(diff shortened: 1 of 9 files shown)


coderimus:
@sivaschenko tests adjustments to have the green status for this PR will be made after the implementation discussion and review.

sivaschenko:
Hi @coderimus thanks for the pull request. This solution looks concerning from performance/load point of view. Wouldn't it be better to track the delete event and update the corresponding field accordingly?

coderimus:
@sivaschenko I see your concern about the getting the image asset data when we start rendering it at the images grid. But I do not know another way to get the asset info independently from other N conditions. With this, we also cover cases when an image deleted by someone else at the same time or another possible action happen.

With listening to the image delete action, we have to: track the image delete action and notify the UI component in the Adobe Stock that it has to change its image presentation. Currently, I do not see a way how we can do this. Do you have any ideas about how to implement this?

sivaschenko:
@coderimus there should definitely be a way to track Delete button click. 'onclick'? Also, considering that there is a plugin for delete operation - data should be up to date after the page (or grid data) reload.

As for the case when the file is removed from the file system, bypassing the UI - I think we could either handle this situation when "locate" functionality is used of utilizing a cron job

coderimus:
@sivaschenko so as the possible solution we can do next:
1. Track the delete button click
1. Wait for the response (I think it implemented with ajax call)
1. Request our service which will identify that the image asset was deleted
1. Tell to the adobe stock image grid that information update
Am I right with this? )

sivaschenko:
@coderimus ideally it would be good to check if we can observe the delete response.

coderimus:
@sivaschenko ok, I will investigate this possibility and try to implement a solution based on our discussion.

coderimus:
Solution for this problem leads to the major changes. Will be done in the next milestone.

PR #729, opened 2019-11-19
MFTF test runs on travis

### LINKED ISSUE #321: Add MFTF tests to Travis build

filmaj:
### Description (*)

Running the MFTF tests  on a travis-hosted magento instance.

This PR fixes #321


filmaj:
BTW I think the static analysis test introduced in #701 needs to be revised in this PR, since it assumes an Adobe Stock Integration installation that uses the `ext/` directory instead of the `app/code/Magento` directory, which this PR introduces. At least, the invocation of it in the `.travis.yml` file likely needs tweaking.

sivaschenko (approved):

hws47a:
Hi @filmaj, I don't understand why have we moved code from composer installation to adding to app/code. Now we have to add dependencies manually and also composer.json files are not tested this way.

filmaj:
@hws47a I do not understand either 😄 there is probably a better way, and my newness to PHP, composer and magento is probably showing.

If you have suggestions, could you send a PR with your proposed way of setting up the project?

I think that what we do to install the project on Travis should reflect our [Installation Guide wiki page](https://github.com/magento/adobe-stock-integration/wiki/Installation-Guide) and/or our [budding devdocs documentation](https://github.com/magento/devdocs/issues/5778).

PR #744, opened 2019-11-20
Ensure we delete any artifacts created during test execution

### LINKED ISSUE #742: MFTF "Preview" Suite Fails due to leftover artifact in AdminAdobeStockSavePreviewTest
### Steps to reproduce (*)

1. Run the 'preview' suite of tests via the command: `vendor/bin/mftf run:group adobe_stock_integration_suite_preview --remove -vvv`

### Expected result (*)

1. All four tests currently making up this suite should pass.

### Actual result (*)

1. Three tests pass, one fails. The `AdminSavesTheImagePreviewWithNewNameTest` fails.
2. You can see this in e.g. [this Travis run](https://travis-ci.org/magento/adobe-stock-integration/jobs/614222307).

### Why Does It Fail?

Having tested this locally, I think it is because the test `AdminAdobeStockSavePreviewTest` executes before `AdminSavesTheImagePreviewWithNewNameTest`. However, both attempt to save _the same image_ from the Stock search results. The issue is that `AdminAdobeStockSavePreviewTest` does not clean up after itself, so when `AdminSavesTheImagePreviewWithNewNameTest` goes to try to save the image preview and click on the "Save Preview" button, that button is not visible - instead the "Locate" button is visible, as expected, since the image is already saved in the Media Gallery.

### Suggested Fix

As part of `AdminAdobeStockSavePreviewTest`'s `after` conditions, remove the saved image.

filmaj:
### Description (*)

The modified test in this pull request saves an image preview but does not delete it as part of its `after` run. This PR ensures the test cleans up after itself.

This PR fixes #742 

I expect this PR to fix the preview-specific MFTF sub-job on the travis build.

1. You can test it by running  the entire group of preview-related tests like so: `vendor/bin/mftf run:group adobe_stock_integration_suite_preview --remove -vvv`
2. Or you can test it by running just the one modified test like so: `vendor/bin/mftf run:test AdminAdobeStockSavePreviewTest --remove --verbose`


filmaj:
Looks like the relevant sub-job for MFTF passed! https://travis-ci.org/magento/adobe-stock-integration/jobs/614705841

I say we merge this in.

sivaschenko (approved):

PR #749, opened 2019-11-21
run license group of mftf tests on travis

filmaj:
The travis env var should reference the Suite name, not the group name. It is confusing 😛


sivaschenko:
@filmaj thanks

filmaj:
This one needs merging in order to see how the license test results look like on Travis 😄

sivaschenko:
Updated the branch and restarted builds

filmaj:
Doesnt seem like it helped with the IMS tests. We will need to debug further to figure that part  out. Possibly it is Adobe IMS reacting to a login from the ASI-Travis IMS account from a 'weird' location like Travis' datacenter? Dunno. Need screenshot support to verify.

filmaj:
One more thought here is to combine these `license` suite tests with the `ims` suite tests since they all rely on Adobe IMS sign-in functionality - which right now these seem to be failing on Travis.

sivaschenko:
Let's merge this in to have correct configuration for running license tests and create a task to resolve the ims issue. @filmaj if you think it's better to merge license and ims suites - I'm ok with that

Issue #675, opened 2019-11-10
Remove original Image from the related images (More from model and More from series)


konarshankar07:
### Preconditions (*)

1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)

1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
Click "Search Adobe Stock" button to open images grid
3. Click on any image to open the image preview

### Expected result (*)

The image shown in the image preview should not be visible in the `More from series` and `More from Model` Tab

### Actual result (*)

The image shown in the image preview is visible in the `More from series` and `More from Model` Tab

sivaschenko:
Removing the original image prevents navigation back to the original image when selecting a related image and navigating through related images in the image preview

After the discussion with PO we decided that it would be better to keep the original image between related images to avoid the problem described above.

PR #752, opened 2019-11-21
Removed original image from the related images

konarshankar07:

1. magento/adobe-stock-integration#675: Remove First Image from the related images (More from model and More from series) 

1. Login to admin panel
4. Click on any image to open the image preview

### Expected result
The image shown in the image preview should not be visible in the More from series and More from Model Tab


sivaschenko (changes requested):
Hi @konarshankar07 thank you for the pull request!

I don't think the frontend implementation should be aware of that manipulation, can you please isolate this on backend only in the GetRelatedImages class.

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -101,10 +101,11 @@ define([
             $.ajax({
                 type: 'GET',
                 url: this.preview().relatedImagesUrl,
+                showLoader: true,
```
sivaschenko: Global loader should not block user from accessing functionality while related images are loading

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -101,10 +101,11 @@ define([
...
                 dataType: 'json',
                 data: {
                     'image_id': record.id,
-                    'limit': this.tabImagesLimit
+                    'limit': this.tabImagesLimit + 1
```
sivaschenko: Please do not modify the limit on the frontend (in javascript). The frontend should not know the details of related images provider logic
```suggestion
                    'limit': this.tabImagesLimit
```

AdobeStockImage/Model/GetRelatedImages.php
```diff
@@ -83,7 +83,9 @@ public function execute(int $imageId, int $limit): array
                     ->setPageSize($limit)
```
sivaschenko:
```suggestion
                    ->setPageSize($limit + 1)
```

AdobeStockImage/Model/GetRelatedImages.php
```diff
@@ -116,8 +120,11 @@ private function serializeRelatedImages(array $images): array
                     }
                     $itemData[$attribute->getAttributeCode()] = $attribute->getValue();
                 }
-                $data[] = $itemData;
+                if ($image->getId() !== $imageId) {
+                    $data[] = $itemData;
+                }
             }
+            $data = (count($data) > $limit-1 ? array_pop($data) : $data);
```
sivaschenko, line 127:
```suggestion
            $data = array_slice($data, 0, $limit);
```

sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request! Please see my comment below

AdobeStockImage/Model/GetRelatedImages.php
konarshankar07: @sivaschenko .. I'm slightly disagree with this suggested change because it will remove the first image regardless of the preview image is available in the related images or not and user/merchant might miss the first image. Correct me if I'm wrong
Thanks

sivaschenko: @konarshankar07 Thanks for the concern! I think if the original image was removed before this line of code - nothing will be sliced as the array length will be equal to limit. Correct me if I'm wrong

sivaschenko:
This PR is pending discussion with PO

AdobeStockImage/Model/GetRelatedImages.php
konarshankar07: Yes, But as we have incremented the limit to get an extra image which we need to handle if preview image is not available in the related image so the length of the array will be greater than the limit is defined in the javascript and it will display an extra image in the related image section.

sivaschenko:
Hi @konarshankar07 , unfortunately after testing the functionality we found a potential UX problem: 

 - Removing the original image prevents navigation back to the original image when selecting a related image and navigating through related images in the image preview

After the discussion with PO we decided that it would be better to keep the original image between related images to avoid the problem described above.

PR #762, opened 2019-11-22
Fixing "See More From Model/Series" tests

filmaj:
### Description (*)

These changes:

- make the tests  more consistent/similar to each other
- ensuring to scroll to the correct positions within the image preview section when interacting with UI elements within that section
- adding appropriate AJAX wait calls in the tests in order to wait for loading the related images (and thus the same series/model UI elements)

These changes should pass the 'see more' suite of MFTF tests  on Travis 🤞


filmaj:
Green MFTF suite in this PR too 😄 https://travis-ci.org/magento/adobe-stock-integration/jobs/615646443?utm_medium=notification&utm_source=github_status

sivaschenko (approved):

PR #763, opened 2019-11-22
fixing up using suite vs group name for running new license suite

filmaj:
Keeping the naming and use of suite vs group names consistent. This should also stop kicking off a PHP 7.2 MFTF sub-build for the new license suite of tests.


filmaj:
LOL, nm, this is a dupe of #749

PR #772, opened 2019-11-24
#720 : Save a licensed image, then delete, then open grid, 'locate' button is shown on preview

konarshankar07:
This PR will fix the button issue if licensed image is deleted from the Media gallery then Open in media gallery will hide and Save button will be visible after reopen the Adobe Stock Modal

1. magento/adobe-stock-integration#720: Save a licensed image, then delete, then open grid, 'locate' button is shown on preview

1. Expand the preview for a licensed image
2. Save the licensed image to the media gallery
3. Select the saved image and delete it from the media gallery
4. Open the adobe stock grid once more, the preview for the previously-saved image should already be open

### Expected Result

`Open in media gallery` button  will be hidden and `save` button will be visible


konarshankar07:
Hello @sivaschenko ,
I think that we don't need `is_licensed_locally` cause it increases the overhead and instead of `is_licensed_locally` we can use the `is_licensed` only from Adobe Stock API. Let me know your thoughts
Thanks

konarshankar07:
I also found other issues like if user closes the image preview and delete the image from media gallery then button is not changing and if image preview is opened and user logged out from Adobe IMS then also button is not changing. Should I fix these issues in this PR or need to create a new one?

sivaschenko:
`is_licensed_locally` and `is_licensed` are both needed for example to identify that user can download licensed image (display "Save" button) - the condition in this case is `!is_licensed_locally && is_licensed_locally`

sivaschenko:
@konarshankar07 good catch, it would be good to check all the places where we are updating the grid content, but not updating image preview (`displayedImage`) please create a separate PR for that :+1:

konarshankar07:
Thanks for the clarification for `is_licensed_locally` :+1:

sivaschenko (dismissed):

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/actions.js
```diff
@@ -67,7 +64,7 @@ define([
                 return;
             }
 
-            this.rows().forEach(function (record) {
+            this.source().data.items.forEach(function (record) {
                 if (record.id === displayedRecord.id) {
                     updatedDisplayedRecord = record;
```
konarshankar07, line 57: Can we add the `break;` here? As we got the displayedRecord then I don't think we need to run the whole loop? Just suggestion.

sivaschenko:
@magento run Static Tests

Issue #736, opened 2019-11-20
Review and update phpdocs for classes and interfaces


sivaschenko:
Review and update phpdocs for classes and interfaces. The phpdoc for interface/class should clearly describe purpose, use case and behaviour of class/service.

PR #776, opened 2019-11-25
Update the phpdoc section for interfaces and classes

coderimus:
### Description (*)
Update the phpdoc section for classes and interfaces.

I provide the next pattern: 
1. An interface declares the data object or service functionality. Every interface description starts with the declare word, and then the description is started.
2. Implementation phpdoc description begins with the Representation word because of implementation represents declared functionality

#736 

N/A

### DIFF
--- a/AdobeIms/Model/Config.php
+++ b/AdobeIms/Model/Config.php
@@ -14,7 +14,7 @@ use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\UrlInterface;
 
 /**
- * Class Config
+ * Represent the Adobe stock config model responsible for retrieving config settings for Adobe Ims
  */
 class Config implements ConfigInterface
 {
--- a/AdobeIms/Model/Config.php
+++ b/AdobeIms/Model/Config.php
@@ -14,7 +14,7 @@ use Magento\Framework\App\Config\ScopeConfigInterface;
 use Magento\Framework\UrlInterface;
 
 /**
- * Represent the Adobe stock config model responsible for retrieving config settings for Adobe Ims
+ * Represent the Adobe IMS config model responsible for retrieving config settings for Adobe Ims
  */
 class Config implements ConfigInterface
 {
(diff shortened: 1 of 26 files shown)


coderimus:
@sivaschenko I provided update only for the IMS module. Please, review it and if it is ok, I will update the next module in the scope of this PR. I think if I update module per commit, this can make check process more straightforward: you check one module - approve - I provide one more portion in a new commit for this PR. What do you think?

AdobeIms/Model/Config.php
```diff
@@ -14,7 +14,7 @@
 use Magento\Framework\UrlInterface;
 
 /**
- * Class Config
+ * Represent the Adobe stock config model responsible for retrieving config settings for Adobe Ims
```
sivaschenko: Adobe Stock should not be mentioned in AdobeIMS module

AdobeImsApi/Api/ConfigInterface.php
```diff
@@ -9,7 +9,8 @@
 namespace Magento\AdobeImsApi\Api;
 
 /**
- * Class Config
+ * Declare the the Adobe stock integration config which is responsible for retrieving config
```
sivaschenko: Adobe Stock should not be mentioned in AdobeIMS module. The keys here are Adobe Creative SDK integration keys

sivaschenko (changes requested):
Hi @coderimus thanks for the pull request, please see my review comments

coderimus:
@sivaschenko I updated requested docs with removing the stock word. If the rest part is ok - please, write this and I will continue with the next module phpdoc update

Also, I see that the IMS module has a tight relation with the AdobeStockAdminUI module through the config class. I think that if the IMS module will play the role of connection manager for the Adobe services we should exclude such hardcoded part as a config class. It can be moved to the separate module - AdobeStockIms. This module will have its config class `AdobeStockConfig` which will implement the `AdobeStockImsApi\Api\AdobeStockConfigInterface` With such approach we can exclude adobe stock part from the IMS module and make it more independent. What do you think about this?

sivaschenko (approved):

PR #778, opened 2019-11-25
fix MFTF tests that previously relied on "Locate" button

filmaj:
since button was renamed, update tests to reflect button label. also add a few annotations to a couple of MFTF tests (use case links, hiptest links)

This should once again fix the `preview` suite of MFTF test runs on Travis


sivaschenko (approved):

filmaj:
Travis MFTF suites fixed by this PR: `preview` and `keywords`

PR #780, opened 2019-11-25
Update configuration fields MFTF tests

### LINKED ISSUE #779: [MFTF] AdobeStockConfigTest is failing since required fields change
There was a recent change where the attributes of the ASI configuration form where changed, in particular, if I recall correctly, the `required` attribute was removed on the fields.

On the configuration form, I can now:

- leave the fields empty and successfully save/submit the form
- enter garbage credentials and successfully save/submit the form
- if I leave the API Key field empty and click the Test Connection button, I see a special message "Please fill the "API Key (Client ID)" field for a connection test"

If the above are correct states and functionality for the config form, then the AdobeStockConfigTest needs to be updated to handle the new states. In particular we should review whether the `AdobeStockConfigRequiredFieldsActionGroup` checks that entering empty credentials yields an error message of "This is a required field"

Final note: the ConnectionTest actiongroup should introduce a waitForAjaxLoad after clicking the test connection button.

Please advise @sivaschenko.

filmaj:
### Description (*)

api and private keys are no longer required fields; so update the test

1. This fixes #779 

1. `vendor/bin/mftf run:test AdminAdobeStockConfigTest --remove --verbose`


filmaj:
The `configuration` suite of MFTF tests passed via this PR, so I think it is good to merge.

sivaschenko (approved):

PR #788, opened 2019-11-26
Refactoring: adjust files service class

coderimus:
### Description (*)
@sivaschenko the Files class is great! I just made some adjustments while researching its functionality. Added empty ids check, crul response status check before any other step and minor style changes.

N/A

N/A

### DIFF
--- a/AdobeStockClient/Model/Client/Files.php
+++ b/AdobeStockClient/Model/Client/Files.php
@@ -11,17 +11,31 @@ namespace Magento\AdobeStockClient\Model\Client;
 use Magento\AdobeImsApi\Api\ConfigInterface as ImsConfig;
 use Magento\AdobeImsApi\Api\GetAccessTokenInterface;
 use Magento\AdobeStockClientApi\Api\ConfigInterface as ClientConfig;
+use Magento\Framework\Exception\IntegrationException;
 use Magento\Framework\HTTP\Client\CurlFactory;
 use Magento\Framework\Locale\ResolverInterface as LocaleResolver;
 use Magento\Framework\Serialize\Serializer\Json;
 use Magento\AdobeStockClientApi\Api\Client\FilesInterface;
-use Magento\Framework\Exception\IntegrationException;
+use Magento\Framework\Webapi\Exception as WebapiException;
 
 /**
  * Command for retrieving files information from Adobe Stock API
  */
 class Files implements FilesInterface
 {
+    /**
+     * Successful curl result code.
+     */
+    private const CURL_STATUS_OK = 0;
+
+    private const FILES = 'files';
+
+    private const QUERY_PARAM_IDS = 'ids';
+
+    private const QUERY_PARAM_LOCALE = 'locale';
+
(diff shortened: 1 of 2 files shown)


AdobeStockClient/Model/Client/Files.php
```diff
@@ -11,17 +11,31 @@
...
 
 /**
  * Command for retrieving files information from Adobe Stock API
  */
 class Files implements FilesInterface
 {
+    /**
+     * Successful curle result code.
+     */
+    private const CURLE_OK = 0;
```
sivaschenko:
```suggestion
    private const CURL_STATUS_OK = 0;
```

AdobeStockClient/Model/Client/Files.php
```diff
@@ -81,19 +97,27 @@ public function __construct(
...
+        }
+
         $locale = $locale ?? $this->localeResolver->getLocale();
 
         $curl = $this->curlFactory->create();
         $curl->setHeaders($this->getHeaders());
         $curl->get($this->getUrl($ids, $columns, $locale));
 
-        $response = $this->json->unserialize($curl->getBody());
+        if (self::CURLE_OK !== $curl->getStatus()) {
```
sivaschenko:
```suggestion
        if (self::CURL_STATUS_OK !== $curl->getStatus()) {
```

AdobeStockClient/Model/Client/Files.php
```diff
@@ -81,19 +97,27 @@ public function __construct(
      */
     public function execute(array $ids, array $columns, string $locale = null): array
     {
+        if (empty($ids)) {
+            throw new IntegrationException(__('Files ids can not be empty.'));
```
sivaschenko, line 101: I'd return an empty array instead of throwing an exception

AdobeStockClient/Model/Client/Files.php
```diff
@@ -81,19 +97,27 @@ public function __construct(
...
         $curl->setHeaders($this->getHeaders());
         $curl->get($this->getUrl($ids, $columns, $locale));
 
-        $response = $this->json->unserialize($curl->getBody());
+        if (self::CURLE_OK !== $curl->getStatus()) {
+            throw new WebapiException(__('An error occurred during retrieve files information.'));
+        }
 
-        if (!isset($response['files'])) {
+        if (!isset($response[self::FILES])) {
```
sivaschenko, line 115: Shouldn't this check be done after unserialization?

sivaschenko (changes requested):
Thanks for the great improvements @coderimus ! Please take a look at my comments

Also, I think it will make sense to merge that to develop. Can you please base your branch on `develop` branch and change the target

AdobeStockClient/Model/Client/Files.php
coderimus: I provide these changes because we have an exception if 'files' are empty in a result of curl query.  Also, I think that the curl request to the Adobe Stock service with empty ids request parameter will return empty files results in exception.

PR #789, opened 2019-11-27
REFACTORING: adjust phpdoc section

coderimus:
### Description (*)
Adjust phpdoc section of Adobe Stock integration modules: `AdobeStockAsset` and `AdobeStockAsset` with API related modules as well.

### Fixed Issues (if relevant)
N/A

### Manual testing scenarios (*)
N/A

PR #794, opened 2019-11-27
FEATURE: migrate check is integration activated to the proper module

coderimus:
### Description (*)

This change was discussed at the https://github.com/magento/adobe-stock-integration/pull/789 PR. The idea is to exclude service for the integration identification from the Asset module to the more suitable module.

N/A

N/A


coderimus:
@sivaschenko we discussed that we need to move this to the `AdobeStockImageAdminUI`. I migrated it to the `AdobeStockAdminUiI`. Not sure that this is a correct level. Please, check this and if the provided level is not right, I will quickly migrate this to the correct one.

coderimus:
@sivaschenko this PR is ready to be reviewed. The UNIT test failure is a mistic thing for me. The test report doesn't show any critical error or else. If you will find something, please, report and I will fix. I attach the view I have at the Unit test report page:
image

AdobeStockAdminUi/Model/IsAdobeStockIntegrationEnabled.php
```diff
@@ -37,11 +36,11 @@ public function __construct(ScopeConfigInterface $scopeConfig)
     }
 
     /**
-     * Is integration enabled
+     * Check is the Adobe Stock integration enabled or not
      *
      * @return bool
      */
-    public function isEnabled(): bool
+    public function checkStatus(): bool
```
sivaschenko, line 43: To be honest I'd preserve the naming of the class and method. `checkStatus` seems confusing to me (status is not usually boolean).

AdobeStockAdminUi/Test/Unit/Model/IsAdobeStockIntegrationEnabledTest.php
```diff
@@ -0,0 +1,54 @@
...
+use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\Framework\TestFramework\Unit\Helper\ObjectManager;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Test functionality which is used for identification is the Adobe Stock integration enabled or not
+ */
+class IsAdobeStockIntegrationEnabledTest extends TestCase
+{
+    private const XML_PATH_ENABLED = 'adobe_stock/integration/enabled';
```
sivaschenko, line 20: This value is used basically to identify to add or not "Search Adobe Stock" button to media gallery. The best place for it is `AdobeStockImageAdminUi` module

sivaschenko (changes requested):
Hi @coderimus thanks for the pull request! Can you please base/target this PR to develop branch?

I don't think we will be able to remove API interface after initial release

AdobeStockAdminUi/Model/IsAdobeStockIntegrationEnabled.php
coderimus: Cool notice about the status! I will keep this for future implementation. You are right.

@sivaschenko is it applicable for you if I will leave the class name as `IsAdobeStockIntegrationEnabled` and method name `execute`?

AdobeStockAdminUi/Test/Unit/Model/IsAdobeStockIntegrationEnabledTest.php
coderimus: Ok, got it. Will change.

PR #806, opened 2019-11-28
Unit test for Magento\AdobeStockAsset\Model\Config.php

konarshankar07:
Unit test for `Magento\AdobeStockAsset\Model\Config.php`

### DIFF
--- a/AdobeStockAsset/Test/Unit/Model/ConfigTest.php
+++ b/AdobeStockAsset/Test/Unit/Model/ConfigTest.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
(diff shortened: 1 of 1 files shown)


sivaschenko (changes requested):
Hi @konarshankar07 thanks for the unit test! Can you please check the static test fails

coderimus:
@konarshankar07 thank you for you PR, but I think that it is not relevant at this moment. This Config class were excluded from the `AdobeStock` module and located to the new area. This decision was made previously and unfortunately was not declared as an issue. Please, see that PR:  #789 where the discussion started and first implementation made #794 . After that this PR made as a final one: https://github.com/magento/adobe-stock-integration/pull/808

Thank you for your work. I am really happy to cooperate with you!

konarshankar07:
Hello @coderimus ,
No worries, I will close this PR :)
Thanks

Issue #801, opened 2019-11-28
Set Acl only for Adobe Integration (Media Gallery resource)  brokes Admin Panel


engcom-Delta:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)

- Login to admin panel
- Set **Advanced->Admin->Security->Add Secret Key to URLs=Yes**
- Go to System->User Roles
- Create new role:
- In Role Resources check only Media Gallery
![image](https://user-images.githubusercontent.com/51681379/69806266-2e4ce700-11eb-11ea-8a24-6fda6caac63f.png)
- Create new admin user with previously created role
- Sign in by new admin user

### Expected result (*)
New admin user is logged into admin panel

### Actual result (*)
User is redirected to frontend page with status code 404. Admin panel is not available for other admins with full permissions
![image](https://user-images.githubusercontent.com/51681379/69808832-21cb8d00-11f1-11ea-85cd-bcf9be32aca3.png)

sivaschenko:
@engcom-Delta can you please check if the issue is the same on vanilla Magento without Adobe Stock Integration (select only Media Gallery resource)

engcom-Delta:
@sivaschenko Results on 2.3-develop:

**1. Set ACL for Media Gallery**
![#801ASIMediaACL](https://user-images.githubusercontent.com/51681379/69956964-361dcb80-150a-11ea-9332-20449bd61701.png)
**Result:**
![#801ASIMedia](https://user-images.githubusercontent.com/51681379/69956980-45047e00-150a-11ea-882d-9a8f677ed53b.png)

**2. Set ACL for Blocks**
![#801ASIBlockACL](https://user-images.githubusercontent.com/51681379/69956996-50f04000-150a-11ea-8516-e88685b7f288.png)
**Result:**
![#801ASIBlocks](https://user-images.githubusercontent.com/51681379/69957009-5b123e80-150a-11ea-9ebe-1aaad3f16851.png)

**3. Set ACL for Widgets**
![#801ASIWidgetsACL](https://user-images.githubusercontent.com/51681379/69957040-69605a80-150a-11ea-948c-bca7150ad226.png)
**Result:**
![#801ASIWidget](https://user-images.githubusercontent.com/51681379/69957054-71b89580-150a-11ea-8675-0b1701088302.png)

sivaschenko:
[...]

PR #810, opened 2019-11-29
Set Acl  for Adobe Integration Depend On Pages

Nazar65:

1. magento/adobe-stock-integration#801:Set Acl only for Adobe Integration (Media Gallery resource) brokes Admin Panel 
2. ...

Same as in Issue description


sivaschenko:
Our acl resource path should not be changed - the issue should be fixed on Magento side

Issue #807, opened 2019-11-29
The "Undiscovered" filter is presented in Sort by drop-down list, not as a separate filter 


chalov-anton:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
### Steps to reproduce (*)
1. From Magento Admin go to Content - Pages, click **Add New Page**
2. Expand **Content** Click **Show / Hide Editor**, select **Insert Image...**
3. Click **Search Adobe Stock**
4. Click **Sort by** dropdown list and observe 
### Expected result (*)
The _Undiscovered_ is a separate filter if check the https://stock.adobe.com/ web page. _Undiscovered_ should be in Filters not in **Sort by**
![undiscovered_filter_AdobeStock](https://user-images.githubusercontent.com/45624059/69853777-61958180-1290-11ea-9984-42ae3aac53dc.png)

### Actual result (*)
The _Undiscovered_ item present in **Sort by** dropdown list
![undiscovered sort_mage](https://user-images.githubusercontent.com/45624059/69853563-e9c75700-128f-11ea-885f-ac410a31e316.png)

sivaschenko:
Adobe Stock API does not support `Undiscovered` as a filter - it is sorting. Adobe Stock website is using a different service

PR #815, opened 2019-11-29
The "Undiscovered" filter is presented in Sort by drop-down list, not as a separate filter

Nazar65:

1. magento/adobe-stock-integration#807: The "Undiscovered" filter is presented in Sort by drop-down list, not as a separate filter

1. ...
2. ...


sivaschenko:
Hi @Nazar65 thanks for the pull request. Looks like the undiscovered filter is conflicting with sorting: sorting does not work if the undiscovered filter is applied.

Looks like we cannot use undiscovered as filter considering the actual Adobe Stock API documentation: https://www.adobe.io/apis/creativecloud/stock/docs.html#!adobe/stock-api-docs/master/docs/api/11-search-reference.md

@chfabbro do you think there is a way to use both `undiscovered` filter and sorting by i.e. `relevance` in the request to Adobe Stock API?

chfabbro:
> @chfabbro do you think there is a way to use both `undiscovered` filter and sorting by i.e. `relevance` in the request to Adobe Stock API?

@sivaschenko They are mutually exclusive, i.e., "either/or". You can _either_ sort by "relevance," or sort by "undiscovered," but not both. Per the docs you quoted:
```
search_parameters[order]
Sort order in which to return found assets. Default is "relevance". String.
Valid strings and their meanings:
  * relevance
How closely it matches your search request, closest matches first.
  * creation
Creation date in descending order (newest first).
  * popularity
In descending order by the number of views by all users.
  * nb_downloads
In descending order by the number of downloads by all users since the asset was added to Adobe Stock.
  * undiscovered
Starting with assets that have not commonly been viewed or downloaded.

sivaschenko:
Yes @chfabbro , we were just concerned that "undiscovered" is a separate filter on stock.adobe.com
image

chfabbro:
@sivaschenko Yes, that UI is new and unfortunately has different functionality from the Stock API. The website does not use the same API, and instead calls directly to the Adobe Search service. Net, changes would need to be made to the Stock API to imitate this functionality, but it is not on the roadmap at this time.

sivaschenko:
Ok, thanks @chfabbro

sivaschenko:
Considering that API does not support this case I am closing both the PR and the issue.

PR #816, opened 2019-11-29
Fix Save Licensed Image MFTF test

filmaj:
Hat tip to @sivaschenko for pointing out that we load image licensing state in the background asynchronously after an IMS sign in, which was causing one of the MFTF tests to fail.

This PR adds a wait in the user profile assertion action group, ensuring asynchronous requests complete before asserting that the IMS profile image is visible. This will also wait for licensing information to load in the background.

Expectation is that the `AdminAdobeStockSavedLicensedImageLocateTest` test passes in this PR.


filmaj:
Woot, down to only a single failing MFTF test!!!

Issue #774, opened 2019-11-24
[1.1-develop] Introduce MediaGalleryUiApi module and ConfigInterface


sivaschenko:
MediaGalleryUiApi module should be added to the repository including all the general necessary module files.

- `.modman` and metapackage `composer.json` should be updated
- `\Magento\MediaGalleryUiApi\Model\ConfigInterface` containing a single `isEnabled` method declaration should be added to the module
 -`\Magento\MediaGalleryUi\Model\Config` should implement `\Magento\MediaGalleryUiApi\Model\ConfigInterface` and be a declared preference for this interface in DI

Please open a pull request for this issue to **1.1-develop** branch (not default currently)

This task can be implemented after https://github.com/magento/adobe-stock-integration/pull/773 will be merged

konarshankar07:
Hello @sivaschenko ,
As mentioned in the task description
> \Magento\MediaGalleryUiApi\Model\ConfigInterface containing a single isEnabled method declaration should be added to the module

I think we need to create `ConfigInterface` in `\Magento\MediaGalleryUiApi\Api\ConfigInterface` instead of `\Magento\MediaGalleryUiApi\Model\ConfigInterface`. Correct me If I'm wrong

Thanks

PR #817, opened 2019-11-29
#774 : MediaGalleryUiApi module added

konarshankar07:
### Description (*)

This PR will introduce MediaGalleryUiApi module and also added service contract for `\Magento\MediaGalleryUi\Model\Config`

1. magento/adobe-stock-integration#774: [1.1-develop] Introduce MediaGalleryUiApi module and ConfigInterface


MediaGalleryUi/Model/Config.php
```diff
@@ -9,11 +9,12 @@
 namespace Magento\MediaGalleryUi\Model;
 
 use Magento\Framework\App\Config\ScopeConfigInterface;
+use Magento\MediaGalleryUiApi\Model\ConfigInterface;
```
konarshankar07: @sivaschenko .. I think we should move `ConfigInterface` from `Model` folder to `Api` folder

MediaGalleryUiApi/composer.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "magento/module-media-gallery-ui-api",
+    "description": "Magento module responsible for the media gallery UI implementation API",
+    "require": {
+        "php": "~7.1.3||~7.2.0||~7.3.0",
+        "magento/framework": "*"
```
konarshankar07, line 6: Do I need to add `MediaGalleryUi` module in require?

sivaschenko: nope, it should be the other way around - MediaGalleryUi is supposed to implement interfaces of MediaGalleryUiApi - So MediaGalleryUiApi should be required for MediaGalleryUi module

MediaGalleryUi/Model/Config.php
sivaschenko: Yes

sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request! Please correct the `ConfigInterface` namespace in the reference and update the MediaGalleryUi composer.json with MediaGalleryUiApi dependency

sivaschenko (approved):

Issue #677, opened 2019-11-10
See more from series/model tabs to hide if result is empty


konarshankar07:
### Preconditions (*)

1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)

1. Login to admin panel
2. Open Magento Media Gallery (i.e. go to Cms -> Pages, edit the page and insert image)
3. Click "Search Adobe Stock" button to open images grid
4. Enter text `81214721` in search by keywords
5. Click on the image to open the image preview

### Expected result (*)

![image](https://user-images.githubusercontent.com/39480008/68542496-3a573d00-03d3-11ea-9277-eed330a7a320.png)

### Actual result (*)

![image](https://user-images.githubusercontent.com/39480008/68542490-23184f80-03d3-11ea-97d6-2888ab378b09.png)

rogyar:
Hi folks. Currently, if there no items in "More from this series", we show the same image that is opened in the preview

![image](https://user-images.githubusercontent.com/2148959/69634884-7cc77d80-1053-11ea-887f-350ef982d13f.png)

Moreover, the very first image from "More from this series" is always the same image that is opened in the preview.

![image](https://user-images.githubusercontent.com/2148959/69635076-e21b6e80-1053-11ea-8262-30f3ccf36c0d.png)

So, technically it may not be empty. Should we change this behavior and hide the tab instead if there's only 1 image in the "More from this series" section? 

Thank you.

cc @sivaschenko.

sivaschenko:
@rogyar you are right, however, it could be empty after merge of https://github.com/magento/adobe-stock-integration/pull/752

rogyar:
Hi @sivaschenko. Thank you for the reference, that makes sense. I will proceed to work on that.

rogyar:
Hi folks. I've investigated the current preview behavior and must admit that this change is not as straightforward as it seems to be. 

The problem is, once the preview is expanded we don't have information about a number of related images until a separate AJAX request for the related images is completed. So in the current implementation, we create empty tabs, wait for the related images request result, fill the tabs with the corresponding images. 
So, in the very beginning, we cannot hide a particular tab since we have no information about its contents. As a solution, we may hide all tabs until we get a response with the related images. However, the related images request takes some time and the visitor might not know that the functionality of the related images exists if the visitor goes through previews faster than he gets the related images. 

[...]

PR #818, opened 2019-12-01
Hide tabs with related images if result is empty

rogyar:
### Description (*)
This PR adjusts the related images tabs behavior. So if there are no related images from the same series or model, the corresponding tab is being hidden.

1. #677 : See more from series/model tabs to hide if result is empty

1. Login to admin panel
4. Enter text `81214721` in search by keywords
5. Click on the image to open the image preview

The "More from this series" tab should be hidden since there are no images from the same series.

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
@@ -111,9 +111,37 @@ define([
 
                 relatedImages.series[record.id] = data.result['same_series'];
                 relatedImages.model[record.id] = data.result['same_model'];
+
                 this.relatedImages(relatedImages);
                 this.preview().updateHeight();
+
+                /* Switch to the model tab if the series tab is hidden */
+                if (relatedImages.series[record.id].length === 0) {
+                    $('#adobe-stock-tabs').data().mageTabs.select(1);
+                }
             }.bind(this));
+        },
+
+        /**
+         * Returns true if the series tab should be show, false otherwise
+         *
+         * @param {Object} record
+         * @returns boolean
+         */
+        showSeriesTab: function (record) {
+            return typeof this.relatedImages().series[record.id] === 'undefined' ||
+                this.relatedImages().series[record.id].length !== 0;
+        },
+
+        /**
+         * Returns true if the model tab should be show, false otherwise
+         *
+         * @param {Object} record
+         * @returns boolean
+         */
+        showModelTab: function (record) {
+            return typeof this.relatedImages().model[record.id] === 'undefined' ||
+                this.relatedImages().model[record.id].length !== 0;
         },
 
         /**
--- a/AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/preview/related.html
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/preview/related.html
@@ -5,23 +5,22 @@
  */
 -->
 <div if="isVisible($row())"
-    class="related-container">
+     class="related-container">
     <div id="adobe-stock-tabs"
          class="adobe-stock-tabs"
          data-bind="mageInit: {
                     'mage/backend/tabs': {
-                    active: 'series_content_' + $row().id,
                     destination: '#adobe-stock-tabs-content',
                     shadowTabs: [],
                 }
             }">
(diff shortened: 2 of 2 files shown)


AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/preview/related.html
```diff
@@ -31,6 +31,7 @@
 
     <div id="adobe-stock-tabs-content"
          class="adobe-stock-related-images-tab-content">
+        <!-- ko if: typeof relatedImages().series !== 'undefined' && relatedImages().series[$row().id] !== false -->
```
sivaschenko: this binding can be added to a div

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -109,8 +109,10 @@ define([
             }).done(function (data) {
                 var relatedImages = this.relatedImages();
 
-                relatedImages.series[record.id] = data.result['same_series'];
-                relatedImages.model[record.id] = data.result['same_model'];
+                relatedImages.series[record.id] = data.result['same_series'].length > 0 ?
```
sivaschenko: I think it would be better to introduce `isTabVisible` method and merge the logic introduced here and in the template as part of that method: `function (record, tabName) { return typeof this.relatedImages()[tabName][record.id] !== 'undefined' && data.result['same_series'].length > 0 }`

AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/preview/related.html
```diff
@@ -16,12 +16,12 @@
                 }
             }">
         <ul class="tabs-horiz">
-            <li>
+            <li data-bind="visible: (typeof relatedImages().series[$row().id] === 'undefined' || relatedImages().series[$row().id] !== false)">
```
sivaschenko: What do you think about using `if` attribute binding instead of `data-bind="visible` - shorter and shouldn't render the element at all

sivaschenko (changes requested):
Thanks for the PR @rogyar ! See my comments below

rogyar:
Hi @sivaschenko. Thank you for the review, appreciate that. 

I have reworked the concept taking your recommendations into account as well. Please, take a look when you have some time. 

Thank you.

AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/preview/related.html
```diff
@@ -5,23 +5,22 @@
...
          data-bind="mageInit: {
                     'mage/backend/tabs': {
-                    active: 'series_content_' + $row().id,
                     destination: '#adobe-stock-tabs-content',
                     shadowTabs: [],
                 }
             }">
         <ul class="tabs-horiz">
-            <li>
+            <li data-bind="if: showSeriesTab($row())">
```
sivaschenko: I believe this syntax should work
```suggestion
            <li if="showSeriesTab($row())">
```

AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -17,6 +17,8 @@ define([
             tabImagesLimit: 4,
             serieFilterValue: '',
             modelFilterValue: '',
+            isSeriesTabVisible: [],
+            isModelTabVisible: [],
```
sivaschenko: Do you think this intermidiate fields are necessary? Why wouldn't `showSeriesTab` method work directly with `this.relatedImages().series[record.id].length !== 0 || typeof this.relatedImages().series[record.id] === 'undefined'`

sivaschenko (changes requested):
@rogyar Thanks for the updates! Please take a look at my comments below

AdobeStockImageAdminUi/view/adminhtml/web/template/grid/column/preview/related.html
rogyar: That makes sense. I've been playing around the way that will make the behavior of the tabs more obvious. However, I've ended up with a lot of redundant code.  

Now it should be better. Thanks for the suggestion.

sivaschenko (approved):

Issue #821, opened 2019-12-03
Bookmarks controls are shifter to the right out of screen


chalov-anton:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
### Steps to reproduce (*)
1. From Magento Admin go to Content - Pages, click **Add New Page**
2. Expand **Content** Click **Show / Hide Editor**, select **Insert Image...**
3. Click **Search Adobe Stock**
4. Click **Filters** and **Apply** any filter, _Price - Standard_ for example (this step is done to enable Save View As button)
5. Click **Default View** and select **Save View As...**
6. Type in any name and click Enter key
7. Click on View dropdown again and observe
![filter_remove_edit](https://user-images.githubusercontent.com/45624059/70053839-fa086a80-15de-11ea-97ad-54ba17a13d3b.gif)
 
### Expected result (*)
The buttons **Save all Changes**, **Edit** and **Delete** should be visible 
See, for example, the View from _Catalog - Products_
![Peek 2019-12-03 14-46](https://user-images.githubusercontent.com/45624059/70053960-40f66000-15df-11ea-9a87-59bc11c4df50.gif)

### Actual result (*)
The **Save all Changes**, **Edit** and **Delete** buttons do not fit into the screen
![remove](https://user-images.githubusercontent.com/45624059/70053813-e3621380-15de-11ea-98d9-e233fd59712f.png)

sivaschenko:
The fix for this issue should be delivered to develop branch

filmaj:
This, possibly, may be good to cover with an MFTF test. Thoughts?

sivaschenko:
@filmaj good idea. @chalov-anton is there a test case(s) that are covering the functionality of buttons that were not available on the Adobe Stock grid?

chalov-anton:
> @filmaj good idea. @chalov-anton is there a test case(s) that are covering the functionality of buttons that were not available on the Adobe Stock grid?
@sivaschenko We have created a Scenario in SucumberStudio which covers this

chalov-anton:
@sivaschenko We have created a Scenario in CucumberStudio which covers this, and already used it in the previous Regression Cycle

sivaschenko:
@chalov-anton could you please create an "[MFTF] Cover Adobe Stock grid bookmarks functionality" issue with the steps from the scenario in CucumberStudio. Thank you!

PR #835, opened 2019-12-03
Shifting view controls to the left by applying a right padding

filmaj:
...so that saved views and related buttons slide back into the viewport.

### Description (*)

The view controls are shifted to the far to the left, so if you want to edit or delete saved views, you are going to have a bad time right now.

The fix is applying a right padding to the header controls. The size of this padding was determined by looking at the `min-width` CSS property for the saved view menu items (see [its definition here](https://github.com/magento/magento2/blob/0eb8677b0b4e35606032e856cc1ef7c80e68829f/app/design/adminhtml/Magento/backend/Magento_Ui/web/css/source/module/data-grid/data-grid-header/_data-grid-action-bookmarks.less#L74)). The right padding needs to be just shy of 50% of the View dropdown item `min-width` for all saved view buttons to be visible.

We should also possibly consider filing a new issue to cover using these buttons in an MFTF test, in order to catch these buttons becoming invisible again in the future.

Screenshots of the new changes in action:

Screen Shot 2019-12-03 at 6 23 28 PM

Screen Shot 2019-12-03 at 6 23 35 PM

1. #821 

1. From Magento Admin go to Content - Pages, click **Add New Page**
2. Expand **Content** Click **Show / Hide Editor**, select **Insert Image...**
3. Click **Search Adobe Stock**
4. Click **Filters** and **Apply** any filter, _Price - Standard_ for example (this step is done to enable Save View As button)
5. Click **Default View** and select **Save View As...**
6. Type in any name and click Enter key
7. Click on View dropdown again
8. You should be able  to see the pencil icon (without this PR, you won't)
9. Even if you click the pencil icon, you should be able to see a Trash can icon after clicking it.

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
@@ -80,7 +80,7 @@
         }
 
         .admin__data-grid-outer-wrap {
-            .admin__data-grid-header {
+            .admin__data-grid-header-row {
                 padding-right: 8rem;
             }
         }
--- a/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
@@ -79,6 +79,12 @@
             }
         }
 
+        .admin__data-grid-outer-wrap {
+            .admin__data-grid-header {
+                padding-right: 8rem;
+            }
+        }
+
         .masonry-image-grid {
             + .admin__data-grid-header {
                 .selectmenu-items {


AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
```diff
@@ -79,6 +79,12 @@
             }
         }
 
+        .admin__data-grid-outer-wrap {
+            .admin__data-grid-header {
```
sivaschenko:
```suggestion
            .admin__data-grid-header-row {
```

sivaschenko (approved):

sivaschenko:
Thanks @filmaj !

Issue #827, opened 2019-12-03
Element '#adobe-stock-tabs-content .three-dots' can be overlapped by "Similar keywords"


TomashKhamlai:
### Preconditions (*)

1. Adobe Stock is configured

### Steps to reproduce (*)

1. On CMS edit page try to add an image by searching
2. In Chrome Dev Tools toggle Device Toolbar and choose Kindle Fire HDX in landscape mode

### Expected result (*)

1. Elements do not overlap

### Actual result (*)

1. Element '#adobe-stock-tabs-content .three-dots' can be overlapped by "Similar keywords"
![image](https://user-images.githubusercontent.com/11827230/70067519-10222500-15f7-11ea-9d7d-2b576f230846.png)

PR #853, opened 2019-12-06
See more link can be overlapped by similar keywords

konarshankar07:
### Description (*)

This PR will fix the overlap issue with see more element in related image section for kindle device

1. magento/adobe-stock-integration#827: Element '#adobe-stock-tabs-content .three-dots' can be overlapped by "Similar keywords"

1. On CMS edit page try to add an image by searching
2. In Chrome Dev Tools toggle Device Toolbar and choose Kindle Fire HDX in landscape mode

### Expected Result
image

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/css/source/_module.less
@@ -200,7 +200,8 @@
                         justify-content: center;
                         position: relative;
                         height: 100px;
-                        width: 150px;
+                        max-width: 130px;
+                        width: 100%;
                         overflow: hidden;
                         margin-right: 10px;
                         vertical-align: middle;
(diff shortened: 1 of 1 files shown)


sivaschenko:
Thanks for the pull request @konarshankar07 ! The PR fixes the issue on iPad, but related images look a bit squashed (not occupying the dedicated space) on bigger screens:

image

I think either relative units or media query should be applied

sivaschenko (changes requested):
Thanks for the pull request @konarshankar07 ! The PR fixes the issue on iPad, but related images look a bit squashed (not occupying the dedicated space) on bigger screens:

image

I think either relative units or media query should be applied

sivaschenko (approved):

engcom-Delta:
:heavy_check_mark: QA passed

Issue #733, opened 2019-11-20
[1.1-develop] ASI Asset module. Use commands and proxy them in repositories


coderimus:
@sivaschenko my suggestion here is to replace delete/getById method in the related repositories with commands. And proxy commands in those methods. Implementation of them will be similar to the Madia Gallery module.

sivaschenko:
Please create PR to 1.1-develop

PR #854, opened 2019-12-08
733: provide commands for the adobe stock asset

coderimus:
### Description (*)
Provide commands for the Adobe Stock asset module. This approach allowed to make related repositories more light-weighted and easy to implement. Makes the module code structure clear and easy to maintain.

After PR approved will cover with tests AssetRepository + others in the Adobe Stock asset module.

#733 

N/A


AdobeStockAsset/Model/AppendAttributes.php
```diff
@@ -7,12 +7,11 @@
 
 namespace Magento\AdobeStockAsset\Model;
 
-use Magento\MediaGalleryApi\Model\Asset\Command\GetByIdInterface;
-use Magento\AdobeStockAsset\Model\ResourceModel\Asset\LoadByIds;
+use Magento\AdobeStockAsset\Model\ResourceModel\Asset\Command\LoadByIds;
 use Magento\Framework\Api\AttributeValueFactory;
-use Magento\Framework\Api\Search\Document;
-use Magento\Framework\Api\Search\SearchResultInterface;
+use Magento\Framework\Api\Search\{Document,SearchResultInterface};
```
diazwatson: @coderimus could you provide some documentation about this way of declare use?
Is this what you mentioned during the call?

Thanks

coderimus: Hello @diazwatson ,

[In the PHP 7.0 the Group Namesapce Declaration was added ](https://wiki.php.net/rfc/group_use_declarations) This approach provides the ability to group imported namespaces a make them more structured. In the PSR-12 which approved to be used (as I know) we have some rules about this feature. You can find information here: [3. Declare Statements, Namespace, and Import Statements](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements)

diazwatson: Thanks :)

AdobeStockAssetApi/Model/Asset/Command/LoadByIdsInterface.php
```diff
@@ -0,0 +1,26 @@
...
+interface LoadByIdsInterface
+{
+    /**
+     * Load Adobe Stock assets by ids
+     *
+     * @param int[] $ids
+     *
+     * @return AssetInterface[]
+     */
+    public function execute(array $ids): array;
```
sivaschenko: @coderimus should we use the bulk approach to all of the Asset commands?

AdobeStockAssetApi/Api/CreatorRepositoryInterface.php
```diff
@@ -19,46 +19,48 @@
 interface CreatorRepositoryInterface
```
sivaschenko: That is an API interface, please do not change the method signatures for it

AdobeStockAssetApi/Api/CategoryRepositoryInterface.php
```diff
@@ -19,46 +19,48 @@
 interface CategoryRepositoryInterface
```
sivaschenko: That is an API interface, please do not change the method signatures for it

sivaschenko (changes requested):
Thanks for the pull request @coderimus , please see my review comments

AdobeStockAssetApi/Model/Asset/Command/LoadByIdsInterface.php
coderimus: @sivaschenko I think no, we should not practise the bulk approach for all Asset related commands. From my point of view, using it everywhere is overhead and, personally, breaks the single responsibility approach. I know that this sounds like a not so good example, but in fact, it is :) When you need to get one item by its id it doesn't seem to be logical to get it with the operation for getting multiple items. We even have different conditions in the SQL (equal and in).

In conclusion: I left the LoadById command and introduced the new one - MultiplyLoadById. We need such functionality at the AppendAttributes class.

coderimus:
@sivaschenko please, check the failed tests. It looks like all of the related to the temporary controller implementation we have for the media gallery. If yes maybe we can suppress them during the development process?

coderimus:
@sivaschenko the majority of the failed tests fixed. There only two left, and I want to ask you for help with them. Maybe they both related to the infrastructure? Can you please check two failed tests?

sivaschenko (approved):

Issue #862, opened 2019-12-12
"Found 2 elements with non-unique id" warning in browser console on Adobe Stock grid


engcom-Delta:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
3. Chrome version(Version 79.0.3945.79)

### Steps to reproduce (*)

- Login to admin panel
- Navigate to Cms Pages
- User clicks Add New Page button
- Expand "Content" section
- User clicks "Insert Image..." button
- Click "Search Adobe Stock" button to open images grid

### Expected result (*)
No warning messages, all id's should be unique

### Actual result (*)
Warning message **"Found 2 elements with non-unique id #adobe_stock_images_listing.adobe_stock_images_listing.listing_top.listing_paging_sizes"** is shown

![Screenshot from 2019-12-12 17-04-53](https://user-images.githubusercontent.com/51681379/70723562-bd84ef00-1d01-11ea-96eb-9d770cadbdf7.png)

drpayyne:
I'm not able to reproduce this error and I don't remember coming across this warning recently either. Is anyone else able to reproduce this?

engcom-Delta:
@drpayyne This console message occurs on latest Chrome version(Version 79.0.3945.79).
For example, Firefox v.71.0 doesn't show this message

drpayyne:
Yes, I see the duplicate IDs. Thanks.

konarshankar07:
Hello guys,
I have investigated this issue and I think this issue also present in the core magento 2. Here is the screenshot of catalog grid console window
![image](https://user-images.githubusercontent.com/39480008/70863743-4eadcd00-1f71-11ea-9cf1-4b7b20b2601a.png)

rogyar:
Hi folks. 

In this particular case, the issue is related to sticky toolbar. Basically, the sticky toolbar duplicates the original toolbar's content but stays hidden. That's why we have two inputs with the same IDs. 
I just wonder if we still need [the sticky toolbar](https://github.com/magento/adobe-stock-integration/blob/60cfa8063cd3aacd7e84e241f1108ea38ecffd13/AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml#L38) functionality in the product listing. 
[...]

PR #875, opened 2019-12-19
Removed sticky toolbar

konarshankar07:
### Description (*)

This PR will fix the duplicated id issue with admin grid toolbar. Currently, we are not using the sticky toolbar so we decided to remove from the grid

1. magento/adobe-stock-integration#862: "Found 2 elements with non-unique id" warning in browser console on Adobe Stock grid

1. Login to admin panel
2. Navigate to Cms Pages
3. User clicks Add New Page button
4. Expand "Content" section
5. User clicks "Insert Image..." button
6. Click "Search Adobe Stock" button to open images grid

### Expected result
No warning messages, all id's should be unique

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
+++ b/AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
@@ -35,7 +35,7 @@
     </dataSource>
     <listingToolbar name="listing_top" template="Magento_AdobeStockImageAdminUi/grid/toolbar">
         <settings>
-            <sticky>true</sticky>
+            <sticky>false</sticky>
         </settings>
         <bookmark name="bookmarks"/>
         <filterSearch name="words"/>
--- a/AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
+++ b/AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
@@ -34,9 +34,6 @@
         </dataProvider>
     </dataSource>
     <listingToolbar name="listing_top" template="Magento_AdobeStockImageAdminUi/grid/toolbar">
-        <settings>
-            <sticky>false</sticky>
-        </settings>
         <bookmark name="bookmarks"/>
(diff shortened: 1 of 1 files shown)


AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
```diff
@@ -35,7 +35,7 @@
     </dataSource>
     <listingToolbar name="listing_top" template="Magento_AdobeStockImageAdminUi/grid/toolbar">
         <settings>
-            <sticky>true</sticky>
+            <sticky>false</sticky>
```
sivaschenko: False appears to be the default value, I believe we can simply remove settings node

sivaschenko (changes requested):
Thanks for the pull request @konarshankar07 ! Please see my comment

sivaschenko (approved):

AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
konarshankar07: @sivaschenko ... you are right. I have removed the settings node from the file.

chalov-anton:
:heavy_check_mark: The PR fixes the issue

AdobeStockImageAdminUi/view/adminhtml/ui_component/adobe_stock_images_listing.xml
mrtuvn: some modules still use this value. Remove settings value seem not fix actually problem. In case we need use true sticky

mrtuvn: Customer, Cms still use sticky in listing grid

sivaschenko: @mrtuvn we don't need a sticky header for Adobe Stock Integration. The root cause for this issue should be addressed in magento2 repository, here is a corresponding issue https://github.com/magento/magento2/issues/26147

Issue #798, opened 2019-11-27
Cover overlay.js with jasmine tests


sivaschenko:
Cover `AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.js` with jasmine tests

konarshankar07:
Hello @sivaschenko ,
I have started working on this ticket and found that jasmine test files are created in `Magento_Root/dev/tests/js/jasmine/tests/` folder but how can we run jasmine test in our repo?
My guess is `ext/magento/stock-integration/AdobeStockImageAdminUi/Test/js/view/adminhtml/web/js/components/grid/column/overlay.test.js`
Let me know your thoughts
Thanks

sivaschenko:
Hi @konarshankar07 , sorry for delay, lets place them in ASI_Root/dev/tests/js/jasmine/tests/ for now

PR #885, opened 2020-01-05
#798 : Cover overlay.js with jasmine tests

konarshankar07:
### Description (*)

This PR will cover the jasmine test for the `ext/magento/stock-integration/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.js` file

1. magento/adobe-stock-integration#798: Cover overlay.js with jasmine tests

1. Follow this [documentation](https://devdocs.magento.com/guides/v2.3/test/js/jasmine.html) to configure jasmine test
2. Run the following command from ASI root directory 
`grunt spec:backend --file="dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js"`


dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                expect(overlayObj.hasOwnProperty('getLabel')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.getLabel;
+                expect(type).toEqual('function');
+            });
+
+            it('Check returned value type if method called with object argument', function () {
+                var row ={
```
Nazar65:
```suggestion
                var row = {
```

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                    id: 123213
+                };
+                expect(typeof overlayObj.getLabel(row)).toEqual('string');
+            });
+
+            it('Check called "this.getLabel" method with object argument', function () {
+                var row ={
```
Nazar65:
```suggestion
                var row = {
```

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                };
+                overlayObj.getLabel = jasmine.createSpy().and.callFake(function () {
+                    return '';
+                });
+                overlayObj.getLabel(row);
+                expect(overlayObj.getLabel).toHaveBeenCalledWith(row);
+            });
+
+            it('Check getLabel and return null if id doesnt found', function () {
+                var row ={
```
Nazar65:
```suggestion
                var row = {
```

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                    id: ''
+                };
+                overlayObj.getLabel = jasmine.createSpy().and.callFake(function () {
+                    return '';
+                });
+                expect(overlayObj.getLabel(row)).toBe('');
+            });
+
+            it('Check getLabel and return string if id found', function () {
+                var row ={
```
Nazar65:
```suggestion
                var row = {
```

Nazar65 (changes requested):

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                expect(overlayObj.hasOwnProperty('initObservable')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.initObservable;
+
+                expect(type).toEqual('function');
+            });
+
+            it('licensed will be defined', function () {
```
filmaj, line 56: This test does not provide any value as we set the `licensed` variable to an object in the `beforeEach`. Thus, this test will always pass. I would suggest either remove the test or remove setting the `licensed` variable in the `beforeEach`.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+
+            var handler = {
+                success: function(data) {
+                    return [1,2]
+                },
+                fail: function(data) {
+                    return 'Error Message';
+                }
+            };
+            var testObj = {
```
filmaj: Hmm, why create a new Ajax-invoking method here housed in a `testObj`? I think we could remove this object and instead, in the `updateLicensed` test below, invoke `Overlay.updateLicensed` directly.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+            it('Check for defined', function () {
+                expect(overlayObj.hasOwnProperty('updateLicensed')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.updateLicensed;
+               expect(type).toEqual('function');
+            });
+
+            it("Check Ajax Success request", function() {
```
filmaj: This test needs to be reworked. This test is only exercising the fake variables/methods that are created in the housing `describe` block above, and not actually exercising the `updateLicensed` method. At a minimum, this test should:
- stub (or spy on) `$.ajax`, which you have set up properly below, to prevent an actual network request, and manipulate the ajax call to return the data you want to fake to exercise the functionality you want (in this case, you want to fake return ajax "success" data).
- invoke `Overlay.updateLicensed`

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                        'response': 'success'
+                    }).promise();
+                });
+                testObj.ajaxFunction();
+                expect(handler.success).toHaveBeenCalled();
+                expect(typeof handler.success()).toBe('object');
+                expect(handler.success().length).toBe(2);
+            });
+
+            it('Check Ajax failure request', function() {
```
filmaj: Same comments as for the ajax success test above, this one should invoke `Overlay.updateLicensed`, and `$.ajax` should be spied on to ensure it returns proper "ajax failure" data.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+
+                expect(type).toEqual('function');
+            });
+
+            it('licensed will be defined', function () {
+                expect(overlayObj.licensed).toBeDefined();
+            });
+        });
+
+        describe('"updateLicensed" method', function () {
```
filmaj, line 61: While you have 'ajax' call success and failure test cases in this `describe` block, the `updateLicensed` method also does something special if the user is not logged in: it does not make an ajax call at all (it `return`s early). I think we should add a test for this.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+            it('Check for defined ', function () {
+                expect(overlayObj.hasOwnProperty('getIds')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.getIds;
+                expect(type).toEqual('function');
+            });
+
+            it('Check returned value type if method called with object argument', function () {
```
filmaj: This test is not valuable as you are stubbing out the `getIds` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+
+                overlayObj.getIds = jasmine.createSpy().and.callFake(function () {
+                    return [1,2,3];
+                });
+
+                expect(typeof overlayObj.getIds()).toEqual('object');
+            });
+
+            it('Check getIds and return array', function () {
```
filmaj: This test is not valuable as you are stubbing out the getIds directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+            it('Check for defined ', function () {
+                expect(overlayObj.hasOwnProperty('getStyles')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.getStyles;
+                expect(type).toEqual('function');
+            });
+
+            it('Check returned value type if method called with object argument', function () {
```
filmaj: This test is not valuable as you are stubbing out the `getStyles` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                        top: '100px'
+                    };
+                });
+
+                expect(typeof overlayObj.getStyles()).toEqual('object');
+            });
+
+            it('Check getStyles and return array', function () {
+
+                overlayObj.getStyles = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `getStyles` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                var type = typeof overlayObj.isVisible;
+                expect(type).toEqual('function');
+            });
+
+            it('Check returned value type if method called with object argument', function () {
+                var row = {
+                    id: 123213
+                };
+
+                overlayObj.isVisible = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `isVisible` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                });
+
+                expect(typeof overlayObj.isVisible(row)).toEqual('boolean');
+            });
+
+            it('Check called "this.isVisible" method with object argument', function () {
+                var row = {
+                    id: 123213
+                };
+                overlayObj.isVisible = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `isVisible` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                });
+                overlayObj.isVisible(row);
+                expect(overlayObj.isVisible).toHaveBeenCalledWith(row);
+            });
+
+            it('Check isVisible and return false if id doesnt found', function () {
+                var row = {
+                    id: ''
+                };
+                overlayObj.isVisible = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `isVisible` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                    return false;
+                });
+                expect(overlayObj.isVisible(row)).toBe(false);
+            });
+
+            it('Check isVisible and return true if id found', function () {
+                var row = {
+                    id: 123213
+                };
+                overlayObj.isVisible = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `isVisible` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                    id: 123213
+                };
+                expect(typeof overlayObj.getLabel(row)).toEqual('string');
+            });
+
+            it('Check called "this.getLabel" method with object argument', function () {
+                var row ={
+                    id: 123213
+                };
+                overlayObj.getLabel = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `getLabel` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                });
+                overlayObj.getLabel(row);
+                expect(overlayObj.getLabel).toHaveBeenCalledWith(row);
+            });
+
+            it('Check getLabel and return null if id doesnt found', function () {
+                var row ={
+                    id: ''
+                };
+                overlayObj.getLabel = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `getLabel` directly before testing it. Either don't stub the method you are testing, or remove this test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,283 @@
...
+                    return '';
+                });
+                expect(overlayObj.getLabel(row)).toBe('');
+            });
+
+            it('Check getLabel and return string if id found', function () {
+                var row ={
+                    id: 123213
+                };
+                overlayObj.getLabel = jasmine.createSpy().and.callFake(function () {
```
filmaj: This test is not valuable as you are stubbing out the `getLabel` directly before testing it. Either don't stub the method you are testing, or remove this test.

filmaj (changes requested):
Thanks for writing up the tests!

I see a major conceptual problem with many of the tests, though: the method-under-test is being stubbed (i.e. temporarily overwritten inside the test) before it gets executed and assertions are ran on the output. This means we are not actually testing the implemented method! I think a rule needs to be followed: if you are testing a method, don't stub it (or spy on it). You can stub or spy on other methods the method-under-test uses, but not the actual method.

Also, be careful with the creation of spies directly in jasmine (using `jasmine.createSpy()`). The lifecycle of spies created in this manner is not managed by jasmine and the test lifecycle. This means that spies like this are left in place forever! This can have unintended consequences across different tests. If you overwrite a class method with a spy directly like this in one test:

```
overlayObj.getLabel = jasmine.createSpy().and.callFake(function () { console.log('hello'); });
```

... in a different test, the overwritten `getLabel` method will remain, and the original method will be gone. Instead, use the jasmine `spyOn` global method, which is hooked into the test lifecycle. Any methods overwritten via `spyOn` will be properly cleaned up/removed and the original method reinstated at the end of a test (i.e. `it` block) or a suite of tests (i.e. `describe` block). For more information, please read through the [jasmine introduction](https://jasmine.github.io/2.0/introduction.html), and especially the [section on spies](https://jasmine.github.io/2.0/introduction.html#section-Spies).

sivaschenko:
@magento run Unit Tests

konarshankar07:
Thanks @filmaj for detail explanation and I will work on the requested changes

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
filmaj: @konarshankar07 as mentioned I think a test checking that no ajax call is issued if the user is not authorized should be added.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,147 @@
...
+                    return ids;
+                });
+                spyOn($, 'ajax').and.callFake(function(e) {
+                    return $.Deferred().resolve({
+                        'response': 'success'
+                    }).promise();
+                });
+                overlayObj.updateLicensed();
+                expect(overlayObj.updateLicensed()).toBeUndefined();
+                expect($.ajax).toHaveBeenCalled();
```
filmaj, line 127: Nice job on this test rewrite! You are on the right track.

In addition to making sure `$.ajax` was called, we also want to make sure that the success logic for a successful AJAX request was executed. In this case, when the AJAX request completes successfully and we have received data from the serve, the `updateLicensed` sets the `licensed` property to the response object received from the AJAX call. So we should assert that this is the case.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,147 @@
...
+                expect(overlayObj.getIds).toHaveBeenCalled();
+            });
+
+            it('Check Ajax failure request', function() {
+                var login = {
+                    user: function () {
+                        return {
+                            isAuthorized: false
```
filmaj, line 138: Stubbing `isAuthorized` out to return `false` doesn't actually execute the AJAX failure flow - rather, it would execute the 'not-logged-in-so-won't-issue-an-AJAX-request' flow. So I would suggest either renaming and restructuring this test so that it reflects this, or re-write the test to actually exercise the 'AJAX request has failed' flow.

Friendly reminder that the `updateLicensed` method has three main logic flows:

1. User is not authorized, so no AJAX request is issued, the `licensed` property is set to an empy object `{}` and the method `return`s early.
2. User is authorized, so we issue the AJAX request and after receiving a successful response from the server, we set the `licensed` property to the server's response.
3. User is authorized, so we issue the AJAX request and after receiving an error response from the server or  the AJAX request  fails, we `return` the error message.

As such, to completely cover the logic of the `updateLicensed` method with unit tests, we should create three tests for this method, one to cover each main logical flow.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,147 @@
...
+                });
+                spyOn($, 'ajax');
+                overlayObj.updateLicensed();
+                expect(overlayObj.updateLicensed()).toBeUndefined();
+                expect($.ajax).not.toHaveBeenCalled();
+                expect(overlayObj.getIds).not.toHaveBeenCalled();
+            });
+        });
+
+        describe('"getIds" method', function () {
```
filmaj, line 158: If we want more complete unit test coverage  for the `getIds` method, then we should exercise  the logic of this method, that is, stub out `masonry().rows()` with a spy to return some fake data controlled in the test, and assert that the array of ids returned by the method match the fake data set up in the test.

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,147 @@
...
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.getIds;
+                expect(type).toEqual('function');
+            });
+
+        });
+
+        describe('"getStyles" method', function () {
```
filmaj, line 178: Same as `getIds`, this set of tests do not exercise `getStyles` logic. It seems to me like the logic of `getStyles` may fail in certain cases. The logic that calculates the `height` variable in this method, for example, is mixing string and numbers in JavaScript and using arithmetic operators on mixed string and number types can yield unexpected results 😄. I think it would be a good idea to write logic tests for this method as I think we may catch a bug here 😄

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,147 @@
...
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.getStyles;
+                expect(type).toEqual('function');
+            });
+
+        });
+
+        describe('"isVisible" method', function () {
```
filmaj, line 212: Once more, we should add logical tests here. Since `isVisible` returns either `true` or `false`, I would write  two additional tests  for this method: when does it return `true`, and when does it return `false`?

dev/tests/js/jasmine/tests/app/code/Magento/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/overlay.test.js
```diff
@@ -0,0 +1,147 @@
...
+                expect(overlayObj.hasOwnProperty('isVisible')).toBeDefined();
+            });
+
+            it('Check method type', function () {
+                var type = typeof overlayObj.isVisible;
+                expect(type).toEqual('function');
+            });
+        });
+
+        describe('"getLabel" method', function () {
```
filmaj, line 249: The  logic tests for `getLabel` would be very similar as `isVisible`, except  the output would be either the empty string or the string `"Licensed"`. Let's add those tests here.

filmaj (changes requested):
Nicely done Shankar! Much better  use of stubs in this iteration.

I have added a few more comments mostly around fleshing out the code coverage to make sure the tests execute most of the important logic in `overlay.js`.

Nice job!

sivaschenko (dismissed):

sivaschenko:
@magento run Static Tests

sivaschenko:
@magento run Static Tests

konarshankar07:
@sivaschenko ... Thanks for the code cleanup

slavvka:
Guys, thank you for the identification of the issue with static tests. We are working on fixing that.

sivaschenko (approved):
@konarshankar07 high five! :raised_hand: We've done it!

filmaj:
Nice work @konarshankar07 !!!!

konarshankar07:
Thanks @filmaj and @sivaschenko for your great support

Issue #874, opened 2019-12-18
[MFTF] Cover Image name popup should not appear when logged in user is saving licensed image if preview version was saved before


chalov-anton:
### Preconditions (*)
1. Installed Magento with Adobe Stock Integration
2. Configured  integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
### Steps to reproduce (*)
Admin user is logged in to admin panel
Navigate to cms/page page
Click Add New Page button
Expand Content fieldset
Click Show/Hide Editor button
Click Insert Image ... button
Click Search Adobe Stock button
Move mouse over the first image in the search result and click on it
User clicks 'License and Save'
Select previously saved licensed image from the Media Gallery
User clicks on 'Delete Selected'
Click "Ok" to confirm
User clicks on Search Adobe Stock
Image details section is expanded
'Save Licensed' button is shown
User signs out from Adobe IMS
'Save Preview' and 'Licence and Save' buttons are displayed
User clicks on 'Save Preview'
User click  Confirm  in popup window
Click 'Search Adobe Stock' button
User signs in to Adobe IMS
User clicks on Save Licensed
### Expected result (*)
No confirmation dialog appears, Image is saved 
Saved image is listed in Media Gallery
### Note(*)
Related issue - https://github.com/magento/adobe-stock-integration/issues/866
Test on Cucumber Studio - https://studio.cucumber.io/projects/131313/test-plan/folders/1054245/scenarios/3995935

Issue #866, opened 2019-12-13
Image name popup should not appear when logged in user is saving licensed image if preview version was saved before


sivaschenko:
### Preconditions (*)
1. Installed Magento with Adobe Stock Integration
2. Configured integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. Login to Admin Panel
2. Open Adobe Stock Integration grid from Media Gallery (i.e. Catalog -> Category -> Select Image)
3. Save preview for the image that is licensed for your account in Adobe Stock
4. Login to Adobe Stock
5. Expand preview for the image from step 3
6. Click "Save" button to save licensed version of the image

### Actual result (*)
The popup appears asking for the image name

### Expected result (*)
Popup should not appear. The preview image file should be replaced by licensed image.

PR #887, opened 2020-01-08
Do not show name confirmation for licensing image with saved preview

rogyar:
### Description (*)
This PR introduces an improvement that hides an image name prompt window in case if the image preview has been previously saved. Instead, the preview image is being replaced by the licensed image (using the name of the saved preview image).

1. #866 : Image name popup should not appear when logged in user is saving licensed image if preview version was saved before
2. #874 : [MFTF] Cover Image name popup should not appear when logged in user is saving licensed image if preview version was saved before

1. Login to Admin Panel
2. Open Adobe Stock Integration grid from Media Gallery (i.e. Catalog -> Category -> Select Image)
3. Save preview for the image that is licensed for your account in Adobe Stock
4. Login to Adobe Stock
5. Expand preview for the image from step 3
6. Click "Save" button to save licensed version of the image

The image name prompt window will not appear. The preview image file will be replaced by the licensed image.


sivaschenko (approved):

sivaschenko:
Hi @rogyar , thanks for the pull request! Is there anything else you'd like to add to it?

sivaschenko (approved):

Issue #830, opened 2019-12-03
Add "Unlicensed" overlay to media gallery grid


sivaschenko:
### See more information in the story:

https://github.com/magento/adobe-stock-integration/issues/460

### Preconditions
1. Install Magento with Adobe Stock Integration
2. Save at least one Adobe Stock image preview to media gallery

### Steps to reproduce (*)
1. Login to Admin Panel
2. Select "Media Gallery Development" from the menu on the right

### Actual result (*)

Images do not include any overlay labels:

![image](https://user-images.githubusercontent.com/2028541/70070972-6f2e6c80-15ec-11ea-9330-ed1d9b07bc30.png)

### Expected result (*)

Images that are saved from Adobe Stock and are not licensed should have "Unlicensed" red overlay label:

![image](https://user-images.githubusercontent.com/2028541/70071106-b61c6200-15ec-11ea-9ce3-b4bc53a8bb90.png)

filmaj:
To clarify, preconditions here include:

- user must be signed into Adobe IMS
- user has saved an image preview

Is the above correct?

sivaschenko:
@filmaj the second precondition is correct. The first one is not - the information about unlicensed assets in media gallery should be retrieved from Magento database (not from API as on Adobe Stock grid) so the login is not required.

Preconditions are updated accordingly

filmaj:
Just a point of clarification: is the licensed/unlicensed DB field implemented already or does that still need to be implemented? And if so, which issue is that tracked in?

sivaschenko:
@filmaj the source field is `adobe_stock_asset.is_licensed` (already there)

PR #892, opened 2020-01-10
Add "Unlicensed" overlay to media gallery grid

Nazar65:
### Description (*)
Add "Unlicensed" overlay to media gallery grid 

1. magento/adobe-stock-integration#830:Add "Unlicensed" overlay to media gallery grid
2. ...

    Login to Admin Panel
    Select "Media Gallery Development" from the menu on the right

### Actual result (*)

Images that are saved from Adobe Stock and are not licensed should have "Unlicensed" red overlay label:
DeepinScreenshot_select-area_20200116142536


AdobeStockImage/Model/AssetIndexer.php
```diff
@@ -0,0 +1,139 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\AdobeStockImage\Model;
+
+use Magento\AdobeStockAsset\Model\AssetRepository;
```
sivaschenko: Please use the AssetRepositoryInterface instead

MediaGalleryUi/Model/ImagesIndexer.php
```diff
@@ -54,6 +54,7 @@ public function __construct(
      */
     public function execute(): void
     {
+        ksort($this->indexers);
```
sivaschenko: please move sorting to assignment to perform it only once

AdobeStockImage/composer.json
```diff
@@ -6,8 +6,10 @@
         "magento/framework": "*",
         "magento/module-adobe-stock-client-api": "*",
         "magento/module-adobe-stock-asset-api": "*",
+        "magento/module-adobe-stock-asset": "*",
```
sivaschenko: Dependency should not include `magento/module-adobe-stock-asset` module

AdobeStockImage/etc/di.xml
```diff
@@ -47,4 +47,11 @@
             </argument>
         </arguments>
     </type>
+    <type name="Magento\MediaGalleryUi\Model\ImagesIndexer">
```
sivaschenko: An interface should be created in MediaGalleryUiApi module and used in di instead of actual implementation class

MediaGalleryUi/Model/AssetIndexer.php
```diff
@@ -71,13 +71,16 @@ public function __construct(
     public function execute(\SplFileInfo $item): void
     {
         $file = $item->getPath() . '/' . $item->getFileName();
+        $relativePath = $this->getMediaDirectory()->getRelativePath($file);
 
+        $path = dirname($relativePath) === '.' ? '/' . $relativePath : $relativePath;
```
sivaschenko: Please extract path calculation to the `private function getPath(\SplFileInfo $item): string`

MediaGalleryUi/view/adminhtml/ui_component/media_gallery_listing.xml
```diff
@@ -94,6 +94,7 @@
         <column name="overlay" component="Magento_Ui/js/grid/columns/overlay">
             <settings>
                 <label translate="true">Overlay</label>
+                <bodyTmpl>Magento_MediaGalleryUi/grid/columns/overlay</bodyTmpl>
```
sivaschenko: It's possible to use `overlay` key in the grid data to output the required overlay without the need to override body template:
 - `overlay: "Unlicensed"` results into "Unlicensed" overlay
- `overlay: false` (or just missing the overlay key) results into no overlay

sivaschenko (changes requested):
Hi @Nazar65 thanks for the pull request! Please see my review comments

AdobeStockImage/Model/AssetIndexer.php
```diff
@@ -0,0 +1,164 @@
...
+    public function execute(\SplFileInfo $item): void
+    {
+        $mediaAsset = $this->getByPathCommand->execute($this->getPathFormMediaAsset($item));
+
+        $this->searchCriteriaBuilder->addFilter('media_gallery_id', $mediaAsset->getId());
+        $searchCriteria = $this->searchCriteriaBuilder->create();
+
+        $assets = $this->assetRepository->getList($searchCriteria)->getItems();
+
+        if (count($assets) > 0) {
```
sivaschenko: I don't think this check is necessary, if $assets is an empty array the foreach loop will be simply skipped

AdobeStockImage/Model/AssetIndexer.php
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param \SplFileInfo $item
+     * @return string
+     */
+    private function getPathFormMediaAsset(\SplFileInfo $item): string
+    {
+        $path = $this->getMediaDirectory()->getRelativePath($item->getPath() . '/' . $item->getFileName());
+
+        if ($this->driver->getParentDirectory($path) === '.') {
+            $path = '/' . $path;
```
sivaschenko, line 125: It's not clear to me why do we need this transformation, can you please provide an example

MediaGalleryUi/Console/Command/IndexAssets.php
```diff
@@ -44,6 +56,7 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $this->state->setAreaCode(Area::AREA_ADMINHTML);
```
sivaschenko: Do we need to set area code in the command if it's also emulated in `UpdateAssetInGrid.php` ?

MediaGalleryUi/Ui/Component/Listing/Columns/UnlicensedOverlay.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    private function isLicensed(array $item): bool
+    {
+        $isLicensed = true;
+
+        if (isset($item['licensed']) &&  !$item['licensed']) {
+            $isLicensed = false;
+        }
+
+        return $isLicensed;
```
sivaschenko:
```suggestion
        return isset($item['licensed']) && $item['licensed'];
```

sivaschenko (changes requested):
Thanks for updates @Nazar65 ! Please see my review comments

MediaGalleryUi/Console/Command/IndexAssets.php
Nazar65: When I try save asset without area code exception is shown
DeepinScreenshot_select-area_20200116151447

sivaschenko (approved):

PR #899, opened 2020-01-20
Moved media gallery to the slide panel

sivaschenko:
The new media gallery is now moved to a slide panel.

To work with the new media gallery - ensure "Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery -> Enabled" is set to "Yes"

To switch to the old media gallery - set "Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery -> Enabled" to "No"


MediaGalleryUi/Plugin/UpdateOpenDialogUrl.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        if (!$this->config->isEnabled()) {
+            return;
+        }
+
+        $data = $component->getData();
+
+        $data['config']['mediaGallery']['openDialogUrl'] = $this->url->getUrl(
+            'media_gallery/index/index',
+            ['_secure' => true]
```
Nazar65: If we will have a non secure connection ? will url be correct ?

Nazar65 (changes requested):

MediaGalleryUi/Plugin/UpdateOpenDialogUrl.php
sivaschenko: Thanks for pointing this out, I've removed the force secure flag

PR #939, opened 2020-01-27
create-common-const-for-index-routepath

RakeshJesadiya:
Create Common Constant for the Route Path of Media gallery Index controller.

### DIFF
--- a/MediaGalleryUi/Model/Config.php
+++ b/MediaGalleryUi/Model/Config.php
@@ -22,6 +22,11 @@ class Config implements ConfigInterface
     private const XML_PATH_ENABLED = 'system/media_gallery/enabled';
 
     /**
+     * Route Path to Media Gallery Index Controller.
+     */
+    const MEDIA_GALLERY_INDEX_ROUTE_PATH = 'media_gallery/index/index';
+
+    /**
      * @var ScopeConfigInterface
      */
     private $scopeConfig;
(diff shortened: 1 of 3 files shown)


sivaschenko (changes requested):
Hi @RakeshJesadiya thanks for the pull request! We are planning to move the plugins to a separate module. I think it's better not to link them with the configuration at this time.

MediaGalleryUi/Plugin/UpdateOpenDialogUrl.php
```diff
@@ -8,8 +8,9 @@
 namespace Magento\MediaGalleryUi\Plugin;
 
 use Magento\Framework\UrlInterface;
-use Magento\Ui\Component\Form\Element\DataType\Media\Image;
```
sivaschenko, line 11: Please do not reoder `use` blocks (or reorder it by alphabet)

MediaGalleryUi/Model/Config.php
```diff
@@ -21,6 +21,11 @@ class Config implements ConfigInterface
      */
     private const XML_PATH_ENABLED = 'system/media_gallery/enabled';
 
+    /**
+     * Route Path to Media Gallery Index Controller.
+     */
+    const MEDIA_GALLERY_INDEX_ROUTE_PATH = 'media_gallery/index/index';
```
sivaschenko, line 27: Please create two private constants instead of one public - there is a possibility the URLs can be different for openDialog and wysiwyg

sivaschenko (changes requested):
Thanks for the pull request @RakeshJesadiya ! Please see my review comments

sivaschenko:
Closing due to inactivity

Issue #933, opened 2020-01-24
Adobe stock icon is broken in media gallery after index


sivaschenko:
### Preconditions (*)
1. Install Magento and Adobe Stock Integration [See documentation](https://github.com/magento/adobe-stock-integration/wiki/Installation-Guide)

### Steps to reproduce (*)
1. Login to Admin panel
2. Turn OFF Enhanced Media Gallery (Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery)
3. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
4. Open the menu for the preview image saved from Adobe Stock
5. Save any image preview
6. Turn ON Enhanced Media Gallery (Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery)
7. Execute `delete from media_gallery_asset_grid` on database
8. Execute `bin/magento media-gallery:index`
9. Open Media gallery again

### Expected result (*)
Adobe Stock icon should be displayed under the saved image preview

### Actual result (*)
Adobe Stock icon is broken

![image](https://user-images.githubusercontent.com/2028541/73076452-0bbb0f80-3eb6-11ea-8da6-ed41401f732b.png)

### Additional info

The icon is displayed correctly if it's added on image save instead of index operation

PR #940, opened 2020-01-30
Fix the Adobe Stock asset source icon url set by indexer

coderimus:
### Description (*)
This PR provides code and some possible business logic improvement for the enhancement media gallery image index.

**Issue related changes:**
Source image URL generates on the fly when the media gallery grid renders.

**Refactoring**
Remove source_icon_ur URL from the media grid. Remove the code which is not used for the new way of source image URL generation.

#933 

Check the information at the #933 + test save the image and media grid reindex with the enhanced media gallery being turned off.


sivaschenko (changes requested):
Thanks for the PR @coderimus 

Only source code should be stored in the grid table. The icon url should be generated on rendering

MediaGalleryUi/Model/UpdateAssetInGrid.php
```diff
@@ -98,34 +115,27 @@ private function getConnection(): AdapterInterface
     }
 
     /**
-     * Return AStock incon url if source is Adobe Stock
+     * Return AStock icon url if source is Adobe Stock
+     *
+     * @param AssetInterface $asset
      *
-     * @return ?string
+     * @return string|null ?string
```
hws47a:
```suggestion
     * @return string|null
```

MediaGalleryUi/Model/UpdateAssetInGrid.php
```diff
@@ -32,59 +33,75 @@ class UpdateAssetInGrid
...
         $this->assetRepository = $assetRepository;
-        $this->appEmulation = $emulation;
         $this->scopeConfig = $scopeConfig;
+        $this->file = $file;
+        $this->logger = $logger;
     }
 
     /**
-     * Update the grid table for the asset
+     * Update the media asst grid with additional data
```
hws47a:
```suggestion
     * Update the media asset grid with additional data
```

hws47a (changes requested):

coderimus:
@sivaschenko discussed changes according to the way the source image is presented to the grid were done. Please, review this PR.
image

MediaGalleryUi/Model/UpdateAssetInGrid.php
```diff
@@ -27,110 +24,63 @@ class UpdateAssetInGrid
...
-            true
-        );
-
-        if (!empty($asset->getSource())) {
-            $iconUrl = $this->assetRepository->getUrlWithParams(
-                'Magento_MediaGalleryUi::images/Astock.png',
-                ['_secure' => $this->getIsSecure()]
+        try {
+            $this->resource->getConnection()->insertOnDuplicate(
+                $this->resource->getTableName('media_gallery_asset_grid'),
```
hws47a, line 65: Could be better to use a constant from resource model

MediaGalleryUi/Model/UpdateAssetInGrid.php
```diff
@@ -27,110 +24,63 @@ class UpdateAssetInGrid
...
+                [
+                    'id' => $asset->getId(),
```
hws47a, line 67: Could be better to have keys as constants in resource model or interface

MediaGalleryUi/Ui/Component/Listing/Columns/ConstructSourceItemUrl.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $sourceIconUrl = null;
+        if (!empty($this->imageSource) && isset($this->imageSource[$sourceName])) {
+            $sourceIconUrl = $this->assetRepository->getUrlWithParams(
+                $this->imageSource[$sourceName],
+                ['_secure' => $this->getIsSecure()]
+            );
+        }
+
+        return $sourceIconUrl;
```
hws47a: Could be a little simpler and cleaner like this:
```php
public function execute(string $sourceName): ?string
{
    return isset($this->imageSource[$sourceName])
        ? $this->assetRepository->getUrlWithParams(
                $this->imageSource[$sourceName],
                ['_secure' => $this->getIsSecure()]
            )
        : null;
}
```

MediaGalleryUi/Ui/Component/Listing/Columns/SourceIconProvider.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Prepare Data Source
+     *
+     * @param array $dataSource
+     * @return array
+     */
+    public function prepareDataSource(array $dataSource): array
+    {
+        if (isset($dataSource['data']['items'])) {
```
hws47a:
```suggestion
        if (isset($dataSource['data']['items']) && is_iterable($dataSource['data']['items'])) {
```

MediaGalleryUi/Ui/Component/Listing/Columns/SourceIconProvider.php
```diff
@@ -0,0 +1,58 @@
...
+            foreach ($dataSource['data']['items'] as &$item) {
+                $item['source_icon_url'] =
+                    (isset($item['source'])) ? $this->constructSourceItemUrl->execute($item['source']) : null;
```
hws47a:
```suggestion
                    isset($item['source']) ? $this->constructSourceItemUrl->execute($item['source']) : null;
```

MediaGalleryUi/Ui/Component/Listing/Columns/UnlicensedOverlay.php
```diff
@@ -36,13 +38,14 @@ public function __construct(
...
      * @param array $dataSource
+     *
      * @return array
      */
     public function prepareDataSource(array $dataSource): array
     {
         if (isset($dataSource['data']['items'])) {
             foreach ($dataSource['data']['items'] as & $item) {
-                $item[$this->getData('name')] = (isset($item['licensed']) &&  !$item['licensed']) ? 'Unlicensed' : '';
+                $item[$this->getData('name')] = (isset($item['licensed']) && !$item['licensed']) ? 'Unlicensed' : '';
```
hws47a:
```suggestion
                $item[$this->getData('name')] = isset($item['licensed']) && !$item['licensed'] ? 'Unlicensed' : '';
```

hws47a (dismissed):
It looks good. I've written a few suggestions.

MediaGalleryUi/Model/UpdateAssetInGrid.php
coderimus: Hi @hws47a ,

Thank you for the suggestions. I implemented the majority of them except this one. Can we have a short conversation about this in comments here?

I think that putting constants into interfaces (with the api notation) and resource models (without api notations or with it) makes implementation more coupled and dependent on the source model. Also, as I can see, in this project we commonly do not use constants in interfaces or source models.

Should we change this way @sivaschenko ?

MediaGalleryUi/Model/UpdateAssetInGrid.php
sivaschenko: Сonsidering that we do not use the resource model for this indexer, I wouldn't use the resource model constants

MediaGalleryUi/Model/UpdateAssetInGrid.php
sivaschenko: Currently we don't have the model so the only place where these fields are declared is db schema. Let's discuss if we need to introduce the model/interface at the next public meeting

sivaschenko (dismissed):

hws47a (dismissed):

sivaschenko (approved):

Issue #685, opened 2019-11-11
AdminAdobeStockLocalizationTest fails consistently


filmaj:
### Steps to reproduce (*)

Run the `AdminAdobeStockLocalizationTest`:

    vendor/bin/mftf run:test AdminAdobeStockLocalizationTest --remove

### Expected result (*)

Passes without failure.

### Actual result (*)

Log when running in headless Chrome on Mac OS:

```
1) AdminAdobeStockLocalizationTestCest: Admin adobe stock localization test
 Test  tests/functional/Magento/FunctionalTest/_generated/adobe_stock_integration_suite/AdminAdobeStockLocalizationTestCest.php:AdminAdobeStockLocalizationTest
 Step  See element "//div[@class='keyword']//span[text()='Авто']"
 Fail  Failed asserting that an array is not empty.

Scenario Steps:

 75. // Exiting Action Group [logout] logout
 74. $I->amOnPage("/admin_hgkq1l/admin/auth/logout/") at tests/functional/Magento/FunctionalTest/_generated/adobe_stock_integration_suite/AdminAdobeStockLocalizationTestCest.php:83
 73. // Entering Action Group [logout] logout
 72. // Exiting Action Group [setAdminInterfaceLocaleToDefaultValue] SetAdminAccountActionGroup
 71. $I->see("You saved the account.","#messages div.message-success") at tests/functional/Magento/FunctionalTest/_generated/adobe_stock_integration_suite/AdminAdobeStockLocalizationTestCest.php:80
 70. $I->waitForElementVisible("#messages div.message-success",10) at tests/functional/Magento/FunctionalTest/_generated/adobe_stock_integration_suite/AdminAdobeStockLocalizationTestCest.php:79
```

### More Context

I think this is failing as we are searching for a generic search term, and looking at the first image preview expansion and expecting a particular keyword to be present. However, the images that come up when using this generic search term may rotate and change.

I will try to, for example, search for a specific image ID, and then assert that one of that image's specific keywords are present and localized.

filmaj:
I will tackle this one.

PR #941, opened 2020-01-31
Fixed failing localized image preview keyword assertion

rogyar:
### Description (*)
This PR introduces a fix for failing keyword localization MFTF test. It also refactors the test body in order to meet the best MFTF development practices.

1. #685 : AdminAdobeStockLocalizationTest fails consistently

1. Run `vendor/bin/mftf run:test AdminAdobeStockLocalizationTest --remove`.
2. Go and make some tea.
3. The test should pass successfully.


filmaj (approved):
LGTM.

If we want to make this test more resilient, I would suggest that, instead of searching for a generic search term, we search for a specific image ID. This way we can be sure of the attribute list associated with the image. By using  a generic search term, we are not guaranteed which image will be retrieved, meaning we are not guaranteed if the keyword being asserted  on will be present.

Issue #905, opened 2020-01-23
Search by keyword field for Media Gallery


sivaschenko:
### Story

[[Story #45] User filter images by their keywords in the Media Gallery](https://github.com/magento/adobe-stock-integration/issues/725)

### Preconditions (*)
1. Install Magento and Adobe Stock Integration [See documentation](https://github.com/magento/adobe-stock-integration/wiki/Installation-Guide)

### Steps to reproduce (*)
1. Login to Admin panel
2. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)

### Expected result (*)
Search by keyword field is available for media gallery. The search results into images with matching title or keywords.
![image](https://user-images.githubusercontent.com/2028541/73017877-6aca4700-3e18-11ea-8a57-e014f8a966f9.png)

### Technical details (*)
Concatenated keywords should be added to `media_gallery_asset_grid` table to enable search

PR #942, opened 2020-02-01
905: searh by keyword filter

coderimus:
### Description (*)
This PR provides the next functionality:

After the Adobe Stock keywords save action all related keywords will be concatenated and saved to the `media_gallery_asset_grid.keywords` column.

The search by keywords consists of full text search by `media_gallery_asset_grid.keywords` and `media_gallery_asset_grid.name`

#905 

N/A


MediaGalleryUi/view/adminhtml/ui_component/media_gallery_listing.xml
```diff
@@ -35,6 +36,7 @@
         </dataProvider>
     </dataSource>
     <listingToolbar name="listing_top">
+        <filterSearch name="keywords"/>
```
sivaschenko: I think it could be more optimal to reuse the out of the box fulltext type. Should work without any changes to data provider and filter classes. Also it would be good if the search will consider not only keywords but all the text fields in the grid table (that should also work by default).
```suggestion
        <filterSearch name="fulltext"/>
```

MediaGalleryUi/etc/db_schema.xml
```diff
@@ -18,6 +18,7 @@
         <column xsi:type="int" name="height" padding="10" unsigned="true" nullable="false" identity="false" default="0" comment="Height"/>
         <column xsi:type="datetime" name="created_at" nullable="true" comment="Uploaded Date"/>
         <column xsi:type="datetime" name="updated_at" nullable="true" comment="Modification Date"/>
+        <column xsi:type="varchar" name="keywords" length="500" nullable="true"  comment="Keywords"/>
```
sivaschenko: why `length="500"`? let's probably make it text

sivaschenko (changes requested):
Thanks for the pull request @coderimus  ! Please take a look at my code review notes

MediaGalleryUi/view/adminhtml/ui_component/media_gallery_listing.xml
coderimus: @sivaschenko to add the default full-text search we need to add the full-text index for desired columns. This can be done in 2 ways: mview or general full-text index. I am not totally sure about this. Please, check.

sivaschenko (dismissed):

sivaschenko:
@magento run Unit Tests

Issue #903, opened 2020-01-23
Database storage and thumbnails are not handled when deleting a preview image


sivaschenko:
### Preconditions (*)
1. Install Magento and Adobe Stock Integration [See documentation](https://github.com/magento/adobe-stock-integration/wiki/Installation-Guide)
2. Have Adobe Stock credits (contact @sivaschenko)
3. Use database media storage: https://docs.magento.com/m2/ce/user_guide/system/media-storage-database.html#use-a-database-for-media-storage

### Steps to reproduce (*)
1. Login to Admin panel
2. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
3. Click "Search Adobe Stock"
4. Save preview of one of the images
5. License the same image

### Actual result (*)
As part of licensing the image that was previously saved as a preview image - the existing preview image is deleted. Database storage and thumbnails are not cleaned when the preview image is deleted

### Expected result (*)
`\Magento\AdobeStockImage\Model\Storage\Delete::execute` should use `\Magento\Cms\Model\Wysiwyg\Images\Storage::deleteFile` to delete image

PR #945, opened 2020-02-03
Handle DB Storage and thumbnails when deleting a preview image

Nazar65:
### Description (*)
Database storage and thumbnails are not handled when deleting a preview image

1. magento/adobe-stock-integration#903: Handle DB STorage ant thumbnails when deleting a preview image

1. Login to Admin panel
2. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
3. Click "Search Adobe Stock"
4. Save preview of one of the images
5. License the same image

### DIFF
--- a/AdobeStockImage/Model/Storage/Delete.php
+++ b/AdobeStockImage/Model/Storage/Delete.php
@@ -9,9 +9,7 @@ declare(strict_types=1);
 namespace Magento\AdobeStockImage\Model\Storage;
 
 use Magento\Cms\Model\Wysiwyg\Images\Storage;
-use Magento\Framework\App\Filesystem\DirectoryList;
 use Magento\Framework\Exception\CouldNotDeleteException;
-use Magento\Framework\Filesystem;
 use Psr\Log\LoggerInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


AdobeStockImage/Model/Storage/Delete.php
```diff
@@ -52,7 +62,7 @@ public function execute(string $path): void
         try {
             $mediaDirectory = $this->filesystem->getDirectoryWrite(DirectoryList::MEDIA);
             if ($mediaDirectory->isFile($path)) {
```
sivaschenko: `isFile` is not necessary as it's performed by the `storage`. Probably we can remove this class completely and use the storage directly instead.

sivaschenko (changes requested):
Thanks for the pull request @Nazar65 ! See my suggestions in comments

AdobeStockImage/Model/Storage/Delete.php
Nazar65: ok

Nazar65:
Done :heavy_check_mark:

sivaschenko (approved):

Issue #934, opened 2020-01-24
Unlicensed overlay is absent for saved Adobe Stock image preview without reindex


sivaschenko:
### Preconditions (*)
1. Install Magento and Adobe Stock Integration [See documentation](https://github.com/magento/adobe-stock-integration/wiki/Installation-Guide)

### Steps to reproduce (*)
1. Login to Admin panel
2. Turn OFF Enhanced Media Gallery (Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery)
3. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
4. Open the menu for the preview image saved from Adobe Stock
5. Save any image preview
6. Turn ON Enhanced Media Gallery (Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery)
7. Open Media gallery again

### Expected result (*)
The saved image preview does not have the "Unlicensed" overlay

### Actual result (*)
The saved image preview should have the "Unlicensed" overlay

### Additional info

The overlay is displayed correctly if it's added during the index operation

PR #947, opened 2020-02-04
Set unlicensed overlay for saved Adobe Stock image preview without reindex

Nazar65:

1. magento/adobe-stock-integration#934: IUnlicensed overlay is absent for saved Adobe Stock image preview without reindex
2. ...

1.  Login to Admin panel
2.  Turn OFF Enhanced Media Gallery (Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery)
3.  Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
4.  Open the menu for the preview image saved from Adobe Stock
5.  Save any image preview
6.  Turn ON Enhanced Media Gallery (Stores -> Configuration -> Advanced -> System -> Enhanced Media Gallery)
7.  Open Media gallery again

**The saved image preview does not have the "Unlicensed" overlay**

### DIFF
--- a/AdobeStockImage/Model/AssetIndexer.php
+++ b/AdobeStockImage/Model/AssetIndexer.php
@@ -10,7 +10,7 @@ namespace Magento\AdobeStockImage\Model;
 use Magento\AdobeStockAssetApi\Api\AssetRepositoryInterface;
 use Magento\Framework\Api\SearchCriteriaBuilder;
 use Magento\Framework\App\Filesystem\DirectoryList;
-use Magento\Framework\App\ResourceConnection;
+use Magento\AdobeStockImage\Model\SetUnlicensedImageMediaGallery;
 use Magento\Framework\Filesystem;
 use Magento\Framework\Filesystem\Directory\ReadInterface;
 use Magento\Framework\Filesystem\Driver\File;
@@ -62,11 +62,11 @@ class AssetIndexer implements IndexerInterface
      * Constructor
      *
      * @param GetByPathInterface $getByPathCommand
-     * @param ResourceConnection $resource
      * @param AssetRepositoryInterface $assetRepository
      * @param SearchCriteriaBuilder $searchCriteriaBuilder
      * @param Filesystem $filesystem
      * @param File $driver
+     * @param SetUnlicensedImageMediaGallery $setUnlicensedImagesMediaGalley
      */
     public function __construct(
         GetByPathInterface $getByPathCommand,
--- a/AdobeStockImage/Model/AssetIndexer.php
+++ b/AdobeStockImage/Model/AssetIndexer.php
@@ -12,7 +12,7 @@ use Magento\Framework\Api\SearchCriteriaBuilder;
 use Magento\Framework\App\Filesystem\DirectoryList;
 use Magento\Framework\Filesystem;
 use Magento\Framework\Filesystem\Directory\ReadInterface;
-use Magento\Framework\Filesystem\Driver\File;
+use Magento\Framework\FileSystem\DriverInterface;
 use Magento\MediaGalleryApi\Model\Asset\Command\GetByPathInterface;
 use Magento\MediaGalleryUi\Model\Filesystem\IndexerInterface;
 
@@ -48,7 +48,7 @@ class AssetIndexer implements IndexerInterface
     private $mediaDirectory;
 
     /**
-     * @var File
+     * @var DriverInterface
      */
     private $driver;
 
@@ -64,7 +64,7 @@ class AssetIndexer implements IndexerInterface
      * @param AssetRepositoryInterface $assetRepository
      * @param SearchCriteriaBuilder $searchCriteriaBuilder
      * @param Filesystem $filesystem
-     * @param File $driver
+     * @param DriverInterface $driver
      * @param SetLicensedInMediaGalleryGrid $setLicensedInMediaGalleryGrid
      */
     public function __construct(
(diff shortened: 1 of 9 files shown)


AdobeStockImage/Model/SetUnlicensedImageMediaGallery.php
```diff
@@ -0,0 +1,77 @@
...
+        try {
+            $this->getConnection()->insertOnDuplicate(
+                $this->resource->getTableName(self::MEDIA_GALLERY_ASSET_GRID_TABLE),
+                [
+                    'id' => $asset->getMediaGalleryId(),
+                    'licensed' => $asset->getIsLicensed()
+                ]
+            );
+        } catch (LocalizedException $exception) {
+            $this->logger->critical($exception->getMessage());
```
coderimus: Maybe we should throw some exception here to notify that something went wrong? Otherwise, this critical error keeps silence for the admin user.

AdobeStockImage/Model/SetUnlicensedImageMediaGallery.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        try {
+            $this->getConnection()->insertOnDuplicate(
+                $this->resource->getTableName(self::MEDIA_GALLERY_ASSET_GRID_TABLE),
+                [
+                    'id' => $asset->getMediaGalleryId(),
+                    'licensed' => $asset->getIsLicensed()
+                ]
+            );
+        } catch (LocalizedException $exception) {
```
coderimus: `LocalizedException` is a child from the`\Exception`. I think that it would be better to catch the parent exception type here because it covers all cases. Catch the specific type is a good idea when a specific logic executes specific exception type. Here, as I understood, we need only to handle any kind of exception. What do you think about this?

Nazar65: Good catch :+1: i agree with you

AdobeStockImage/Model/AssetIndexer.php
```diff
@@ -55,34 +54,34 @@ class AssetIndexer implements IndexerInterface
     private $driver;
 
     /**
-     * @var ResourceConnection
+     * @var SetUnlicensedImageMediaGallery
      */
-    private $resource;
+    private $setUnlicensedImagesMediaGalley;
```
sivaschenko:
```suggestion
    private $setLicensedInMediaGalleryGrid;
```

AdobeStockImage/Model/SetUnlicensedImageMediaGallery.php
```diff
@@ -0,0 +1,77 @@
...
+use Magento\AdobeStockAssetApi\Api\Data\AssetInterface;
+use Magento\Framework\App\ResourceConnection;
+use Magento\Framework\DB\Adapter\AdapterInterface;
+use Magento\Framework\Exception\LocalizedException;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Set unlicensed label for media gallery image
+ */
+class SetUnlicensedImageMediaGallery
```
sivaschenko:
```suggestion
class SetLicensedInMediaGalleryGrid
```

AdobeStockImage/Model/SetUnlicensedImageMediaGallery.php
sivaschenko: Agree 👍

AdobeStockImage/Model/SetUnlicensedImageMediaGallery.php
sivaschenko: That's indexing exception. If an exception occurs here - it's not admin user's fault, so I don't think we should forward exception to user. Loggin and skipping seems ok to me

sivaschenko (changes requested):
Hi @Nazar65 thanks for the pull request! Looks great, the only thing is that "Unlicensed" naming does not really match the contents. Considering that we are working with `is_licensed` property most of the time and only once render it as an "Unlicensed" label. I thinks the "licensed" should be used for naming instead of "unlicensed". Please see my propositions in comments

Nazar65:
Done :heavy_check_mark:

AdobeStockImage/Model/AssetIndexer.php
```diff
@@ -10,8 +10,6 @@
 use Magento\AdobeStockAssetApi\Api\AssetRepositoryInterface;
 use Magento\Framework\Api\SearchCriteriaBuilder;
 use Magento\Framework\App\Filesystem\DirectoryList;
-use Magento\Framework\App\ResourceConnection;
-use Magento\Framework\DB\Adapter\AdapterInterface;
 use Magento\Framework\Filesystem;
 use Magento\Framework\Filesystem\Directory\ReadInterface;
 use Magento\Framework\Filesystem\Driver\File;
```
coderimus: Like an alternative, you can avoid using direct class loading by using the `use Magento\Framework\FileSystem\DriverInterface`. To inject it you can declare a new constructor parameter and set the `Magento\Framework\Filesystem\Driver\File` through the di.xml. See this as an example: https://github.com/magento/adobe-stock-integration/pull/940/files#diff-adb30514bd65a37c94ac4bcf1fccb969R43

Nazar65: Thank you !

sivaschenko (approved):

Nazar65:
@magento run all tests

engcom-Golf:
@magento run all tests

PR #949, opened 2020-02-04
Make UserProfile interface nullable, cover implementation with unit test

hws47a:
### Description (*)
- Created a unit test for `AdobeIms/Model/UserProfile.php`
- Made UserProfileInterface methods nullable which allows to reset a state of an object.


AdobeImsApi/Api/Data/UserProfileInterface.php
```diff
@@ -19,24 +19,24 @@ interface UserProfileInterface extends ExtensibleDataInterface
...
 
     /**
      * Set user ID
      *
-     * @param int $value
+     * @param int|null $value
      * @return void
      */
-    public function setUserId(int $value): void;
+    public function setUserId(?int $value): void;
```
sivaschenko: What is the use case for setting the user id to null? (as well as for other fields)

hws47a: Because every field is nullable, there are should be a way to return fields or the entire object to an initial state.

hws47a: Decided to remove nullable setters for now because it's too debatable.

sivaschenko (approved):

sivaschenko (changes requested):
Thanks for update @hws47a !

Currently the unit tests build is failing with a fatal error:
`PHP Fatal error:  Uncaught Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@testWith" in method Magento\AdobeIms\Test\Unit\Model\UserProfileTest::testAccessToken() was never imported. Did you maybe forget to add a "use" statement for this annotation? in /var/www/html/vendor/doctrine/annotations/lib/Doctrine/Common/Annotations/AnnotationException.php:54`

AdobeIms/Test/Unit/Model/UserProfileTest.php
```diff
@@ -0,0 +1,191 @@
...
+    {
+        $this->objectManager = new ObjectManager($this);
+        $this->model = $this->objectManager->getObject(UserProfile::class);
+    }
+
+    /**
+     * Test setAccessToken
+     *
+     * @param string|null $value
+     * @testWith ["value1"]
```
Nazar65: @hws47a This feature available only for PHPUnit 7.1 Magento uses PHPUnit 6.5.14 https://phpunit.readthedocs.io/en/7.1/annotations.html#testwith

Nazar65 (changes requested):

sivaschenko (dismissed):

sivaschenko (approved):

PR #951, opened 2020-02-07
Added PHPUnit test for AdobeIms/Model/Config

drpayyne:
### Description

Adding PHPUnit tests for `AdobeIms/Model/Config`

### Manual testing scenarios
-  Assert PHPUnit tests pass for `AdobeIms/Model/Config`

### DIFF
--- a/AdobeIms/Model/Config.php
+++ b/AdobeIms/Model/Config.php
@@ -17,13 +17,13 @@ use Magento\Framework\UrlInterface;
  */
 class Config implements ConfigInterface
 {
-    public const XML_PATH_API_KEY = 'adobe_ims/integration/api_key';
-    public const XML_PATH_PRIVATE_KEY = 'adobe_ims/integration/private_key';
-    public const XML_PATH_TOKEN_URL = 'adobe_ims/integration/token_url';
-    public const XML_PATH_AUTH_URL_PATTERN = 'adobe_ims/integration/auth_url_pattern';
-    public const XML_PATH_LOGOUT_URL_PATTERN = 'adobe_ims/integration/logout_url';
-    public const XML_PATH_DEFAULT_PROFILE_IMAGE = 'adobe_ims/integration/default_profile_image';
-    public const XML_PATH_IMAGE_URL_PATTERN = 'adobe_ims/integration/image_url';
+    private const XML_PATH_API_KEY = 'adobe_ims/integration/api_key';
+    private const XML_PATH_PRIVATE_KEY = 'adobe_ims/integration/private_key';
+    private const XML_PATH_TOKEN_URL = 'adobe_ims/integration/token_url';
+    private const XML_PATH_AUTH_URL_PATTERN = 'adobe_ims/integration/auth_url_pattern';
+    private const XML_PATH_LOGOUT_URL_PATTERN = 'adobe_ims/integration/logout_url';
+    private const XML_PATH_DEFAULT_PROFILE_IMAGE = 'adobe_ims/integration/default_profile_image';
+    private const XML_PATH_IMAGE_URL_PATTERN = 'adobe_ims/integration/image_url';
 
     /**
      * @var ScopeConfigInterface
--- a/AdobeIms/Model/Config.php
+++ b/AdobeIms/Model/Config.php
@@ -94,7 +94,7 @@ class Config implements ConfigInterface
     }
 
     /**
-     * Retrieve token URL
+     * Get locale
      *
      * @return string
      */
--- a/AdobeIms/Model/Config.php
+++ b/AdobeIms/Model/Config.php
@@ -3,7 +3,6 @@
  * Copyright © Magento, Inc. All rights reserved.
  * See COPYING.txt for license details.
  */
-
 declare(strict_types=1);
 
 namespace Magento\AdobeIms\Model;
@@ -18,13 +17,13 @@ use Magento\Framework\UrlInterface;
  */
 class Config implements ConfigInterface
 {
-    private const XML_PATH_API_KEY = 'adobe_ims/integration/api_key';
-    private const XML_PATH_PRIVATE_KEY = 'adobe_ims/integration/private_key';
-    private const XML_PATH_TOKEN_URL = 'adobe_ims/integration/token_url';
-    private const XML_PATH_AUTH_URL_PATTERN = 'adobe_ims/integration/auth_url_pattern';
-    private const XML_PATH_LOGOUT_URL_PATTERN = 'adobe_ims/integration/logout_url';
-    private const XML_PATH_DEFAULT_PROFILE_IMAGE = 'adobe_ims/integration/default_profile_image';
-    private const XML_PATH_IMAGE_URL_PATTERN = 'adobe_ims/integration/image_url';
+    public const XML_PATH_API_KEY = 'adobe_ims/integration/api_key';
+    public const XML_PATH_PRIVATE_KEY = 'adobe_ims/integration/private_key';
+    public const XML_PATH_TOKEN_URL = 'adobe_ims/integration/token_url';
+    public const XML_PATH_AUTH_URL_PATTERN = 'adobe_ims/integration/auth_url_pattern';
+    public const XML_PATH_LOGOUT_URL_PATTERN = 'adobe_ims/integration/logout_url';
+    public const XML_PATH_DEFAULT_PROFILE_IMAGE = 'adobe_ims/integration/default_profile_image';
+    public const XML_PATH_IMAGE_URL_PATTERN = 'adobe_ims/integration/image_url';
 
     /**
      * @var ScopeConfigInterface
--- a/AdobeIms/Test/Unit/Model/ConfigTest.php
+++ b/AdobeIms/Test/Unit/Model/ConfigTest.php
@@ -20,6 +20,11 @@ use PHPUnit\Framework\TestCase;
 class ConfigTest extends TestCase
 {
     /**
+     * Constant for string return value
+     */
+    private const STRING_RETURN = 'STRING_RETURN_VALUE';
+
+    /**
      * @var Config
      */
     private $config;
@@ -52,9 +57,10 @@ class ConfigTest extends TestCase
     {
         $this->scopeConfigMock->expects($this->once())
             ->method('getValue')
-            ->with(Config::XML_PATH_API_KEY);
+            ->with(Config::XML_PATH_API_KEY)
+            ->willReturn(self::STRING_RETURN);
 
-        $this->config->getApiKey();
+        $this->assertEquals(self::STRING_RETURN, $this->config->getApiKey());
     }
 
     /**
@@ -64,9 +70,10 @@ class ConfigTest extends TestCase
     {
         $this->scopeConfigMock->expects($this->once())
             ->method('getValue')
-            ->with(Config::XML_PATH_PRIVATE_KEY);
+            ->with(Config::XML_PATH_PRIVATE_KEY)
+            ->willReturn(self::STRING_RETURN);
 
-        $this->config->getPrivateKey();
+        $this->assertEquals(self::STRING_RETURN, $this->config->getPrivateKey());
     }
 
     /**
@@ -77,9 +84,9 @@ class ConfigTest extends TestCase
         $this->scopeConfigMock->expects($this->once())
             ->method('getValue')
             ->with(Config::XML_PATH_TOKEN_URL)
-            ->willReturn('URL');
+            ->willReturn(self::STRING_RETURN);
 
-        $this->config->getTokenUrl();
+        $this->assertEquals(self::STRING_RETURN, $this->config->getTokenUrl());
     }
 
     /**
@@ -93,12 +100,11 @@ class ConfigTest extends TestCase
                 [Config::XML_PATH_API_KEY],
                 [Custom::XML_PATH_GENERAL_LOCALE_CODE],
                 [Config::XML_PATH_AUTH_URL_PATTERN]
-            )
-            ->willReturn('URL');
+            )->willReturn(self::STRING_RETURN);
 
-        $this->urlMock->expects($this->once())->method('getUrl')->willReturn('URL');
+        $this->urlMock->expects($this->once())->method('getUrl')->willReturn(self::STRING_RETURN);
 
-        $this->config->getAuthUrl();
+        $this->assertEquals(self::STRING_RETURN, $this->config->getAuthUrl());
     }
 
     /**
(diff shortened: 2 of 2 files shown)


AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+
+    /**
+     * Test for \Magento\AdobeIms\Model\Config::getTokenUrl
+     */
+    public function testGetTokenUrl(): void
+    {
+        $this->scopeConfigMock->expects($this->once())
+            ->method('getValue')
+            ->with(Config::XML_PATH_TOKEN_URL)
+            ->willReturn('URL');
```
drpayyne: Is it alright to return a string `'URL'` here since we just need it to return a string type? Or is it better to return an actual URL?

AdobeIms/Model/Config.php
```diff
@@ -18,13 +17,13 @@
  */
 class Config implements ConfigInterface
 {
-    private const XML_PATH_API_KEY = 'adobe_ims/integration/api_key';
-    private const XML_PATH_PRIVATE_KEY = 'adobe_ims/integration/private_key';
-    private const XML_PATH_TOKEN_URL = 'adobe_ims/integration/token_url';
-    private const XML_PATH_AUTH_URL_PATTERN = 'adobe_ims/integration/auth_url_pattern';
-    private const XML_PATH_LOGOUT_URL_PATTERN = 'adobe_ims/integration/logout_url';
-    private const XML_PATH_DEFAULT_PROFILE_IMAGE = 'adobe_ims/integration/default_profile_image';
-    private const XML_PATH_IMAGE_URL_PATTERN = 'adobe_ims/integration/image_url';
```
Nazar65: Is this necessary ? to change const ?

drpayyne: I made it public so I can access the same constants from the associated unit test class instead of duplicating those strings.

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+    /**
+     * Test for \Magento\AdobeIms\Model\Config::getPrivateKey
+     */
+    public function testGetPrivateKey(): void
+    {
+        $this->scopeConfigMock->expects($this->once())
+            ->method('getValue')
+            ->with(Config::XML_PATH_PRIVATE_KEY);
+
+        $this->config->getPrivateKey();
```
hws47a: Please check that correct value is returned instead of asserting the number of internal calls

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+     * Test for \Magento\AdobeIms\Model\Config::getTokenUrl
+     */
+    public function testGetTokenUrl(): void
+    {
+        $this->scopeConfigMock->expects($this->once())
+            ->method('getValue')
+            ->with(Config::XML_PATH_TOKEN_URL)
+            ->willReturn('URL');
+
+        $this->config->getTokenUrl();
```
hws47a: Please check that correct value is returned instead of asserting the number of internal calls

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+            ->withConsecutive(
+                [Config::XML_PATH_API_KEY],
+                [Custom::XML_PATH_GENERAL_LOCALE_CODE],
+                [Config::XML_PATH_AUTH_URL_PATTERN]
+            )
+            ->willReturn('URL');
+
+        $this->urlMock->expects($this->once())->method('getUrl')->willReturn('URL');
+
+        $this->config->getAuthUrl();
```
hws47a: Please check that correct value is returned instead of asserting the number of internal calls

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+     * Test for \Magento\AdobeIms\Model\Config::getCallBackUrl
+     */
+    public function testGetCallBackUrl(): void
+    {
+        $this->urlMock->expects($this->once())
+            ->method('getUrl')
+            ->with('adobe_ims/oauth/callback')
+            ->willReturn('URL');
+
+        $this->config->getCallBackUrl();
```
hws47a: Please check that correct value is returned instead of asserting the number of internal calls

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+     * Test for \Magento\AdobeIms\Model\Config::getLogoutUrl
+     */
+    public function testGetLogoutUrl(): void
+    {
+        $this->scopeConfigMock->expects($this->once())
+            ->method('getValue')
+            ->with(Config::XML_PATH_LOGOUT_URL_PATTERN)
+            ->willReturn('URL');
+
+        $this->config->getLogoutUrl('Access token');
```
hws47a: Please check that correct value is returned and what method do with `'Access token'` value

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+    {
+        $this->scopeConfigMock->expects($this->exactly(2))
+            ->method('getValue')
+            ->withConsecutive(
+                [Config::XML_PATH_API_KEY],
+                [Config::XML_PATH_IMAGE_URL_PATTERN]
+            )
+            ->willReturn('URL');
+
+        $this->config->getProfileImageUrl();
```
hws47a: Please check that correct value is returned instead of asserting the amount of internal calls

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+     * Test for \Magento\AdobeIms\Model\Config::getDefaultProfileImage
+     */
+    public function testGetDefaultProfileImage(): void
+    {
+        $this->scopeConfigMock->expects($this->once())
+            ->method('getValue')
+            ->with(Config::XML_PATH_DEFAULT_PROFILE_IMAGE)
+            ->willReturn('URL');
+
+        $this->config->getDefaultProfileImage();
```
hws47a: Please check that correct value is returned instead of asserting the amount of internal calls

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +30,129 @@ class ConfigTest extends TestCase
...
+     * Test for \Magento\AdobeIms\Model\Config::getApiKey
      */
     public function testGetApiKey(): void
     {
         $this->scopeConfigMock->expects($this->once())
             ->method('getValue')
-            ->with(self::CONFIG_XML_PATH_API_KEY);
+            ->with(Config::XML_PATH_API_KEY);
+
         $this->config->getApiKey();
```
hws47a: Please check that correct value is returned instead of asserting the number of internal calls.

It could look like:
```
        $result = 'API KEY';
        $this->scopeConfigMock->method('getValue')
             ->with(Config::XML_PATH_API_KEY)
             ->willReturn($result);

         $this->assertSame($result, $this->config->getApiKey());
```

hws47a (changes requested):

drpayyne:
Hi @hws47a, thanks for reviewing my PR. I've implemented your suggested changes. Thank you!

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +35,129 @@ class ConfigTest extends TestCase
...
+
+        $this->assertEquals(self::STRING_RETURN, $this->config->getApiKey());
+    }
+
+    /**
+     * Test for \Magento\AdobeIms\Model\Config::getPrivateKey
+     */
+    public function testGetPrivateKey(): void
+    {
+        $this->scopeConfigMock->expects($this->once())
```
hws47a: I think it's better not to check for the number of calls, just check the output.
```suggestion
        $this->scopeConfigMock->method('getValue')
             ->with(Config::XML_PATH_PRIVATE_KEY)
             ->willReturn(self::STRING_RETURN);
```
and in all other tests as well

hws47a (changes requested):

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +35,129 @@ class ConfigTest extends TestCase
...
+     */
+    public function testGetAuthUrl(): void
+    {
+        $this->scopeConfigMock->expects($this->exactly(3))
+            ->method('getValue')
+            ->withConsecutive(
+                [Config::XML_PATH_API_KEY],
+                [Custom::XML_PATH_GENERAL_LOCALE_CODE],
+                [Config::XML_PATH_AUTH_URL_PATTERN]
+            )->willReturn(self::STRING_RETURN);
```
hws47a: If you have more than one call they shouldn't return the same value, otherwise, it's hard to check which value exactly was returned. Also, I don't see any of these returned value been used.

AdobeIms/Test/Unit/Model/ConfigTest.php
drpayyne: Sure, I will do that. But why do we not want to check for number of calls as well? It further makes the test solid, right? Or are there any specific situations where call frequency is not recommended?

AdobeIms/Test/Unit/Model/ConfigTest.php
drpayyne: It is being used in the string replace, but since we don't use the actual format, nothing is really replaced and the original string is returned.

drpayyne: Also, all of them should be returning strings, so the string type is effectively what we are checking, right? And not the difference in actual string values?

AdobeIms/Test/Unit/Model/ConfigTest.php
drpayyne: Fixed this, thanks!

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -36,28 +35,121 @@ class ConfigTest extends TestCase
...
         $this->scopeConfigMock = $this->createMock(ScopeConfigInterface::class);
-        $this->config = $this->objectManager->getObject(
-            Config::class,
-            [
-                'scopeConfig' => $this->scopeConfigMock
-            ]
-        );
+        $this->urlMock = $this->createMock(UrlInterface::class);
+
+        $this->config = new Config($this->scopeConfigMock, $this->urlMock);
```
Nazar65, line 98: Why you removed objectManager and ccreate class directly ?

hws47a: The only point of having an object manager helper is to mock some dependencies for you. Especially when you are a complex object with a lot of dependencies. If you mock everything by yourself there is no reason to use it.

Also from technical guidelines:
> 11.2.3. ObjectManagerHelper MAY BE used to automatically mock all dependencies of the object under test.

Which means that it's not required.

AdobeIms/Test/Unit/Model/ConfigTest.php
hws47a: We have to check the logic. In most cases, the logic is that we get some value and return it back. 
For this method, we have to actually check what method does.

This is an example of how it could look like. You can use `@testWith` annotation or `@dataProvider` method to test for more than one set of values:
```php
        $this->scopeConfigMock->method('getValue')
            ->willReturnMap([
                [Config::XML_PATH_API_KEY, 'ClientIdApiKey'],
                [Custom::XML_PATH_GENERAL_LOCALE_CODE, 'EN'],
                [Config::XML_PATH_AUTH_URL_PATTERN, 'https://example.com/authorize?client_id=#{client_id}&redirect_uri=#{redirect_uri}&locale=#{locale}&scope=openid'],
            ]);

        $this->urlMock->method('getUrl')
            ->willReturn('https://backend.com/adobe_ims/oauth/callback');
        $expectedResult = 'https://example.com/authorize?client_id=ClientIdApiKey&redirect_uri=https://backend.com/adobe_ims/oauth/callback&locale=EN&scope=openid';
        $this->assertEquals($expectedResult, $this->config->getAuthUrl());
```

More about return map:
https://phpunit.de/manual/6.5/en/test-doubles.html#test-doubles.stubs.examples.StubTest5.php

hws47a (dismissed):
Thanks for all of improvements

hws47a:
@drpayyne please check travis ci pipeline for static test issues

AdobeIms/Test/Unit/Model/ConfigTest.php
```diff
@@ -8,22 +8,31 @@
...
+    private const LOCALE_CODE = 'en_US';
+    private const PRIVATE_KEY = 'PRIVATE_KEY';
+    private const REDIRECT_URI = 'REDIRECT_URI';
+    private const ACCCESS_TOKEN = 'ACCCESS_TOKEN';
+    private const TOKEN_URL = 'https://token-url.com/integration';
+    private const IMAGE_URL_DEFAULT = 'https://image-url.com/default';
+    private const CALLBACK_URL = 'https://magento-instance.com/adobe_ims/oauth/callback';
+    private const IMAGE_URL_PATTERN = 'https://image-url.com/pattern?api_key=#{api_key}';
+    private const LOGOUT_URL_PATTERN = 'https://logout-url.com/pattern?access_token=#{access_token}&redirect_uri=#{redirect_uri}';
+    private const AUTH_URL_PATTERN = 'https://auth-url.com/pattern?client_id=#{client_id}&redirect_uri=#{redirect_uri}&locale=#{locale}';
```
drpayyne: Hi @hws47a, any suggestions on how to split this string constant into a multiline?

drpayyne:
@hws47a, I've fixed the static test issues as well. Thank you for your review!

Nazar65 (dismissed):

AdobeIms/Model/Config.php
sivaschenko: Please do not change constants access level only for testing purposes, create required private constants in tests instead

sivaschenko (changes requested):
Thanks for the test @drpayyne , please take a look at my review comment:

Please do not change constants access level only for testing purposes, create required private constants in tests instead

AdobeIms/Test/Unit/Model/ConfigTest.php
sivaschenko: @Nazar65 @hws47a @drpayyne ObjectManagerHelper is always preferable for unit tests. It prevents the test from failing if additional dependency is introduced to the class under test and also makes further work with class and test easier.

sivaschenko (approved):

sivaschenko:
@magento run Static Tests

AdobeIms/Test/Unit/Model/ConfigTest.php
Nazar65: agree

hws47a (approved):
@drpayyne Nice work!

Issue #864, opened 2019-12-13
User is not Signed Out from IMS after the User Role was changed


chalov-anton:
### Preconditions (*)
1. Installed Magento with Adobe Stock Integration
2. Configured integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. From Magento Admin go to _System - Permissions - User Roles_, click **Add New Role**
2. Select the following **Role Resources** and **Save Role**
- **_Content_** 
![content](https://user-images.githubusercontent.com/45624059/70789906-9fb89800-1d9c-11ea-80e6-83be7502d78d.png)

- **_Adobe IMS_**
![ims](https://user-images.githubusercontent.com/45624059/70790067-f2924f80-1d9c-11ea-9161-ce5393a58568.png)
3. Go to _System - Permissions - All Users_. **Add New User** and assign the previously created **User Role**. **Save User**
4. Sign In Magento Admin with the previously created User, and go to _Content - Add New Page - Content - Show / Hide Editor - Insert Image..._
5. Click **Search Adobe Stock** and **Sign In** to IMS
6. **Do not Sign Out from IMS** Sign Out from Magento Admin
7. Sign In Magento with Admin Account and change the User Role
Go to  _System - Permissions - User Roles_, select the previously created Role. Go to Role Resources and uncheck _Adobe IMS_
![ims uncheckedpng](https://user-images.githubusercontent.com/45624059/70791190-47cf6080-1d9f-11ea-97ea-923c063313bd.png)
**Save Role**
8. Sign In Magento Admin with the previously created User
9. Go to _Content - Add New Page - Content - Show / Hide Editor - Insert Image...- Search Adobe Stock_ And try to Sign Out from IMS

### Actual result (*)
Sign Out is not completed, a _"POST 403 (Forbidden)"_ error appears in Dev Console.
**User can continue Licensing the images**
![licence continue](https://user-images.githubusercontent.com/45624059/70792092-28d1ce00-1da1-11ea-9f4f-fd0470a99dbd.gif)

### Expected result (*)
User is not Signed In IMS

### Technical details (*)

The plugin/observer should be added to clean admin user tokens stored in `adobe_user_profile` db table when the permissions for users role are updated and _Adobe IMS_ permissions are removed

sivaschenko:
@tkacheva can you please advise what is the expected behaviour in this situation?:

PR #956, opened 2020-02-12
Added observer to role update and updated/created unit tests

gabrieldagama:
### Description (*)
Created an observer to flush the Access and Refresh tokens from the DB when an updated Role had Magento_AdobeIms::logout removed from the resources list.

Also updated the FlushUserTokens class to just flush the tokens when the tokens aren't empty, to avoid unnecessary savings.

Updated and created unit tests for the changes mentioned above.

1. magento/adobe-stock-integration#864: User is not Signed Out from IMS after the User Role was changed

Follow the steps described here: https://github.com/magento/adobe-stock-integration/issues/864

### DIFF
--- a/AdobeIms/Test/Unit/Observer/FlushUsersTokensObserverTest.php
+++ b/AdobeIms/Test/Unit/Observer/FlushUsersTokensObserverTest.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+namespace Magento\AdobeIms\Test\Unit\Observer;
+
+use Magento\Authorization\Model\Role;
+use Magento\Framework\App\RequestInterface;
+
+/**
+ * Flush users tokens observer tests
+ */
+class FlushUsersTokensObserverTest extends \PHPUnit\Framework\TestCase
+{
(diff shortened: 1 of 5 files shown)


sivaschenko (approved):

AdobeIms/Test/Unit/Observer/FlushUsersTokensObserverTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Test flush tokens observer
+     */
+    public function testFlushUsersTokensObserver()
+    {
+        /** @var \Magento\Framework\Event\Observer|\PHPUnit_Framework_MockObject_MockObject $eventObserverMock */
+        $eventObserverMock = $this->createMock(\Magento\Framework\Event\Observer::class);
+        $eventObserverMock->expects($this->exactly(2))->method("getDataByKey")
+            ->will($this->returnCallback(function ($args) {
```
sivaschenko: I'd suggest using `returnValueMap` instead of `returnCallback `

sivaschenko (changes requested):
Thanks for the pull request @gabrieldagama ! Looks great!

The only thing is the commits are not linked to your github account and see my comment for the test

sivaschenko (approved):

PR #987, opened 2020-02-26
DI configuration for the content type option provider

konarshankar07:
### Description (*)

This PR is introducing the DI configuration for the content type option provider for the `Magento\Framework\Data\OptionSourceInterface`.


sivaschenko (changes requested):
Hi @konarshankar07 thanks for the pull request! As we discussed on the public meeting it's possible to use a class configuration similar to approach implemented for `\Magento\AdobeStockClient\Model\SearchParametersProvider\SimpleFilters`. Virtual type is not necessary in this case

konarshankar07:
Hello @sivaschenko ...
Yes, I understood that but I still have a confusion but may be I'm wrong. If we use the real class then we still need to create separate class for each filter and repeating same code for each filter is not making sense to me. Correct me if I'm wrong
Thanks

sivaschenko:
@konarshankar07 you will be able to extend arguments di.xml configuration for the real class the same way as for virtual class

AdobeStockImageAdminUi/Ui/Component/Listing/Columns/ContentType/Options.php
```diff
@@ -11,26 +11,36 @@
...
+     */
+    public function __construct(array $options = [])
+    {
+        $this->_options = $options;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function toOptionArray()
```
sivaschenko:
```suggestion
    public function toOptionArray(): array
```

sivaschenko (dismissed):

AdobeStockImageAdminUi/Ui/Component/Listing/Columns/ContentType/Options.php
```diff
@@ -11,26 +11,36 @@
...
  */
 class Options implements OptionSourceInterface
 {
     /**
-     * Get options
-     *
-     * @return array
+     * @var array
+     */
+    protected $_options = [];
```
sivaschenko:
```suggestion
    private $_options = [];
```

sivaschenko (approved):

Issue #1004, opened 2020-03-18
Save media asset relations when content is saved


sivaschenko:
### Story

https://github.com/magento/adobe-stock-integration/issues/717 [Story 43] User sees pages the image is used in Magento

### Description

In order to save information about media asset used in the content (field of an entity) observers should be introduced for the following events:
 - catalog_product_save_after
 - catalog_category_save_after
 - cms_page_save_after
 - cms_block_save_after 

The observer for each event should parse the content of predefined fields (provided to the observer from di configuration) in order to find the paths within the media directory.

 - If the path within the media directory is found and it matches the media asset - the association between the content and asset should be saved.
 - If the association for the piece of content was saved before but the path of the media asset is not found in the field content - the association should be removed

Observers should use the utility class (generic for all entities) that will contain the common logic. This utility class should be provided with regex pattern from DI configuration

PR #1016, opened 2020-03-19
FEATURE: Save media asset relations when content is saved 

coderimus:
### Description (*)
This PR provides implementation fro the #1004 Currently work in progress.

#1004 

N/A


MediaContent/Model/ContentProcessor.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /**
+     * @param string $contentEntityId
+     * @param array $contentData
+     * @param string $contentType
+     *
+     * @throws IntegrationException
+     */
+    public function execute(string $contentEntityId, array $contentData, string $contentType): void
```
sivaschenko: This function is too complicated, please extract a couple of private functions to make it more readable

MediaContent/Model/ContentProcessor.php
```diff
@@ -0,0 +1,111 @@
...
+    ) {
+        $this->extractAssetFromContent = $extractAssetFromContent;
+        $this->assignAsset = $assignAsset;
+        $this->getAssetsUsedInContent = $getAssetsUsedInContent;
+        $this->unassignAsset = $unassignAsset;
+        $this->logger = $logger;
+    }
+
+    /**
+     * @param string $contentEntityId
```
sivaschenko: please add method description to all introduced methods

MediaContent/Observer/CatalogCategorySaveAfter.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @throws IntegrationException
+     */
+    public function execute(Observer $observer): void
+    {
+        $content = [];
+        /** @var Category $category */
+        $category = $observer->getEvent()->getData('category');
+        if ($category->dataHasChangedFor(self::IMAGE_FIELD)) {
+            $content[self::IMAGE_FIELD] = $category->getData(self::IMAGE_FIELD);
```
sivaschenko: Fields of interest should be declared in DI configuration and passed to the constructor.

MediaContent/Model/ExtractAssetFromContent.php
```diff
@@ -0,0 +1,86 @@
...
+use Magento\Framework\Exception\IntegrationException;
+use Magento\MediaContentApi\Api\ExtractAssetFromContentInterface;
+use Magento\MediaGalleryApi\Api\Data\AssetInterface;
+use Magento\MediaGalleryApi\Model\Asset\Command\GetByPathInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Used for extracting media asset list from a media content by the search pattern.
+ */
+class ExtractAssetFromContent implements ExtractAssetFromContentInterface
```
sivaschenko:
```suggestion
class GetAssetsUsedInContent implements ExtractAssetFromContentInterface
```

MediaGalleryUi/etc/db_schema_whitelist.json
```diff
@@ -24,4 +24,4 @@
             "MEDIA_GALLERY_ASSET_GRID_ID_MEDIA_GALLERY_ASSET_ID": true
         }
     }
-}
+}
```
sivaschenko: I'd keep that space at the end of the file as it's readded by PHP with default settings automatically

MediaContent/Observer/CatalogProductSaveAfter.php
```diff
@@ -0,0 +1,62 @@
...
+use Magento\Catalog\Api\Data\ProductInterface;
+use Magento\Framework\Event\Observer;
+use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Exception\IntegrationException;
+use Magento\MediaContent\Model\ContentProcessor;
+
+/**
+ * Observe the catalog_product_save_after event and run processing relation between media content and media asset
+ */
+class CatalogProductSaveAfter implements ObserverInterface
```
sivaschenko: Please avoid using the event name in the name of the observer class name, it can potentially be used to observe different event
```suggestion
class CatalogProduct implements ObserverInterface
```

MediaContentApi/Api/ExtractAssetFromContentInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ * Extract media asset from a media content
+ */
+interface ExtractAssetFromContentInterface
+{
+    /**
+     * @param string $contentData
+     *
+     * @return array
+     */
+    public function execute(string $contentData): array;
```
sivaschenko:
```suggestion
    public function execute(string $content): array;
```

MediaContent/composer.json
```diff
@@ -5,7 +5,9 @@
         "php": "~7.1.3||~7.2.0||~7.3.0",
         "magento/framework": "*",
         "magento/module-media-content-api": "*",
-        "magento/module-adobe-stock-asset-api": "*"
+        "magento/module-media-gallery-api": "*",
+        "magento/module-media-gallery": "*",
+        "magento/module-cms": "*"
```
sivaschenko: cms and catalog should be soft dependencies for the implementation - please add them to suggest section of composer.json

sivaschenko (changes requested):
Thanks for the PR @coderimus , please see my review comments

MediaContent/Model/ExtractAssetFromContent.php
coderimus: `GetAssetsUsedInContent` already exists and it gets data from the data storage. Should we change this?

MediaGalleryUi/etc/db_schema_whitelist.json
coderimus: This change added automatically by the CLI Magento command.

MediaContent/Model/ContentProcessor.php
coderimus: I think that adding description, for example, to the observer with the one execute method will duplicate class description which already has a meaningful description of the class action. For example, 
```text
/**
 * Observe the catalog_category_save_after event and run processing relation between category content and media asset
 */
```
is a description of action provided by the class. If we will have a method description they both will be identical. Maybe we can remove doc block from the class and use only method description in case when there is a straight forward logic with one method?

As an exception, I can say that ContentProcessor requires adding a description to the method.

coderimus:
@sivaschenko 
I implemented the majority of requested changes. One thing I didn't change - `ExtractAssetFromContent` Please, see my comment to your review.

I want to add some notes about `ContentProcessor` logic:
1. at the `addNewRelation` method the class creates new relation if there are no relation existed.
2. after it I use continue to skip the next step because no update is required if we add new relation.
3. at the `updateRelation` method as a first step we check whether the content is null or not. This condition added because it is possible to have content equals to null if it was completely cleared.
4. If the content is not null we try to get assets from it. If we find something we compare this with the current relation. If something new provided we insert it. 
5. And as the last step if we compare relation with the asset list from the content: if a relation has an asset which is not in the content - remove relation entry.

MediaContent/etc/di.xml
```diff
@@ -10,4 +10,43 @@
...
+        <arguments>
+            <argument name="searchPattern" xsi:type="array">
+                <item name="media_gallery" xsi:type="string">({{media url="(.*?)"}})</item>
+                <item name="catalog_image" xsi:type="string">(media\/(.*))</item>
+            </argument>
+        </arguments>
+    </type>
+    <type name="Magento\MediaContent\Observer\CatalogCategory">
+        <arguments>
+            <argument name="contentField" xsi:type="array">
```
sivaschenko: Let's add s to the end of the array variable names for better readability
```suggestion
            <argument name="contentFields" xsi:type="array">
```

MediaContent/etc/di.xml
```diff
@@ -10,4 +10,43 @@
     <preference for="Magento\MediaContentApi\Api\UnassignAssetInterface" type="Magento\MediaContent\Model\UnassignAsset"/>
     <preference for="Magento\MediaContentApi\Api\GetAssetsUsedInContentInterface" type="Magento\MediaContent\Model\GetAssetsUsedInContent"/>
     <preference for="Magento\MediaContentApi\Api\GetContentWithAssetInterface" type="Magento\MediaContent\Model\GetContentWithAsset"/>
+    <preference for="Magento\MediaContentApi\Api\ExtractAssetFromContentInterface" type="Magento\MediaContent\Model\ExtractAssetFromContent"/>
+    <type name="Magento\MediaContentApi\Api\ExtractAssetFromContentInterface">
+        <arguments>
+            <argument name="searchPattern" xsi:type="array">
```
sivaschenko:
```suggestion
            <argument name="searchPatterns" xsi:type="array">
```

MediaContentApi/Api/ExtractAssetFromContentInterface.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Magento\MediaContentApi\Api;
+
+/**
+ * Extract media asset from a media content
+ */
+interface ExtractAssetFromContentInterface
+{
+    /**
+     * @param string $content
```
sivaschenko: Please add method description

MediaContent/Observer/CmsPage.php
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @throws IntegrationException
+     */
+    public function execute(Observer $observer): void
+    {
+        $content = [];
+        /** @var Page $cmsPage */
+        $cmsPage = $observer->getEvent()->getObject();
+        $cmsPageData = $cmsPage->getData();
+        foreach ($this->contentField as $key => $field) {
```
sivaschenko: Key is not necessary
```suggestion
        foreach ($this->contentField as $field) {
```

MediaContent/Observer/CmsPage.php
```diff
@@ -0,0 +1,68 @@
...
+            if ($cmsPage->dataHasChangedFor($field)) {
+                $content[$field] = $cmsPageData[$field];
```
sivaschenko: I think retrieving all the data is unnecessary
```suggestion
                $content[$field] = $cmsPage->getData($field);
```

MediaContent/Observer/CatalogProduct.php
```diff
@@ -0,0 +1,68 @@
...
+        foreach ($this->contentField as $key => $field) {
+            if ($product->dataHasChangedFor($field)) {
+                $content[$field] = $productData[$field];
+            }
+        }
+
+        if (!empty($content)) {
+            $this->contentProcessor->execute((string)$product->getId(), $content, self::CONTENT_TYPE);
+        }
+    }
```
sivaschenko: Looks like this logic is the same for all observers, should we move it to the contentProcessor?
```suggestion
    public function execute(Observer $observer): void
    {
        /** @var ProductInterface $product */
        $product = $observer->getEvent()->getData('product');
        $productData = $product->getData();
        $this->contentProcessor->execute(self::CONTENT_TYPE, (string)$product->getId(), $productData, $this->fields);
    }
```

MediaContent/Model/ContentProcessor.php
```diff
@@ -0,0 +1,162 @@
...
+     * @param string $contentType
+     *
+     * @throws IntegrationException
+     */
+    public function execute(string $contentEntityId, array $contentData, string $contentType): void
+    {
+        try {
+            foreach ($contentData as $contentField => $content) {
+                $relations = $this->getAssetsUsedInContent->execute($contentType, $contentEntityId, $contentField);
+                if (empty($relations)) {
```
sivaschenko: I don't think the system should be a separate logic branch for empty relations - updateRelation method should cover this case

sivaschenko (changes requested):
Thanks for updates @coderimus ! Please see my core review comments

sivaschenko (dismissed):

coderimus:
@sivaschenko thank you for your test results review. I made additional fix for the catalog save. Please, review it. As I see the majority of tests passed. Other looks like skipped or temporarily wrong because of settings.

coderimus:
@sivaschenko I have a question: do we really need to have all string types stored in the `media_content_asset` null allowed? I do not see cases when we can have a type, filed, entity_id be null by default. Or I missed something?

coderimus:
@sivaschenko I made changes related to the null value for the `media_content_asset` fields. Please, review PR.

Issue #1014, opened 2020-03-19
Do not show excluded directories in the media gallery directory tree


sivaschenko:
### Preconditions (*)
Install magneto and adobe stock integration

### Steps to reproduce (*)
1. Login to Admin panel
2. Go to Catalog -> Category
3. Select any category
4. Expand Content fieldset
5. Click "Select from Gallery" button next to "Category Image" field
6. Observer the media gallery directory tree

### Expected Result

Excluded directories like captcha, downloadable, catalog/product etc. should not be displayed in the directory tree, the files from the excluded directories should not be present in displayed media assets

### Technical details

Excluded directories can be retrieved from `Magento\Cms\Model\Wysiwyg\Images\Storage` 
class and are based on the following DI configuration https://github.com/magento/magento2/blob/bf03292db7b1f1cd642bc754696e0dbdb3a448f6/app/code/Magento/Cms/etc/di.xml#L57-L90

PR #1025, opened 2020-03-23
Do not show excluded directories and images in the media gallery

Nazar65:

1. magento/adobe-stock-integration#1014: Do not show excluded directories and images in the media gallery
2. ...

Install magneto and adobe stock integration

1. 
2.     Login to Admin panel
3.     Go to Catalog -> Category
4.     Select any category
5.     Expand Content fieldset
6.     Click "Select from Gallery" button next to "Category Image" field
7.     Observer the media gallery directory tree

Excluded directories like captcha, downloadable, catalog/product etc. should not be displayed in the directory tree, the files from the excluded directories should not be present in displayed media assets

### DIFF
--- a/MediaGalleryUi/Model/Directories/ExcludedDirectories.php
+++ b/MediaGalleryUi/Model/Directories/ExcludedDirectories.php
@@ -8,35 +8,33 @@ declare(strict_types=1);
 namespace Magento\MediaGalleryUi\Model\Directories;
 
 /**
- * Excluded directories validator
+ * Directory paths that should not be included in the media gallery
  */
 class ExcludedDirectories
 {
     /**
      * @var array
      */
-    private $excludedDirectories;
+    private $patterns;
 
     /**
-     * Constructor
-     *
-     * @param array $excludedDirectories
+     * @param array $patterns
      */
     public function __construct(
-        array $excludedDirectories
+        array $patterns
     ) {
-        $this->excludedDirectories = $excludedDirectories;
+        $this->patterns = $patterns;
     }
 
     /**
-     * Validate if directory might be excluded
+     * Check if the path is excluded from displaying in the media gallery
      *
      * @param string $path
      * @return bool
      */
-    public function validate(string $path): bool
+    public function isExcluded(string $path): bool
     {
-        foreach ($this->excludedDirectories as $pattern) {
+        foreach ($this->patterns as $pattern) {
             preg_match($pattern, $path, $result);
 
             if ($result) {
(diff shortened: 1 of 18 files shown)


MediaGalleryUi/etc/di.xml
```diff
@@ -15,6 +15,20 @@
             </argument>
         </arguments>
     </type>
+     <type name="Magento\MediaGalleryUi\Model\FilesIndexer">
+         <arguments>
+	      <argument name="excludedDirectories" xsi:type="array">
```
sivaschenko: I think it might make sense to introduce a single class providing the excluded directories, configure it via DI and use this single class in both FilesIndexer and FolderTree

sivaschenko (changes requested):
Thanks for the pull request @Nazar65 ! Please take a look at the review comment

MediaGalleryUi/etc/di.xml
Nazar65: ok

Nazar65: @sivaschenko done

MediaGalleryUi/Model/Directories/ExcludedDirectories.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param array $excludedDirectories
+     */
+    public function __construct(
+        array $excludedDirectories
+    ) {
+        $this->excludedDirectories = $excludedDirectories;
+    }
+
+    /**
+     * Validate if directory might be excluded
```
sivaschenko:
```suggestion
     * Is path excluded from displaying in the media gallery
```

MediaGalleryUi/Model/Directories/ExcludedDirectories.php
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param string $path
+     * @return bool
+     */
+    public function validate(string $path): bool
```
sivaschenko: The result of this function is opposite to expected validation, in my opinion. Let's have more specific method name to avoid confusion
```suggestion
    public function isExcluded(string $path): bool
```

MediaGalleryUi/Model/Directories/ExcludedDirectories.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    private $excludedDirectories;
+
+    /**
+     * Constructor
+     *
+     * @param array $excludedDirectories
+     */
+    public function __construct(
+        array $excludedDirectories
```
sivaschenko: I think this name will better describe the content of the variable and property
```suggestion
        array $patterns
```

sivaschenko (changes requested):
Thanks for updates @Nazar65 ! there is a couple more things that can be improved

sivaschenko (approved):

PR #1087, opened 2020-03-31
1083: fix adobe stock asset save action

### LINKED ISSUE #1083: Adobe Stock Integration: not possible to save image.
### Preconditions (*)
Magento 2.4-develop
Adobe-Stock-Integration 1.1-develop
Internet Explorer 11 or Microsoft Edge 44...

1. Installed Magento with Adobe Stock Integration
3. Disable Enhanced Media Gallery in Stores -> Configuration -> Advanced -> System 
3. Enable Adobe Stock Integration in Stores -> Configuration -> Advanced -> System 

### Steps to reproduce (*)
1. Open a media grid
2. Press Search Adobe Stock
3. Select image
4. Press Save preview
5. An error occurred and shown at the selected image above the Save preview button

### Expected result (*)
Image successfully saved to the adobe_stock_asset, media_gallery_asset_grid

### Actual result (*)
An error occurred, image not saved in the data storage: adobe_stock_asset, media_gallery_asset_grid

coderimus:
### Description (*)
Fix for the save image preview or licensed image related to the #1083 
The problem was in the size parameter which is not a media asset part but required for the media gallery grid table.

#1083 

Use steps to reproduce from the #1083


sivaschenko (changes requested):
Thanks for the pull request @coderimus ! 

I think it would be better to update 
 - `\Magento\MediaGalleryUi\Model\AssetIndexer`
- `\Magento\AdobeStockImage\Model\SaveImage::execute`
to save the size in the Media Asset entity in the first place (rather than only setting the size for the grid).

What do you think?

coderimus:
@sivaschenko I tried to change the `Magento\AdobeStockImage\Model\SaveImage::execute` at the beginning but faced with next:` \Magento\AdobeStockImage\Model\Extract\MediaGalleryAsset` converts search result to the `\Magento\MediaGalleryApi\Api\Data\AssetInterface` and exclude size because it is not a getter of a MediaGalleryAsset :( and we do not have size as a storage column.

If we would add the size parameter as a part of a MediaGallery asset everything will be ok and easy to automate. From the other point of view, this looks not required. We can encapsulate knowledge about the size in the scope of the media_gallery_grid.

So we have 2 ways, I think:
1. Update MediaGallery asset with the new property `size`
2. Use PR's provided implementation to manage size information at storing it to the `media_gallery_grid`

sivaschenko:
@coderimus please update your Magento codebase:
https://github.com/magento/magento2/blob/89bc0786ca0d5897362e5d30d0189aee0a6606d7/app/code/Magento/MediaGalleryApi/Api/Data/AssetInterface.php#L73

coderimus:
@sivaschenko I implemented requested changes and adjust unit tests. Functionality works.

I have a suggestion: maybe we can exclude this part from the SaveImage class https://github.com/magento/adobe-stock-integration/blob/e3628a866d4e3570bbaea3f2fc44ad97afc0d1f0/AdobeStockImage/Model/SaveImage.php#L143-L169

and put it into a new SaveImageFile class where we can process all save file-related functionality and return MediaGallery asset? This should fix the code mess static test: https://public-results-storage-prod.magento-testing-service.engineering/reports/magento/adobe-stock-integration/pull/1087/bbbcbd733449dbca99dd5314b49e423a/Statics/allure-report-b2b/index.html#suites/51d3ab4d5d77a0bc6a7ea344e9e7e6be/4afb21c8e4b4da9/

coderimus:
@sivaschenko  I think there is no other way at the first look and I implemented SaveImage class refactoring. Decouple save an Adobe Stock image and file processes. Unit tests provided. Please, review implementation.

AdobeStockImage/Model/SaveImageFile.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param string $destinationPath
+     *
+     * @return int
+     * @throws CouldNotSaveException
+     * @throws AlreadyExistsException
+     * @throws CouldNotDeleteException
+     * @throws FileSystemException
+     * @throws NoSuchEntityException
+     */
+    public function execute(Document $document, string $url, string $destinationPath): int
```
sivaschenko: Let's bread down this into two classes: `SaveImageFile` (lines 96-103) and `SaveMediaGalleryAsset` (lines 105-124)

AdobeStockImage/Model/SaveImageFile.php
```diff
@@ -0,0 +1,126 @@
...
+        /* If the asset has been already saved, delete the previous version */
+        if (null !== $pathAttribute && $pathValue) {
+            $this->storageDelete->execute($pathValue);
+        }
+
+        $path = $this->storageSave->execute($url, $destinationPath);
+
+        $mediaDirectory = $this->fileSystem->getDirectoryRead(DirectoryList::MEDIA);
+        $absolutePath = $mediaDirectory->getAbsolutePath($path);
+        $fileSize = $mediaDirectory->stat($absolutePath)['size'];
```
sivaschenko: Please extract the size retrieval to a private function

AdobeStockImage/Model/SaveImage.php
```diff
@@ -7,64 +7,28 @@
...
 use Magento\Framework\Exception\NoSuchEntityException;
-use Magento\MediaGalleryApi\Model\Asset\Command\GetByIdInterface;
-use Magento\MediaGalleryApi\Model\Asset\Command\SaveInterface;
 use Magento\MediaGalleryApi\Model\Keyword\Command\SaveAssetKeywordsInterface;
 
 /**
- * Test saving Adobe Stock Image to filesystem and database
+ * Save an image provided with the adobe Stock integration.
  */
 class SaveImage implements SaveImageInterface
```
sivaschenko, line 27: I like the refactoring!

sivaschenko (changes requested):
Hi @coderimus thanks for the pull request! Please see my review comments

coderimus:
@sivaschenko the final change made in the scope of the PR main goal and general improvements. Please, review.

AdobeStockImage/Model/SaveImageFile.php
```diff
@@ -0,0 +1,75 @@
...
+        $pathAttribute = $document->getCustomAttribute('path');
+        $pathValue = $pathAttribute->getValue();
+        /* If the asset has been already saved, delete the previous version */
+        if (null !== $pathAttribute && $pathValue) {
+            $this->storageDelete->execute($pathValue);
+        }
+
+        $path = $this->storageSave->execute($url, $destinationPath);
+
+        return $path;
```
lenaorobei: `SaveImage` sounds like a command and should not return anything.

coderimus: @lenaorobei What about next change `SaveImageFileProcessor`? If ok, I will also change `SaveMediaGalleryAsset` to `SaveMediaGalleryAssetProcessor`.

lenaorobei: As per CQRS principle method should do whether do something or return something. In case of `save` it should return `void`.

lenaorobei: As per CQRS principle method should do whether do something or return something. In case of save it should return void.

AdobeStockImage/Model/SaveImage.php
```diff
@@ -86,83 +51,64 @@ class SaveImage implements SaveImageInterface
...
      * @param string $destinationPath
      *
-     * @throws CouldNotSaveException
      * @throws AlreadyExistsException
      * @throws CouldNotDeleteException
+     * @throws CouldNotSaveException
      * @throws FileSystemException
      * @throws NoSuchEntityException
      */
     public function execute(Document $document, string $url, string $destinationPath): void
```
lenaorobei, line 126: I believe this method should  throw `CouldNotSaveException` exception only.

coderimus:
@lenaorobei thank you for your review. I added some changes to the points you mentioned. Please, review when you have time.

coderimus:
@sivaschenko I made adjustments according to review. Please, review.

AdobeStockImage/Model/GetSavedImageFilePath.php
```diff
@@ -0,0 +1,71 @@
...
+use Magento\AdobeStockImage\Model\Storage\Save as StorageSave;
+use Magento\Framework\Api\Search\Document;
+use Magento\Framework\Exception\AlreadyExistsException;
+use Magento\Framework\Exception\CouldNotDeleteException;
+use Magento\Framework\Exception\FileSystemException;
+
+/**
+ * Save asset file and retrieve its path.
+ */
+class GetSavedImageFilePath implements GetSavedImageFilePathInterface
```
sivaschenko: The `Get` prefix in the name of this service is misleading. It actually deletes/saves the file. Let's change it to `SaveImageFile`

AdobeStockImage/Model/GetSavedImageFilePath.php
```diff
@@ -0,0 +1,71 @@
...
+     * Save asset file to filesystem and return its path.
+     *
+     * @param Document $document
+     * @param string $url
+     * @param string $destinationPath
+     *
+     * @return string
+     * @throws AlreadyExistsException
+     * @throws CouldNotDeleteException
+     * @throws FileSystemException
```
sivaschenko: Let's throw only Could not save exception from this service for consistency

AdobeStockImage/Model/GetSavedImageFilePathInterface.php
```diff
@@ -0,0 +1,37 @@
...
+     * @param Document $document
+     * @param string $url
+     * @param string $destinationPath
+     *
+     * @return string
+     * @throws CouldNotSaveException
+     * @throws AlreadyExistsException
+     * @throws CouldNotDeleteException
+     * @throws FileSystemException
+     * @throws NoSuchEntityException
```
sivaschenko: phpdoc does not match with implementation, I'd use @inheritdoc in the implementation to avoid such discrepancy

AdobeStockImage/Model/GetSavedMediaGalleryAssetId.php
```diff
@@ -0,0 +1,113 @@
...
+use Magento\Framework\Exception\NoSuchEntityException;
+use Magento\Framework\Filesystem;
+use Magento\AdobeStockAssetApi\Api\Data\AssetInterface;
+use Magento\MediaGalleryApi\Api\Data\AssetInterface as MediaGalleryAssetInterface;
+use Magento\MediaGalleryApi\Model\Asset\Command\SaveInterface;
+
+/**
+ * Process save action of the MediaGalleryAsset.
+ */
+class GetSavedMediaGalleryAssetId implements GetSavedMediaGalleryAssetIdInterface
```
sivaschenko:
```suggestion
class SaveMediaGalleryAsset implements GetSavedMediaGalleryAssetIdInterface
```

AdobeStockImage/Model/GetSavedMediaGalleryAssetId.php
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Process saving MediaGalleryAsset based on the search document and destination path.
+     *
+     * @param Document $document
+     * @param string $destinationPath
+     *
+     * @return int
+     * @throws CouldNotSaveException
+     * @throws NoSuchEntityException
```
sivaschenko: Let's throw only `CouldNotSaveException` (if the asset could not be loaded after we tried to save it - it's the result of the fact that we were not able to save the asset in the first place)

AdobeStockImage/Model/SaveImage.php
```diff
@@ -7,64 +7,26 @@
...
-use Magento\Framework\Exception\NoSuchEntityException;
-use Magento\MediaGalleryApi\Model\Asset\Command\GetByIdInterface;
-use Magento\MediaGalleryApi\Model\Asset\Command\SaveInterface;
 use Magento\MediaGalleryApi\Model\Keyword\Command\SaveAssetKeywordsInterface;
+use Psr\Log\LoggerInterface;
 
 /**
- * Test saving Adobe Stock Image to filesystem and database
+ * Save an image provided with the adobe Stock integration.
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
```
sivaschenko, line 24: I think SuppressWarnings can be removed as the coupling was reduced by reducing exception types variety

sivaschenko (changes requested):
Hi @coderimus thanks for the updates, please take a look at my review comments

AdobeStockImage/Model/GetSavedImageFilePath.php
coderimus: @sivaschenko I changed the `SaveImageFile` name to this one based on the discussion with @lenaorobei 
image

coderimus: My opinion that `SaveImageFile` good but also it returns the path of a file. As a variant, I used `RetrievePathFromImageSave`. What do you think about this?

sivaschenko: @lenaorobei I'd keep the return of the `SaveImageFile` in order to prevent discrepancies in case that path is adjusted by the storage. There is no way to retrieve the correct file path later in this case.

Please let me know what you think

sivaschenko: @coderimus Olena meant that the service should either save the file or provide the file path. This particular service is used to save the file, so the naming should not be changed. The question is whether to return the $path

coderimus: @sivaschenko thank you for details. I will check this. Also, as I understood, we need to return the name of the SaveMediaGalleryAsset, right? But in that case, we have to return id of the asset.

sivaschenko: If the service can be used each time I need to retrieve the path - "RetrievePath..." naming should be ok.

If the service is primarily used to save the file - it should be named accordingly.

`SaveImageFile` sounds good to me

sivaschenko: @coderimus after conversation with @lenaorobei we come to the next conclusions:
- The class name should be `SaveImageFile`
- The execute method should not return the path (use the argument destinationPath in clients instead) - return type `void`
- Update `\Magento\AdobeStockImage\Model\Storage\Save::execute` to make destinationPath a required parameter and chance the return type to `void`

AdobeStockImage/Model/SaveImage.php
coderimus: Unfortunately, reducing a number of exceptions didn't help so I didn't remove SuppressWarnings.

AdobeStockImage/Model/Storage/Save.php
```diff
@@ -52,7 +52,7 @@ public function __construct(
      * @throws AlreadyExistsException
      * @throws FileSystemException
      */
-    public function execute(string $imageUrl, string $destinationPath = '') : string
+    public function execute(string $imageUrl, string $destinationPath) : string
```
sivaschenko:
```suggestion
    public function execute(string $imageUrl, string $destinationPath): void
```

sivaschenko (changes requested):
Thanks for the updates @coderimus ! One final adjustment is required, please see my comment

AdobeStockImage/Model/Storage/Save.php
coderimus: @sivaschenko changes done :) Thank you.

sivaschenko (dismissed):

AdobeStockImage/Model/SaveMediaGalleryAssetInterface.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Process saving MediaGalleryAsset based on the search document and destination path.
+     *
+     * @param Document $document
+     * @param string $destinationPath
+     *
+     * @return int
+     * @throws CouldNotSaveException
+     */
+    public function execute(Document $document, string $destinationPath): int;
```
lenaorobei: Why the return type is `int`?

coderimus: We return type int to process the next step of saving the Adobe Stock asset. We need it to save asset keywords and the adobe stock asset itself.

lenaorobei: I believe we need to return `void`. The same like we did for `SaveImageFileInterface`.

coderimus: @lenaorobei @sivaschenko I am going to provide changes in a hour.

coderimus: @lenaorobei @sivaschenko I provided requested changes. Now `SaveImageFileInterface` also returns `void`. I used directly a command which gets media asset by a path.

coderimus:
@magento run all tests

lenaorobei:
@coderimus could you please address PHPMD failure?

AdobeStockImage/Model/SaveImage.php
```diff
@@ -88,6 +95,7 @@ public function __construct(
         DocumentToAsset $documentToAsset,
         SaveAssetKeywordsInterface $saveAssetKeywords,
         DocumentToKeywords $documentToKeywords,
+        SetLicensedInMediaGalleryGrid $setLicensedInMediaGalleryGrid,
```
lenaorobei, line 98: This dependency will be removed in https://github.com/magento/adobe-stock-integration/issues/1146

lenaorobei:
@magento run Unit Tests

lenaorobei:
@magento run Functional Tests CE, Functional Tests B2B

coderimus:
@lenaorobei @sivaschenko 
I checked failed MFTF test for the Adobe Stock Integration and find the reason why they failed. Currently, the Adobe stock integration returns the first image which doesn't have `More from this series` 
image
This leads to several issues: assertion failed when 4 images expected and click event also failed when it is looking for the "Show more" button. As a suggestion I provided the next adjustments: search images by the 'car' word :) https://github.com/magento/adobe-stock-integration/pull/1087/commits/2f48c34d84c0cddd36206d81400702bfb7b6b287#diff-83b62ecfd324289923a7b83edf6e8d0eR30-R32

lenaorobei:
@magento run Functional Tests B2B

lenaorobei:
@magento run Functional Tests B2B

AdobeStockImage/Test/Unit/Model/SaveMediaGalleryAssetTest.php
```diff
@@ -0,0 +1,195 @@
...
+            SaveMediaGalleryAsset::class,
+            [
+                'saveMediaAsset' =>  $this->saveMediaAssetMock,
+                'documentToMediaGalleryAsset' =>  $this->documentToMediaGalleryAssetMock,
+                'fileSystem' => $this->fileSystemMock
+            ]
+        );
+        $reflection = new \ReflectionClass(get_class($this->saveMediaAsset));
+        $reflectionMethod = $reflection->getMethod('calculateFileSize');
+        $reflectionMethod->setAccessible(true);
```
orlangur, line 81: As per conversation with @coderimus looked into this test implementation: it seems that refection was not used after all, but even if we would use it there is no need to call this method directly.

Please just call `execute` method in test as you already do and assert second parameter of `convert` call of media asset mock.

Issue #1084, opened 2020-03-31
Exclude directories starting with dot and their contents from media gallery


sivaschenko:
### Preconditions (*)
Install Magento 2.4-x with Adobe Stock Integration 1.1-x

### Steps to reproduce (*)
1. Login to Admin Palel
2. Open Media Gallery i.e.:
   - From Admin go to **Content** - **Pages**, click **Add New Page**)
   - Expand **Content**,click **Show / Hide Editor**, click **Insert Image...**

### Technical Details

`\Magento\MediaGalleryUi\Model\Directories\ExcludedDirectories` di.xml configuration should be updated to eliminate directories with names starting with dot and images from them from media gallery:

### Current Media Gallery folders (*)

![image](https://user-images.githubusercontent.com/2028541/78028619-35c9fb80-7357-11ea-94e8-8501ff60cda4.png)

### New Media Gallery folders (*)

![image](https://user-images.githubusercontent.com/2028541/78028568-23e85880-7357-11ea-9f6c-a6f77bafbcdc.png)

Issue #1071, opened 2020-03-30
Page builder doesn't load images from Enhanced Media gallery, instead the "old" Media Gallery is used


chalov-anton:
### Preconditions (*)
Magento 2.4-develop
Adobe-Stock-Integration 1.1-develop

1. Installed Magento with Adobe Stock Integration
2. Configured integration in Stores -> Configuration -> Advanced-> System -> Adobe Stock Integration fieldset
3. Enable Enhanced Media Gallery in Stores -> Configuration -> Advanced -> System 
![enhansedMG](https://user-images.githubusercontent.com/45624059/75229948-9b0c5900-57bb-11ea-83d3-d150e4ae3db8.png)
4. **Enabled Page builder**

### Steps to reproduce (*)
1. From Admin go to **Content** - **Pages**, click **Add New Page**
2. Expand **Content**, Expand **Media**
3. Drag and drop **Image** on the Row section
4. Click **Select from Gallery** 

### Expected result (*)
**Not** Enhanced Media Gallery is opened, 

### Actual result (*)
The new _Enhanced Media Gallery_ is opened

PR #1091, opened 2020-04-01
Exclude dot directories and images in it from mediaGallery && Page builder doesn't load images from Enhanced Media gallery, instead the "old" Media Gallery is used

Nazar65:

1. magento/adobe-stock-integration#1084: Exclude directories starting with dot and their contents from media gallery
2. magento/adobe-stock-integration#1071: Page builder doesn't load images from Enhanced Media gallery, instead the "old" Media Gallery is used

1. Login to Admin Palnel
2.     Open Media Gallery i.e.:
3.         From Admin go to Content - Pages, click Add New Page)
4.         Expand Content,click Show / Hide Editor, click Insert Image...

Directories  starting with dots ignored from directory tree
Images from dot directories not displayed in media gallery grid

**Scenario 2**

1. Enabled Page builder
2. From Admin go to Content - Pages, click Add New Page
3. Expand Content, Expand Media
4. Drag and drop Image on the Row section
5. Click Select from Gallery

**The new Enhanced Media Gallery is opened**

### DIFF
--- a/MediaGalleryUi/etc/di.xml
+++ b/MediaGalleryUi/etc/di.xml
@@ -26,7 +26,7 @@
                 <item name="theme" xsi:type="string">/pub\/media\/theme/</item>
                 <item name="theme_customization" xsi:type="string">/pub\/media\/theme_customization/</item>
                 <item name="tmp" xsi:type="string">/pub\/media\/tmp/</item>
-		<item name="directories-with-dots" xsi:type="string">/\./</item>
+                <item name="directories-with-dots" xsi:type="string">/\./</item>
 	    </argument>
         </arguments>
      </type>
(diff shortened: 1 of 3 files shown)


MediaGalleryUi/etc/di.xml
```diff
@@ -26,6 +26,7 @@
                 <item name="theme" xsi:type="string">/pub\/media\/theme/</item>
                 <item name="theme_customization" xsi:type="string">/pub\/media\/theme_customization/</item>
                 <item name="tmp" xsi:type="string">/pub\/media\/tmp/</item>
+                <item name="directories-with-dots" xsi:type="string">/\./</item>
```
sivaschenko: @Nazar65 we need to exclude only directories that have the name starting with dot. Please take a look on my suggestion
```suggestion
                <item name="directories-with-dots" xsi:type="string">/^\./</item>
```

sivaschenko (approved):

sivaschenko (approved):

PR #1113, opened 2020-04-02
hotfix: Correct pattern to exclude directores started with dots .thumbnail

Nazar65:
### Description (*)

Directories started with dots like ,thumbnail, not excluded from directory tree

n/a

Create a lot of dots directories like .thumbnail or .thumbs, 
Expected result

Directory have no any directories started with dots, and media gallery grid have no images with this directories


hws47a (commented):
Hi @Nazar65, thank you for your contribution. 
Is there a way to write an automated test to check this scenario?

Nazar65:
@hws47a yes, but this will take some time, i think we can create a separate task for this.

sivaschenko (approved):

hws47a (approved):

PR #1131, opened 2020-04-04
FIX: clean up MediaGallery module

coderimus:
### Description (*)
Provide general clean up for the `MediaGallery` module.

N/A

N/A


sivaschenko (changes requested):
@coderimus can you please test if the MediaGalleryIndexerTrigger plugin is still functional after this change

coderimus:
@sivaschenko thank you for your review. Yes, a tricky thing eliminated :) I made a fix and push changes. Please, review.

MediaGalleryUi/Plugin/MediaGalleryIndexerTrigger.php
```diff
@@ -42,8 +42,8 @@ public function __construct(
      */
     public function afterSave(Value $config, Value $result): Value
     {
-        $isMediaGallery = $result->getPath() == self::MEDIA_GALLERY_CONFIG_VALUE;
-        if ($isMediaGallery && $result->isValueChanged() && $result->getValue() == self::MEDIA_GALLERY_ENABLED_VALUE) {
+        $isMediaGallery = $result->getPath() === self::MEDIA_GALLERY_CONFIG_VALUE;
+        if ($isMediaGallery && $result->isValueChanged() && (int) $result->getValue() === self::MEDIA_GALLERY_ENABLED_VALUE) {
```
hws47a: There is static content error:
```
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------
 46 | WARNING | Line exceeds 120 characters; contains 126 characters
----------------------------------------------------------------------
```

hws47a: I personaly would remove a variable and put all checks in one if statement like this
```suggestion
        if ($result->getPath() === self::MEDIA_GALLERY_CONFIG_VALUE 
            && $result->isValueChanged() 
            && (int)$result->getValue() === self::MEDIA_GALLERY_ENABLED_VALUE
        ) {
```

hws47a (changes requested):

MediaGalleryUi/Plugin/MediaGalleryIndexerTrigger.php
coderimus: @hws47a thank you for your patient to details! I appreciate this.

I made some changes. I moved `isValueChanged` from the if condition. I made this based on the next conclusion: identifying that the path is a media config and that the value changed looks for me as a stand-alone action. So good to keep them separate. What do you think?

hws47a: I think introducing variables for not complex parts of the code makes it slower to read.

Also for the current case with having a second variable, you call `isValueChanged` even then it's not needed, I mean if `$isMediaGallery` will be false.

coderimus: Agree. Will provide changes in a few minutes.

coderimus: @hws47a thank you for details. I made the recommended changes. Let's wait for the static tests finished and review :)

hws47a:
Hi @coderimus, it seems that Functional tests are failing for your changes. Could you please check them?

coderimus:
Hi @hws47a, I made an investigation and find that at the 1.1-develop the next bug appeared. It possibly caused functional tests failed result. https://github.com/magento/adobe-stock-integration/issues/1144

hws47a (approved):

coderimus:
@hws47a I merged the latest 1.1-develop and looks like everything is ok :)

hws47a (approved):

sivaschenko (approved):

PR #1135, opened 2020-04-05
Integration tests for MediaGalleryIntegration module and unit test adjustments

coderimus:
### Description (*)
This PR provides unit tests adjustments and integration tests for the `MediaGalleryIntegration` module.

#1216


MediaGalleryIntegration/Plugin/UpdateWysiwygOpenDialogUrl.php
```diff
@@ -60,14 +60,15 @@ public function afterGetConfig(DefaultConfigProvider $provider, DataObject $conf
...
+        $currentTreePath = $this->imagesHelper->idEncode(Config::IMAGE_DIRECTORY);
+        $filesBrowserWindowUrl = $this->url->getUrl(
+            'media_gallery/index/index',
+            [
+                'current_tree_path' => $currentTreePath,
+            ]
         );
+        $config->setData('files_browser_window_url', $filesBrowserWindowUrl);
+
+        return $config;
```
orlangur: Why?) It was more readable

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
```diff
@@ -66,9 +66,19 @@ protected function setUp(): void
...
+            ->method('isEnabled')
+            ->willReturn(false);
+
+        $this->imageComponentMock->expects($this->never())
+            ->method('getData');
+
+        $this->urlMock->expects($this->never())
+            ->method('getUrl');
+
+        $this->imageComponentMock->expects($this->never())
```
orlangur: Such strict assertions make test more fragile, why need to add them?

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
```diff
@@ -78,9 +88,32 @@ public function testAfterPrepareWhenConfigDisabled(): void
...
+        $this->imageComponentMock->expects($this->once())
+            ->method('getData')
+            ->willReturn([]);
+
+        $this->urlMock->expects($this->once())
+            ->method('getUrl')
+            ->with($urlPath)
+            ->willReturn($openDialogUrl);
+
+        $this->imageComponentMock->expects($this->once())
```
orlangur: Same here.

There is no need to "adjust unit tests to be closer to code", instead, we need to write tests in such a way that they won't need any adjustment in most cases of code changes.

orlangur (changes requested):

MediaGalleryIntegration/Plugin/UpdateWysiwygOpenDialogUrl.php
coderimus: I followed the next principle: move all data manipulations outside the `setData` method to avoid data creation in method parameters :)

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
coderimus: As I understood this was added to check that if the https://github.com/magento/adobe-stock-integration/blob/29acaec609d4d760cb00176d9bdd837824d2093d/MediaGalleryIntegration/Plugin/UpdateWysiwygOpenDialogUrl.php#L59 returns false no other execution will be not fired from the code perspective.

MediaGalleryIntegration/Plugin/UpdateWysiwygOpenDialogUrl.php
orlangur: Where it is derived from?

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
orlangur: In such case I would recommend asserting `$config` remain unchanged instead. Try to rely on behavior rather than on execution flow in tests.

orlangur (changes requested):

MediaGalleryIntegration/Plugin/UpdateWysiwygOpenDialogUrl.php
coderimus: It is only my personal representation. I think that it is not so clear to have aa more than one operation in the data set. Be came more complicated to read when you see a code at first :)

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
coderimus: @orlangur thank you for your review and suggestions. This test implemented like this because of the method returns `void` Please, advise how can I check that Image object which is a plugin param here doesn't have new data such as `['config']['mediaGallery']['openDialogUrl']`?

`$this->imageComponentMock->expects($this->never())` is a kind of behaviour: if config false we do not execute it and this means that we do not change original object.

MediaGalleryIntegration/Plugin/UpdateWysiwygOpenDialogUrl.php
sivaschenko: I think the existing code was better readable too. Also, I'd avoid creating variables for a single usage unless it's absolutely necessary. Can you please revert these changes

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
sivaschenko: I think it's ok to assert that there were no calls to `imageComponentMock` for a unit test, however, I don't see the point to mock a return value in this case

MediaGalleryIntegration/Test/Unit/Plugin/UpdateOpenDialogUrlTest.php
sivaschenko: The current approach is regular for Magento unit tests, however, I'd agree that integration tests would be more useful

sivaschenko (changes requested):
Thanks for the pull request @coderimus ! Can you please take a look at the review comment

coderimus:
this pr has a dependency on https://github.com/magento/adobe-stock-integration/pull/1186

coderimus:
@sivaschenko the #1186 will leave only one plugin as an entry point for updating open dialog URL. I will adjust this PR when it will be merged. But I have a question: should I leave integration tests on cases previously covered with 3 plugins? With leaving them we can be sure that in all existed cases open dialog URL will be correct. No matter it opens from tinyMce or wysiwig.

sivaschenko:
@coderimus I like the idea of verifying the correct url for media gallery in all cases

coderimus:
@sivaschenko @orlangur 
I merged the latest changes from the magento/2.4-develop and asi 1.1-develop and adopt integration tests to the current functionality. Also, change the unit test :) Please, review.

coderimus:
@sivaschenko @orlangur looks like failed tests do not have things in common with the PR. If no I will change.

MediaGalleryIntegration/Test/Integration/Plugin/UpdateOpenDialogUrlTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+
+declare(strict_types=1);
+
+namespace Magento\MediaGalleryIntegration\Test\Integration\Plugin;
```
sivaschenko: There is no `Plugin` namespace in MediaGalleryIntegration module anymore. Can you please correct namespaces and class names to reflect the test objective correctly

MediaGalleryIntegration/Test/Integration/Plugin/UpdateOpenDialogUrlTest.php
```diff
@@ -0,0 +1,71 @@
...
+use Magento\Framework\UrlInterface;
+use Magento\TestFramework\Helper\Bootstrap;
+use Magento\Ui\Component\Form\Element\DataType\Media\Image;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Provide integration tests covers update open dialog url functionality for media editor.
+ * @magentoAppArea adminhtml
+ */
+class UpdateOpenDialogUrlTest extends TestCase
```
sivaschenko:
```suggestion
class ImageComponentOpenDialogUrlTest extends TestCase
```

MediaGalleryIntegration/Test/Integration/Plugin/UpdateWysiwygDialogUrlTinyMceTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Magento\TestFramework\Helper\Bootstrap;
+use Magento\Framework\DataObject;
+use Magento\Tinymce3\Model\Config\Gallery\Config;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Provide integration tests covers update open dialog url functionality for media editor.
+ * @magentoAppArea adminhtml
+ */
+class UpdateWysiwygDialogUrlTinyMceTest extends TestCase
```
sivaschenko:
```suggestion
class TinyMceOpenDialogUrlTest extends TestCase
```

MediaGalleryIntegration/Test/Integration/Plugin/UpdateWysiwygDialogUrlTinyMceTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->objectManger = Bootstrap::getObjectManager();
+        $this->tinyMce3Config = $this->objectManger->create(Config::class);
+        $this->configDataObject = $this->objectManger->create(DataObject::class);
+
+        $url = $this->objectManger->create(UrlInterface::class);
+        $this->fileBrowserWindowUrl = $url->getUrl('media_gallery/index/index');
+    }
+
+    /**
+     * Test image open dialog url when enhanced media gallery not enabled.
```
sivaschenko: I'd add `@magentoConfigFixture default/system/media_gallery/enabled 0` for all tests to keep passing even if enhanced media gallery will be enabled by default

MediaGalleryIntegration/Test/Integration/Plugin/UpdateWysiwygOpenDialogUrlTest.php
```diff
@@ -0,0 +1,81 @@
...
+use Magento\Framework\ObjectManagerInterface;
+use Magento\Framework\UrlInterface;
+use Magento\TestFramework\Helper\Bootstrap;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Provide integration tests covers update wysiwyg editor dialog url update when media gallery enabled.
+ * @magentoAppArea adminhtml
+ */
+class UpdateWysiwygOpenDialogUrlTest extends TestCase
```
sivaschenko:
```suggestion
class WysiwygDefaultConfigOpenDialogUrlTest extends TestCase
```

MediaGalleryIntegration/Test/Unit/Model/OpenDialogUrlProviderTest.php
```diff
@@ -0,0 +1,77 @@
...
+    private $configMock;
+
+    /**
+     * @var OpenDialogUrlProvider
+     */
+    private $openDialogUrlProvider;
+
+    /**
+     * @inheritDoc
+     * @SuppressWarnings(PHPMD.UnusedFormalParameter)
```
sivaschenko: Unnecessary suppress warning

MediaGalleryIntegration/Test/Unit/Model/OpenDialogUrlProviderTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Magento\Framework\TestFramework\Unit\Helper\ObjectManager as ObjectManagerHelper;
+use Magento\MediaGalleryIntegration\Model\OpenDialogUrlProvider;
+use Magento\MediaGalleryUiApi\Api\ConfigInterface;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Unit test for \Magento\MediaGalleryIntegration\Plugin\UpdateWysiwygDialogUrlTinyMceTest
+ */
+class OpenDialogUrlProviderTest extends TestCase
```
sivaschenko: I would be good to cover `Magento\MediaGalleryIntegration\Model\OpenDialogUrlProvider` by integration test instead of the unit test:
1. Use `@magentoConfigFixture default/system/media_gallery/enabled` fixture to set the system state
2. Call `getUrl` method

sivaschenko (changes requested):
Thanks for updates @coderimus ! Please take a look at my code review comments

coderimus:
@sivaschenko thank you for the review. I implemented the requested changes. Please, review.

sivaschenko (approved):

sivaschenko:
@magento run Functional Tests EE

sivaschenko:
@magento run Functional Tests EE

PR #1159, opened 2020-04-07
The "See more" button doesn't work when More images are loaded and the same image is previewed again

### LINKED ISSUE #850: The "See more" button doesn't work when More images are loaded and the same image is previewed again
### Preconditions (*)
2. Configure the integration in Stores -> Configuration -> System -> Adobe Stock Integration fieldset
### Steps to reproduce (*)
1. From Admin go to **Content** - **Pages**, click **Add New Page**
2. Expand **Content**,click **Show / Hide Editor**, select **Insert Image...**
3. Click **Search Adobe Stock**
4. Click, for example on the first image to open the Preview
5. Click **See more**
see more link
6. From the updated grid, select the same image you've opened previously, and click **See more** 

### Expected result (*)
The page is scrolled to the top

### Actual result (*)
Nothing happens 

see more

engcom-Charlie:

1. magento/adobe-stock-integration#850: The "See more" button doesn't work when More images are loaded and the same image is previewed again
2. ...

1. ...
2. ...

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
@@ -14,7 +14,7 @@ define([
         defaults: {
             template: 'Magento_AdobeStockImageAdminUi/grid/column/preview/related',
             filterChipsProvider: 'componentType = filters, ns = ${ $.ns }',
-            filterInputSelector: 'input#words',
+            filterInputSelector: '#words',
             tabImagesLimit: 4,
             serieFilterValue: '',
             modelFilterValue: '',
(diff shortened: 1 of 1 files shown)


AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
```diff
@@ -14,7 +14,7 @@ define([
         defaults: {
             template: 'Magento_AdobeStockImageAdminUi/grid/column/preview/related',
             filterChipsProvider: 'componentType = filters, ns = ${ $.ns }',
-            filterTitleSelector: '.admin__current-filters-title-wrap',
+            filterInputSelector: '#words',
```
VladimirZaets: Hi, it's not a best practice depends on the element id. In general, it's not the best practice to use id in templates. Current frameworks reuse the template and sometimes we have the cases when template renders on the page a few times, and if we use ID we will get some untraceable issues because due to HTML rule, an ID can be presented only one time on the page.  Please take a is it possible to depend on html class instead of id

VladimirZaets (changes requested):
Hi @engcom-Charlie. Please take a look at my comments below

sivaschenko (approved):

PR #1161, opened 2020-04-07
#911 : Add "Used in" section to media gallery image details

### LINKED ISSUE #911: Add "Used in" section to media gallery image details
### Story

https://github.com/magento/adobe-stock-integration/issues/717 [Story 43] User sees pages the image is used in Magento

### Preconditions (*)
1. Install Magento and Adobe Stock Integration [See documentation](https://github.com/magento/adobe-stock-integration/wiki/Installation-Guide)

### Steps to reproduce (*)
1. Login to Admin panel
2. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
3. Open the menu for any image
4. Click "View details"

### Expected result (*)
"Used in" section is present in the image details and shows the content where the image was inserted

image

### Technical details (*)
A table has to be added providing the relation between the asset and cms blocks, pages, categories, products, any other entities containing WYSIWYG field with a possiblity to insert image form the media gallery.
The table should be updated when the content is saved with an image (or without an image if it was previously there)

Depends on: https://github.com/magento/magento2/pull/27536

konarshankar07:
### Description (*)

This PR will add the Used In section in the Image detail modal

1. magento/adobe-stock-integration#911: Add "Used in" section to media gallery image details

1. Login to Admin panel
2. Open Media Gallery (i.e. Catalog -> Category -> expand Content -> Select from Gallery button)
3. Open the menu for any image
4. Click "View details"

### Expected Result
image

Depends on: https://github.com/magento/magento2/pull/27536


MediaGalleryUi/Model/GetImageDetailsByAssetId.php
```diff
@@ -244,4 +268,28 @@ private function formatImageSize(int $imageSize): string
...
+     *
+     * @param int $assetId
+     * @return array
+     */
+    private function getUsedIn(int $assetId): array
+    {
+        $usedIn = [];
+        try {
+            foreach ($this->assetContentTypes as $assetContentTypeKey => $assetContentType) {
+                $getAssetIds = $this->getAssetsUsedInContentInterface->execute($assetContentTypeKey);
```
coderimus: I think for the performance perspective it would be better to represent a group selection where data will be filtered by the `$assetId` and `contentTypes`: `select main_table.asset_id from main_table.media_content_asset where main_table.asset_id = seetId and main_table.type in (contentTypes)` Also, it looks like a good point to exclude this part from the `GetImageDetailsByAssetId`

coderimus (changes requested):

coderimus:
@konarshankar07 thank you for your contribution! Please, see my comment about the getting data for the `Used In` calculation.

MediaGalleryUi/Model/GetImageDetailsByAssetId.php
konarshankar07: Yes, You are right. It might create performance issue. So should I add the function in `GetAssetsUsedInContent` class or update the existing one?
Thanks

coderimus: @konarshankar07 I think that this logic should be implemented in a separate class in the `MediaGalleryUi` module. The `GetAssetUsedIn` sounds good for its name, as for me.

konarshankar07: Okay got it

sivaschenko:
Thanks for the pull request! I am merging it to the media-content branch where we'll continue to work on functionality. It will be merged to 1.1-develop shortly

Issue #1228, opened 2020-04-17
Image preview is broken (Licensing)


chalov-anton:
### Preconditions (*)
Magento 2.4-develop
Adobe-Stock-Integration 1.1-develop

1. Installed Magento with Adobe Stock Integration
2. Configured integration in Stores -> Configuration -> Advanced-> System -> Adobe Stock Integration fieldset

### Steps to reproduce (*)
1. From Admin go to **Content** - **Media - Media Gallery** 
2. Click **Search Adobe Stock**, select an unlicensed image to open Preview
3. **Select the second image from More from this series** and click **Save Preview**
(the image appeared in Media Gallery and it is selected/highlighted)
4. Click on _three dots_ and select **License**. **Confirm** licensing
5. Click **Search Adobe Stock** and observe the image preview
![image_preview_2020-04-17](https://user-images.githubusercontent.com/45624059/79548788-a898dd80-809e-11ea-9eb5-e3aaea20d429.gif)

### Expected result (*)
![im_preview_1704_er](https://user-images.githubusercontent.com/45624059/79548279-e21d1900-809d-11ea-9169-b40a463e06d6.png)

### Actual result (*)
The image preview is broken
![im_preview_1704_ar](https://user-images.githubusercontent.com/45624059/79548170-ab470300-809d-11ea-8b0d-3a1850f204d7.png)

### Note 
There is another way to reproduce this issue. It will be reproducible after merging https://github.com/magento/adobe-stock-integration/pull/1249 I will also leave an identical comment for visibility  
If the secret key is not entered, and user tries to License ann image from Media Gallery
![preview_is broken_license](https://user-images.githubusercontent.com/45624059/79972838-3dd11300-849f-11ea-9896-3fbcdcaaa128.gif)

gabrieldagama:
Hi @chalov-anton, I've tried to reproduce this issue locally but haven't been able to, not sure if I'm doing something wrong, my branches are up to date and magento is updated.

![test-issue-1228](https://user-images.githubusercontent.com/3008360/79632525-c5ddb280-8157-11ea-837f-846ebab4098e.gif)

chalov-anton:
Hello @gabrieldagama 
We were able to reproduce this on two different environments.
The main step is the Step number 3. You need to switch to another image from _More from this series_ and Save it's preview. And after, license it from Media Gallery.

chalov-anton:
[...]

PR #1263, opened 2020-04-22
Fixed issue with preview on image license

gabrieldagama:
### Description (*)
I've added the adobe stock object to the details plugin, that is being used by JS and actually setting an adobe asset image to the displayedRecord instead of a media gallery image. There might be other ways to fix this issue, happy to discuss the approach.

1. magento/adobe-stock-integration#1228: Image preview is broken (Licensing) 

Follow the steps on #1228


coderimus:
Hello @gabrieldagama! Thank you for the provided changes. Please, resolve conflicts.

AdobeStockImage/Model/SerializeImage.php
```diff
@@ -0,0 +1,49 @@
...
+                $data[$attribute->getAttributeCode()] = $attribute->getValue();
+            }
+            return $data;
+        } catch (\Exception $exception) {
+            throw new SerializationException(
+                __(
+                    'An error occurred during image serialization: %error',
+                    ['error' => $exception->getMessage()]
+                )
+            );
```
coderimus, line 46:
```suggestion
            throw new SerializationException( __('An error occurred during image serialization.));
```
According to the latest changes in the ASI an exception should be thrown but not with a specified error message. We specified exception message previously but this appeared confusing to the end-user. An exception should be logged but not thrown. In that particular case, log message is not required because of the log present in the method which calls this one.

AdobeStockImageApi/Api/SerializeImageInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+declare(strict_types=1);
+
+namespace Magento\AdobeStockImageApi\Api;
+
+use Magento\Framework\Api\Search\Document;
+
+/**
+ * Class SerializeImageInterface
+ * @api
```
coderimus, line 15: @sivaschenko should we marked it as an `@api`?

AdobeStockImage/Model/SerializeImage.php
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Magento\AdobeStockImage\Model;
+
+use Magento\AdobeStockImageApi\Api\SerializeImageInterface;
+use Magento\Framework\Api\AttributeInterface;
+use Magento\Framework\Api\Search\Document;
+use Magento\Framework\Exception\SerializationException;
+
+/**
+ * Class used to serialize a Document object
```
coderimus, line 16:
```suggestion
 * Used for image attributes data serialization
```

AdobeStockImage/Model/SerializeImage.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+class SerializeImage implements SerializeImageInterface
+{
+    /**
+     * Serializes a Document object
```
coderimus, line 21:
```suggestion
     * Serializes an image document object
```

AdobeStockImage/Test/Unit/Model/SerializeImageTest.php
```diff
@@ -0,0 +1,76 @@
...
+        $attributeCategoryMock = $this->createMock(AttributeInterface::class);
+        $attributeCategoryMock->method('getAttributeCode')->willReturn('category');
+        $attributeCategoryMock->method('getValue')->willReturn(123);
+        $attribute240ThumbMock = $this->createMock(AttributeInterface::class);
+        $attribute240ThumbMock->method('getAttributeCode')->willReturn('thumbnail_240_url');
+        $attribute240ThumbMock->method('getValue')->willReturn("image-url");
+        $this->documentMock->expects($this->once())
+            ->method('getCustomAttributes')
+            ->willReturn([$attribute240ThumbMock, $attributeCategoryMock]);
+        $expectedResult = ['thumbnail_url' => 'image-url', 'category' => 123];
```
coderimus, line 56: Please, declare `private cosnt` for `iamge-url` and 123. 
Just a note :) as I can see in Magento in tests the most applicable value for any kind of id is 42. So we can follow the common way and change 123 to 42.

coderimus (changes requested):
@gabrieldagama thank you for resolving conflicts and static test fixes. I added some comments below. Please, check them.

AdobeStockImageApi/Api/SerializeImageInterface.php
sivaschenko: `SerializeImage` is an internal operation of `AdobeStockImage` module, I don't think we need an interface for this class at all

coderimus: Aha, got it. So, if we need this serialization can we declare `imageAttributesSerializer` as a parameter in di.xml for the `AddAdobeStockImageDetailsPlugin` and `GetRelatedImages`?

sivaschenko:
Merged to https://github.com/magento/adobe-stock-integration/pull/1275

PR #1279, opened 2020-04-26
1245: fix image still exists after delete

### LINKED ISSUE #1245: Deleting doesn't remove image from pub/media, thus it appears in Media Gallery after synchronizing
### Preconditions (*)
Magento 2.4-develop
Adobe-Stock-Integration 1.1-develop

1. Installed Magento with Adobe Stock Integration
3. Have an image uploaded to Media Gallery

### Steps to reproduce (*)
1. Go to **Content - Media Gallery**
2. Select an image from Media Gallery, click on _three dots_ and select **Delete**
3. Verify that the image disappears from Media Gallery 
4. In terminal run ```php bin/magento media-gallery:sync```
5. Open **Content - Media Gallery** again

### Expected result (*)
The image is deleted from ```pub/media``` and not displayed in Media Gallery after synchronizing 

### Actual result (*) 
The image is not deleted from ```pub/media```, thus it appears in Media Gallery after synchronizing

im_delete_2020-04-21

coderimus:
### Description (*)
To fix the reported issue I provide this PR: use `Magento\Cms\Model\Wysiwyg\Images\Storage::deleteFile` for image delete process. This method removes the image physically and has a plugin which will remove data from a data storage - `\Magento\MediaGallery\Plugin\Wysiwyg\Images\Storage::afterDeleteFile` I also choose this method because of `Magento_MediaGalleryUi` has a dependency and use the `Magento_CMS module`.

#1245 

Depends on: https://github.com/magento/magento2/pull/27994

### DIFF
--- a/MediaGalleryUi/Model/DeleteImage.php
+++ b/MediaGalleryUi/Model/DeleteImage.php
@@ -9,6 +9,8 @@ namespace Magento\MediaGalleryUi\Model;
 
 use Magento\Cms\Model\Wysiwyg\Images\Storage;
 use Magento\Framework\App\Filesystem\DirectoryList;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\MediaGalleryApi\Api\IsPathBlacklistedInterface;
 use Magento\Framework\Filesystem;
 use Magento\MediaGalleryApi\Api\Data\AssetInterface;
 
@@ -23,6 +25,11 @@ class DeleteImage
     private $imagesStorage;
 
     /**
+     * @var IsPathBlacklistedInterface
+     */
+    private $isPathBlacklisted;
+
+    /**
      * @var Filesystem
      */
     private $filesystem;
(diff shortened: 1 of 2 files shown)


MediaGalleryUi/Model/DeleteImage.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Delete asset from a storage
+     *
+     * @param AssetInterface $asset
+     */
+    public function execute(AssetInterface $asset): void
+    {
+        $mediaDirectory = $this->filesystem->getDirectoryRead(DirectoryList::MEDIA);
+        $absolutePath = $mediaDirectory->getAbsolutePath($asset->getPath());
```
sivaschenko, line 69: Please check if the path is blacklisted before deleting an image (`IsPathBlacklistedInterface`).
Also, please add the information that the asset is removed from the database via the plugin

sivaschenko (changes requested):
Thanks for the pull request, @coderimus ! please see my comment

sivaschenko:
@coderimus does this pull request depends on https://github.com/magento/magento2/pull/27994 ? Or what is the dependency?

coderimus:
@sivaschenko thank you for your review!
At the beginning of working on this PR, I faced with the issue I thought made those 2 PR related. But I was wrong. Dependencies removed.

sivaschenko (dismissed):

engcom-Bravo (changes requested):
:x: QA Failed

Now the behavior is opposite, the image is deleted from ```pub/media``` but, as I understand, it is not deleted from the database. As a result - broken images are displayed on Media Gallery grig

coderimus:
@engcom-Bravo thank you for the QA. May I asked you to make one more test: as I can see you added an image to the root directory and I faced with the same behaviour as you but could reproduce it only several times and than it was gone. Thought that it was my local problem.

Test scenario: add an image to any kind of folder except root media directory and then test delete it. 

Thank you

coderimus:
@sivaschenko I would like to request your expertise :) Yesterday when I made this PR [magento/magento2#27994](https://github.com/magento/magento2/pull/27994) I faced with an issue which I considered as my local. But it appears to be the real one according to the current PR QA result.

Steps to reproduce: add an image to the root media gallery folder. It's path should be `/download.jpg` for example. After that try to delete it. The result will be next: image deleted, an entry in the  `media_content_asset` still exists.

I found that the reason for this is that the `$path` [variable here](https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/MediaGallery/Model/ResourceModel/DeleteAssetsByPaths.php#L51) equals to the image name without `/` for cases as I described before.

I tried to fix it like on my screenshot and it seems to be work but I am not totally sure that this is the correct way. Please, check this.

image

sivaschenko:
@coderimus I don't think the path should be changed in the delete service, the correct path should be passed by client

Nazar65:
:heavy_check_mark: QA Passed  images deleted from physical storage, folder deleted from physical storage and from the media_gallery_asset
**Note** Tested with linked PR on Magento 2 https://github.com/magento/magento2/pull/27994

PR #1349, opened 2020-05-13
Issue #1343 - Fixed bug with filter and bookmarks

### LINKED ISSUE #1343: Adobe Stock grid filters are not responsive after removing applied serie/model filter from different browser
### Preconditions (*)
2. Have two different browsers (i.e. Chrome and Firefox)

### Short steps to reproduce (*)
1. Apply series/model filter from one browser
2. Clear filters in the second browser
3. Check filters in the first browser

### Steps to reproduce (*)
**In Browser 1**
1. Open Media Gallery from i.e. (Catalog -> Category -> Image)
2. Click Search Adobe Stock button to open Adobe Stock Grid
3. Click on any image to expand the preview
4. Click on "See more" link in the "More from this series" or "More from this model" tab to **apply series/model filter**

**In Browser 2**
5. Log in to the same Magento admin account from another browser (chrome incognito mode is not suitable) and remove the applied series/model filter
6. Open Media Gallery from i.e. (Catalog -> Category -> Image)
7. Click Search Adobe Stock button to open Adobe Stock Grid
8. Click "Clear all" link in the toolbar to drop applied filters

**Back to Browser 1**
9. Return to the original browser
10. Login and reopen the gird if the user was logged out
11. Refresh the page
12. Try to apply any filters

### Expected result (*)
Filters can be applied

### Actual result (*)
Filters cannot be applied

### Workaround (*)

Option 1. Click on Default view in the bookmarks dropdown
Option 2. Clear local storage
Option 3. Clear ui_bookmark table

### Technical details (*)

`"filters":{"applied":[]}` saved to ui_bookmarks table (instead of`"filters":{"applied":{}}`)
The issue is related to storing `serieFilterValue` and `modelFilterValue` in the local storage by `AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js` These stateful properties should be eliminated.

gabrieldagama:
### Description (*)

From my investigation, the problem was related to storing an empty applied filter object like this:
`"filters":{"applied":{}}`
This was happening because we were applying the series/model filter directly to the component property and not using the methods available to do so. When the clear method runs the bookmark gets saved with the following filter object:
`"filters":{"applied":{"placeholder":true}}`
Apparently that placeholder property is required for the `app/code/Magento/Ui/view/base/web/js/grid/filters/filters.js` work properly (possibly a bug).
By using the `app/code/Magento/Ui/view/base/web/js/grid/filters/filters.js` methods to apply the series/model filter it guarantees that the placeholder property won't be missed and the filters will work.

I didn't see any relation with serieFilterValue and modelFilterValue states, we can discuss a refactor to remove those from the `app/code/Magento/Ui/view/base/web/js/grid/filters/filters.js` component.

### Fixed Issues 

1. magento/adobe-stock-integration#1343: Adobe Stock grid filters are not responsive after removing applied serie/model filter from different browser

1. Follow the steps described on #1343

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/preview/related.js
@@ -222,7 +222,9 @@ define([
 
             data[typeId] = recordId;
 
-            this.filterChips().set('applied', data);
+            this.filterChips().clear();
+            this.filterChips().setData(data, true);
+            this.filterChips().apply();
 
             this.scrollToFilter();
         },


coderimus (dismissed):
Hello @gabrieldagama ! Thank you for the provided solution and detail description. I checked it from the code side and in 2 browsers to validate the expected result. It works 👍

coderimus:
Failed tests do not belong to the PR changes.

chalov-anton:
:heavy_check_mark: QA Passed

sivaschenko:
@magento run all tests

sivaschenko:
@magento run Functional Tests B2B,Functional Tests EE

sivaschenko (approved):

Issue #1342, opened 2020-05-12
Possible to save Adobe Stock grid view to bookmarks after opening and closing image preview


sivaschenko:
### Preconditions (*)
1. Install Magento with Adobe Stock Integration
2. Adobe Stock Integration is [configured](https://github.com/magento/adobe-stock-integration/wiki/Get-API-keys)

### Steps to reproduce (*)
1. Open Media Gallery (i.e. from Catalog -> Category -> Category Image)
2. Click Search Adobe Stock button to open Adobe Stock Grid
3. Click on any image to open the preview
4. Click on the image again to close the preview
5. Open bookmarks dropdown

### Expected result (*)
"Save view as..." option should not be available as the view exactly matches the Default view

### Actual result (*)
"Save view as..." option is available despite the view exactly matches the Default view
![image](https://user-images.githubusercontent.com/2028541/81680951-db3eb780-944b-11ea-84e7-027947f0f9fb.png)

PR #1352, opened 2020-05-14
Issue #1342 - Fixed issue with bookmarks when close preview

gabrieldagama:
### Description (*)

The problem is related to the lastOpenedImage property which isn't saved on the Default bookmark, the preview object looked like this: `"preview":{"visible":true,"sorting":false}`

The issue here is that lastOpenedImage is a state of the preview component, and when we close the preview, that is set to null, as consequence the bookmark object gets updated with this: `"preview":{"visible":true,"sorting":false,"lastOpenedImage":null}`

As a result, the comparison of the bookmarks is different from the Default, so the component allows us to save a new bookmark.

From my investigation, the state wasn't being saved on the Default bookmark because the lastOpenedImage is defined as null, and from what I could see, it's ignored. By setting that property as false, it allows the component to initialize the state correctly and the Default bookmarks receive the lastOpenedImage as null.

1. magento/adobe-stock-integration#1342: Possible to save Adobe Stock grid view to bookmarks after opening and closing image preview

1. Create a new admin user (otherwise, the user will have the Default bookmark saved on the DB) or truncate ui_bookmarks table.
2. Follow the steps here #1342

### DIFF
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/components/grid/column/image-preview.js
@@ -20,6 +20,7 @@ define([
             mediaGallerySelector: '.media-gallery-modal:has(#search_adobe_stock)',
             adobeStockModalSelector: '.adobe-search-images-modal',
             activeMediaGallerySelector: 'aside.modal-slide.adobe-stock-modal._show',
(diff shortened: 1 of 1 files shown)


gabrieldagama:
Closing this PR, the issue will be handled on M2 repository:
https://github.com/magento/magento2/pull/28220

PR #1507, opened 2020-06-19
#1476 :- Generate renditions for all images uploaded to magento

### LINKED ISSUE #1476: Generate renditions for all images uploaded to magento
### Description
Currently when the image is uploaded to the media gallery, the thumbnail is created for the image. The idea is to use a similar approach for generating the optimized version of the image.
 - When image is uploaded to the media gallery the optimized version of the image should be created ("Rendition")
 - Rendition should be created based on width/height specified in the configuration of MediaGalleryRenditions module
 - If the image width is greater than the width specified in the configuration - the image should be resized to the configuration width
 - If the image height is greater than the height specified in the configuration - the image should be resized to the configuration height
 - The image should be resized proportionally
 - Resized version of the image should be saved to the .renditions directory
 - Image path/name should be generated using the same approach as thumbnail generation
(see `\Magento\Cms\Model\Wysiwyg\Images\Storage::resizeFile`)

### Implementation
 - A `\Magento\MediaGalleryRenditionsApi\Api\GenerateRentitionsInterface` should be introduced with a single method `execute(string $path): void`
- An interface should be implemented by `\Magento\MediaGalleryRenditions\Model\GenerateRentitions`
- `GenerateRentitionsInterface` should be used to execute rendition generation during the image upload

konarshankar07:
### Description (*)

This PR will generate renditions image when image is uploaded using Magento Enhanced Media Gallery Image Uploader

1. Fixes magento/adobe-stock-integration#1476: Generate renditions for all images uploaded to magento

1.  Configure width/height field from system configuration fields ()
2. Upload image using Enhanced media gallery
3. Check pub/media directory to see the `.renditions` images

### Expected Result
image

image


konarshankar07:
@magento run all tests

konarshankar07:
@magento run all tests

MediaGalleryUi/Model/UploadImage.php
```diff
@@ -72,6 +81,7 @@ public function execute(string $targetFolder, string $type = null): void
         }
         $absolutePath = $mediaDirectory->getAbsolutePath($targetFolder);
         $uploadResult = $this->imagesStorage->uploadFile($absolutePath, $type);
+        $this->generateRenditions->execute($uploadResult['path'] . '/' . $uploadResult['file']);
```
sivaschenko: To cover all the cases of image upload:
 - Please move the generating renditions to synchronization process. Considering that `SynchronizeFilesInterface` is used here, using the renditions generation will cover both upload from media gallery and `bin/magento media-gallery:sync`
 - Please add renditions generation call to `\Magento\MediaGalleryIntegration\Plugin\SaveImageInformation::afterSave`
 - Please add renditions generation call to `\Magento\MediaGalleryIntegration\Plugin\SaveBaseCategoryImageInformation::afterMoveFileFromTmp`

MediaGalleryRenditionsApi/Api/ConfigInterface.php
```diff
@@ -6,7 +6,7 @@
 
 declare(strict_types=1);
 
-namespace Magento\MediaGalleryRenditionsApi\Model;
+namespace Magento\MediaGalleryRenditionsApi\Api;
```
sivaschenko: The `ConfigInterface` is not a service so it's better to keep it in the `Model` namespace (it is still the API module).

MediaGalleryRenditions/Model/GenerateRenditions.php
```diff
@@ -0,0 +1,94 @@
...
+use Magento\Framework\Image\AdapterFactory;
+use Magento\MediaGalleryRenditionsApi\Api\ConfigInterface;
+use Magento\MediaGalleryRenditionsApi\Api\GenerateRenditionsInterface;
+
+class GenerateRenditions implements GenerateRenditionsInterface
+{
+    /**
+     * @var AdapterFactory
+     */
+    private $_imageFactory;
```
sivaschenko: Please do not use the `_` property names prefix

MediaGalleryRenditions/Model/RenditionsImageManagement.php
```diff
@@ -0,0 +1,134 @@
...
+use Magento\Framework\Filesystem;
+use Magento\Framework\Filesystem\DriverInterface;
+use Magento\Framework\Filesystem\Io\File;
+
+class RenditionsImageManagement
+{
+    /**
+     * @var Filesystem\Directory\WriteInterface
+     */
+    private $_directory;
```
sivaschenko:
```suggestion
    private $directory;
```

MediaGalleryRenditions/Model/RenditionsImageManagement.php
```diff
@@ -0,0 +1,134 @@
...
+    public function execute(string $path) :string
+    {
+        $realPath = $this->_directory->getRelativePath($path);
+        if (!$this->_directory->isFile($realPath) || !$this->_directory->isExist($realPath)) {
+            throw new LocalizedException(__('Directory or File %1 does not exist in media directory.', $realPath));
+        }
+        $renditionImageDirectoryPath = $this->getRenditionsImageDirectory($path);
+        $renditionImageDirectory = $this->_directory->getRelativePath($renditionImageDirectoryPath);
+        if (!$this->_directory->isExist($renditionImageDirectory)) {
+            $this->_directory->create($renditionImageDirectory);
```
sivaschenko: Let's create the directory only when creating a rendition. Let's just return the constructed rendition path from this service and do not check if directory or file exists

MediaGalleryRenditions/Model/RenditionsImageManagement.php
```diff
@@ -0,0 +1,134 @@
...
+
+use Magento\Framework\App\Filesystem\DirectoryList;
+use Magento\Framework\App\ObjectManager;
+use Magento\Framework\Exception\FileSystemException;
+use Magento\Framework\Exception\LocalizedException;
+use Magento\Framework\Filesystem;
+use Magento\Framework\Filesystem\DriverInterface;
+use Magento\Framework\Filesystem\Io\File;
+
+class RenditionsImageManagement
```
sivaschenko: Let's rename the class and add a `Magento\MediaGalleryRenditionsApi\Api\GetRenditionPathInterface`
```suggestion
class GetRenditionPath implements GetRenditionPathInterface
```

MediaGalleryRenditions/Model/RenditionsImageManagement.php
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Returns Rendition image path
+     *
+     * @param string $path
+     * @return string
+     * @throws LocalizedException
+     * @throws FileSystemException
+     */
+    public function execute(string $path) :string
```
sivaschenko: I think it would be handier to provide ability to get rendition path based on asset id
```suggestion
    public function execute(int $assetId): string
```

MediaGalleryRenditionsApi/Api/GenerateRenditionsInterface.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /**
+     * Generate Renditions image
+     *
+     * @param string $path
+     * @throws FileSystemException
+     * @throws LocalizedException
+     * @throws Exception
+     */
+    public function execute(string $path): void;
```
sivaschenko: Considering that there are use cases of assets processing in bulk, I think it would be better to make this service bulk as well
```suggestion
    public function execute(array $paths): void;
```

MediaGalleryRenditionsApi/Api/GenerateRenditionsInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+interface GenerateRenditionsInterface
+{
+    /**
+     * Generate Renditions image
+     *
+     * @param string $path
+     * @throws FileSystemException
+     * @throws LocalizedException
+     * @throws Exception
```
sivaschenko: It would be good to ensure only LocalizedException can be thrown from an API service call

sivaschenko (changes requested):
Thanks for the pull request @konarshankar07 ! Please see my review comments

sivaschenko:
@konarshankar07 as for test coverage: I think it's better to cover the whole functionality by the integration tests for `GetRenditionInterface` and `GenerateRentionsInterface` services. I don't think unit tests are necessary

MediaGalleryRenditionsApi/Api/ConfigInterface.php
konarshankar07: Okay, But I moved into the `Api` namespace because of the inconsistency with the other module. If you check the other `ConfigInterface` in the repository they are under the `Api` namespace so should I move the other `ConfigInterface` to the `Model` namespace?

MediaGalleryRenditions/Model/GenerateRenditions.php
konarshankar07: Okay, But may I ask why its not recommended?
Thanks

sivaschenko: Based on the PSR https://www.php-fig.org/psr/psr-2/#42-properties

MediaGalleryRenditionsApi/Api/ConfigInterface.php
sivaschenko: ConfigInterface is not designed to be used outside of the MediaGalleryRenditions module so it's not part of actual module's API, but it still can be customized (preference, plugin, etc)

konarshankar07:
@magento run all tests

MediaGalleryRenditions/Model/GenerateRenditions.php
```diff
@@ -0,0 +1,95 @@
...
+        $this->getRenditionPath = $getRenditionPath;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute(string $path): void
+    {
+        $isResizeable = $this->isResizeable($path);
+        if ($isResizeable) {
```
gabrieldagama:
```suggestion
        if ($this->isResizeable($path)) {
```

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Renditions path getter
+     *
+     * @param  string $filePath original file path
+     * @param  bool $checkFile OPTIONAL is it necessary to check file availability
+     * @return string|false
+     */
+    private function getRenditionsPath($filePath, $checkFile = false)
```
gabrieldagama: Since this is a private function and we (currently) are not using the $checkFile I would remove it and just not perform the check, or, if the check is important I would always do it. I don't think we need to add this complexity here as an option for a client on private function.

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        $mediaRootDir = $this->directory->getAbsolutePath('');
+        if (strpos($filePath, (string) $mediaRootDir) === 0) {
```
gabrieldagama: I think we should negate this condition and throw an exception if true, you can place the rest of the function outside the if. It will make it easier to understand the validation and you will be able to set strict type on function return.

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,135 @@
...
+     * @param bool|string $filePath Path to the file
+     * @return string
+     */
+    private function getRenditionsImageDirectory($filePath = false) :string
+    {
+        $renditionRootDir = $this->getRenditionsRoot();
+
+        if ($filePath) {
+            $renditionImagePath = $this->getRenditionsPath($filePath, false);
+            if ($renditionImagePath) {
```
gabrieldagama: By having getRenditionsPath throwing an exception we will be able to avoid this if.

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,135 @@
...
+        if (!$this->directory->isExist($renditionImageDirectory)) {
+            $this->directory->create($renditionImageDirectory);
+        }
+        if (!$this->directory->isExist($renditionImageDirectory)) {
+            throw new LocalizedException(__(
+                'Directory %1 does not exist in media directory.',
+                $renditionImageDirectory
+            ));
+        }
+        return $renditionImageDirectoryPath . '/' . $this->ioFile->getPathInfo($path)['basename'];
```
gabrieldagama:
```suggestion
        return $renditionImageDirectoryPath . DIRECTORY_SEPARATOR . $this->ioFile->getPathInfo($path)['basename'];
```

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,135 @@
...
+     */
+    private function getRenditionsPath($filePath, $checkFile = false)
+    {
+        $mediaRootDir = $this->directory->getAbsolutePath('');
+        if (strpos($filePath, (string) $mediaRootDir) === 0) {
+            $relativeFilePath = substr($filePath, strlen($mediaRootDir));
+            // phpcs:ignore Magento2.Functions.DiscouragedFunction
+            $renditionPath = $relativeFilePath === basename($filePath)
+                ? $this->getRenditionsRoot() . DIRECTORY_SEPARATOR . $relativeFilePath
+                : $this->getRenditionsRoot() . $relativeFilePath;
```
gabrieldagama:
```suggestion
            $renditionPath = $this->getRenditionsRoot();
            if ($relativeFilePath === basename($filePath)) {
                $renditionPath .= DIRECTORY_SEPARATOR;
            }
            $renditionPath .= $relativeFilePath;
```

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,135 @@
...
+        return $renditionImageDirectoryPath . '/' . $this->ioFile->getPathInfo($path)['basename'];
+    }
+
+    /**
+     * Return renditions directory path for file/current directory
+     *
+     * @param bool|string $filePath Path to the file
+     * @return string
+     */
+    private function getRenditionsImageDirectory($filePath = false) :string
```
gabrieldagama: Lets make this $filePath strict string and adapt the validation?

gabrieldagama (changes requested):
Hi @konarshankar07 thanks for the PR, it look great! Please follow some comments, let me know if you have any questions!

MediaGalleryRenditions/Model/RenditionsImageManagement.php
konarshankar07: @sivaschenko ... I think we can get the assets by paths using `Magento\MediaGalleryApi\Api\GetAssetsByPathsInterface` so instead of passing assetId we can pass the `Magento\MediaGalleryApi\Api\Data\AssetInterface` as parameter. Let me know your thoughts

sivaschenko:
Hi @konarshankar07 thanks for the pull request! Would you like to continue progress on it?

konarshankar07:
Hello @sivaschenko ..
Yes, I'm working on this but Can you please give the reply for https://github.com/magento/adobe-stock-integration/pull/1507#discussion_r445981166?
Thanks

MediaGalleryRenditions/Model/RenditionsImageManagement.php
sivaschenko: If it's more comfortable for the clients to use the path as an argument - that's ok with me

sivaschenko:
@konarshankar07 replied
Thanks

konarshankar07:
@sivaschenko and @gabrieldagama 
I've finished the feedback changes.
Thanks

MediaGalleryIntegration/Plugin/SaveBaseCategoryImageInformation.php
```diff
@@ -49,13 +50,18 @@ class SaveBaseCategoryImageInformation
      * @var Storage
      */
     private $storage;
+    /**
```
sivaschenko: please add an empty line between properties

MediaGalleryRenditions/Model/GenerateRenditions.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return $width > $this->getResizedWidth() || $height > $this->getResizedHeight();
+    }
+
+    /**
+     * Get resized image width
+     *
+     * @return int
+     */
+    private function getResizedWidth() :int
```
sivaschenko: For the consistency
```suggestion
    private function getResizedWidth(): int
```

MediaGalleryRenditions/Model/GenerateRenditions.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        return $this->config->getWidth();
+    }
+
+    /**
+     * Get resized image height
+     *
+     * @return int
+     */
+    private function getResizedHeight() :int
```
sivaschenko:
```suggestion
    private function getResizedHeight(): int
```

MediaGalleryRenditions/Model/GenerateRenditions.php
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Check if image needs to resize or not
+     *
+     * @param int $height
+     * @param int $width
+     * @return bool
+     */
+    private function isResizeable(int $height, int $width) :bool
```
sivaschenko:
```suggestion
    private function isResizeable(int $height, int $width): bool
```

MediaGalleryRenditions/Model/GenerateRenditions.php
```diff
@@ -0,0 +1,109 @@
...
+        $this->imageFactory = $imageFactory;
+        $this->config = $config;
+        $this->getRenditionPath = $getRenditionPath;
+        $this->createAssetFromFile = $createAssetFromFile;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function execute(array $files): void
```
sivaschenko: Should be an array of `paths` to comply with the interface.

However, returning to the discussion we had, it looks like all the places calling this method have access to an `AssetInterface` and this method implementation requires an `AssetInterface` as well, to avoid double `createAssetFromFile` call, wouldn't it be better to accept an `AssetInterface`  as a parameter?

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @param Filesystem $filesystem
+     * @param DriverInterface|null $file
+     * @throws FileSystemException
+     */
+    public function __construct(
+        Filesystem $filesystem,
+        DriverInterface $file = null
+    ) {
+        $this->directory = $filesystem->getDirectoryWrite(DirectoryList::MEDIA);
```
sivaschenko: Write access is not required
```suggestion
        $this->directory = $filesystem->getDirectoryRead(DirectoryList::MEDIA);
```

MediaGalleryRenditions/Model/GetRenditionPath.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Returns Rendition image path
+     *
+     * @param AssetInterface $asset
+     * @return string
+     * @throws LocalizedException
+     */
+    public function execute(AssetInterface $asset) :string
+    {
+        return $this->getRenditionsImageDirectory($asset->getPath());
```
sivaschenko: I wonder if we can simplify this method implementation to
```suggestion
        if (!$this->directory->isFile($asset->getPath())) {
             throw new InvalidArgumentExcetpion(__('Incorrect asset path!'))
        }
        if (!$this->isResizable($asset)) {
             return $asset->getPath();
        }
        return $this->directory->getRelativePath(RENDITIONS_DIRECTORY_NAME . '/' . $this->directory->getRelativePath($asset->getPath()));
```
The first condition should ensure that asset path is within media directory

MediaGalleryIntegration/Plugin/SaveImageInformation.php
```diff
@@ -84,24 +91,28 @@ public function __construct(
...
     {
-        $file = $this->splFileInfoFactory->create($result['path'] . '/' . $result['file']);
+        $file = $this->splFileInfoFactory->create($result['path'] . DIRECTORY_SEPARATOR . $result['file']);
         if (!$this->isApplicable($file->getPathName())) {
             return $result;
         }
         $this->saveAsset->execute([$this->createAssetFromFile->execute($file)]);
-        $this->storage->resizeFile($result['path'] . '/' . $result['file']);
-
+        $this->storage->resizeFile($result['path'] . DIRECTORY_SEPARATOR . $result['file']);
```
sivaschenko: @konarshankar07 @gabrieldagama I would prefer to use `/` instead of `DIRECTORY_SEPARATOR`.

Do you see any possible issues with this approach?

sivaschenko (changes requested):
Hi @konarshankar07 , very nice implementation, please see my comments.

Also, can you please cover the API interfaces with integration tests

MediaGalleryRenditions/Model/GetRenditionPath.php
konarshankar07: Write access is required to create the rendition directory path

MediaGalleryIntegration/Plugin/SaveImageInformation.php
konarshankar07: I think @gabrieldagama prefer to use `DIRECTORY_SEPARATOR` because of  Windows and Linux have different directory separator but Magento is not compatible with Windows so I think both the approaches is fine for me

konarshankar07:
@sivaschenko ... If you approve these changes then I will work on integration testing
Thanks

MediaGalleryIntegration/Plugin/SaveImageInformation.php
gabrieldagama: I usually go for this constant because I find it easier to read, but '/' is fine as well.

sivaschenko: I think '/' should be fine for Windows if we are constructing paths. Using the `DIRECTORY_SEPARATOR` constant can result into issues where '\\' is used in URL links (considering that Magento filesystem can be switched to https adapter).

So functionally I think using `DIRECTORY_SEPARATOR` is more risky for Windows OS comparing to '/' in most cases.

MediaGalleryRenditions/Model/GetRenditionPath.php
sivaschenko: This class is responsible for constructing and returning a rendition path only, I don't think it should not create directories

sivaschenko:
@konarshankar07 there are comments that have not been addressed, can you please check them out?

MediaGalleryRenditions/Model/GetRenditionPath.php
konarshankar07: I don't think we need to return the asset path if the image size is not resizeable cause GetRenditionPath should be called if the image is resizeable. Let me know your thoughts
Thanks

MediaGalleryRenditions/Model/GetRenditionPath.php
konarshankar07: So how we will handle the issue if the rendition directory is not available in the media directory?

konarshankar07:
@sivaschenko ... I have replied to your comments
Thanks

MediaGalleryIntegration/Plugin/SaveImageInformation.php
konarshankar07: So Do you think we need to change all `DIRECTORY_SEPARATOR` to '/'?

sivaschenko: It's not necessary for now, just shared my thoughts to bring the attention, I believe it requires a discussion to take a final decision and reflect it in our standards

MediaGalleryRenditions/Model/GetRenditionPath.php
sivaschenko: I think returning the asset path if image is not resizable might simplify the client implementation

MediaGalleryRenditions/Model/GetRenditionPath.php
sivaschenko: My point is that `GetRenditionPath` path class does not create a directory anyways, so the write access is not needed. It will be needed during the operation of saving the rendition to the filesystem

sivaschenko:
@konarshankar07 thank you! Please see my responses

MediaGalleryRenditions/Model/GetRenditionPath.php
konarshankar07: Sorry, But I'm not able to understand how it simplify the implementation because as per the service name which is `GetRenditionPath` and it should return the rendition path but we are returning the asset path if the image is not resizeable which doesn't make sense to me. Another point if we return the asset path then we need to add an extra condition in the GenerateRenditions service to check if the asset path is not same as generate rendition path before image resizing operation and I think it will make the implementation complex. let me know your thoughts

sivaschenko: I believe it will be comfortable to use `GetRenditionPath` service without additional checks and conditions. The call to the service should return the path to the image that is ready to be used for the content (either from `.rendition` directory or not).
There is a check for `isResizeable` in the `GenerateRenditions` and I don't think any additional checks are needed.

konarshankar07:
@sivaschenko and @gabrieldagama .. Please review this PR
Thanks

sivaschenko (changes requested):
@konarshankar07 can you please update the branch and resolve conflicts

MediaGallerySynchronization/etc/di.xml
```diff
@@ -15,6 +15,7 @@
             <argument name="importers" xsi:type="array">
                 <item name="0" xsi:type="object">Magento\MediaGallerySynchronization\Model\ImportMediaAsset</item>
                 <item name="1" xsi:type="object">Magento\MediaGallerySynchronization\Model\ImportImageFileKeywords</item>
+                <item name="generateRenditionImages" xsi:type="object">Magento\MediaGallerySynchronization\Model\GenerateRenditionImages</item>
```
sivaschenko: Can you please move this configuration to renditions module

MediaGallerySynchronization/composer.json
```diff
@@ -5,6 +5,7 @@
         "php": "~7.3.0||~7.4.0",
         "magento/framework": "*",
         "magento/module-media-gallery-api": "*",
+        "magento/module-media-gallery-renditions-api": "*",
```
sivaschenko: The dependency can be reveresed by extending configuration from renditions module

sivaschenko (changes requested):
Thanks  for updates @konarshankar07 ! Please see my comments

MediaGallerySynchronization/etc/di.xml
sivaschenko: Also, please use integer array key as it will be used as a sort order

MediaGalleryRenditions/Model/GenerateRenditionImages.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->generateRenditions = $generateRenditions;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function execute(string $path): void
+    {
+        $file = $this->splFileInfoFactory->create($path);
+        $this->generateRenditions->execute([$this->createAssetFromFile->execute($file)]);
```
sivaschenko: I'd like to avoid the dependency on non-api module. Looks like to achieve this it's needed to allow generating renditions based on paths (to remove the need to create asset object)

MediaGalleryRenditions/etc/module.xml
```diff
@@ -6,5 +6,9 @@
  */
 -->
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Module/etc/module.xsd">
-    <module name="Magento_MediaGalleryRenditions" />
+    <module name="Magento_MediaGalleryRenditions">
+        <sequence>
+            <module name="Magento_MediaGallerySynchronizationApi" />
```
sivaschenko: The sequece is not needed, it should not matter which of these modules is loaded first

sivaschenko (changes requested):
Thanks for updates @konarshankar07 , please see my review comments.

Also, there is a huge update in 2.0-develop branch, those updates do affect this pull request, however I believe there is no critical conflicts. Could you please update the branch and recheck the functionality to be sure.

MediaGalleryRenditions/etc/module.xml
konarshankar07: If I'm not adding the sequence then `ImportFilesComposite` is not executing the `GenerateRenditionImages`

sivaschenko:
Integration test coverage extracted to https://github.com/magento/adobe-stock-integration/issues/1751

sivaschenko:
@magento run all tests

sivaschenko (approved):

PR #1510, opened 2020-06-22
#1496: Add hash to asset on uploading the image to magento

### LINKED ISSUE #1496: Add hash to asset on uploading the image to magento
Story: https://github.com/magento/adobe-stock-integration/issues/1484

### Preconditions (*)
1. Install Magento 2.4-develop with ASI 2.0-develop
2. Enable Enhanced Media Gallery

### Steps to reproduce (*)
1. Login to Admin panel
2. Open the media gallery. I.e. from the Catalog -> Category -> Content fieldset -> Select image from gallery
3. Click "Upload image" button and upload an image

### Expected result (*)
The images information should be saved to the `media_gallery_asset` table. The `hash` filed should be populated with the `sha1` hash of the image content

### Testing

Should be checked for the next scenarios:
 - Uploading the image to the enhanced media gallery
 - Uploading the image to the old media gallery
 - Uploading the image to the category
 - Uploading the image to the Page builder
 - Runing `bin/magento media-gallery:sync` for images added manually to pub/media directory

### Technical Details

The changes should be applied to `\Magento\MediaGallerySynchronization\Model\CreateAssetFromFile::execute` method

joweecaquicla:
### Description (*)

This PR will implement the adding of hash to the asset on uploading the image to Magento.

1. Fixes https://github.com/magento/adobe-stock-integration/issues/1496: Add hash to asset on uploading the image to magento
2. ...


joweecaquicla:
@magento run all tests

MediaGallerySynchronization/Model/CreateAssetFromFile.php
```diff
@@ -137,4 +145,20 @@ private function getMediaDirectory(): Read
...
+     * @param string $path
+     * @return string
+     * @throws FileSystemException
+     * @throws ValidatorException
+     */
+    private function getHashImageContent(string $path): string
+    {
+        $mediaDirectory = $this->filesystem->getDirectoryRead(DirectoryList::MEDIA);
+        $imageContent = $mediaDirectory->readFile($this->getRelativePath($path));
+        $hashedImageContent = sha1($imageContent);
```
sivaschenko: Could you please wrap the `sha1` hashing call to a new class to be able to reuse it.

`Magento/MediaGallerySynchronization/Model/GetContentHash` class implementing `Magento/MediaGallerySynchronizationApi/Model/GetContentHashInterface` with a single `public function execute(string $content): string` returning the hash.

sivaschenko (changes requested):
Thanks @joweecaquicla ! Please take a look at the review comment.

MediaGallerySynchronization/Model/CreateAssetFromFile.php
lenaorobei: Why cannot we use `sha1_file` here? Another point I would consider is moving this "anti corruption layer" layer to framework.

joweecaquicla: @sivaschenko this is noted.
@lenaorobei I’ve looked up about `sha1` and `sha1_file` earlier and I think `sha1_file` should be used instead since we’re dealing with image files.

joweecaquicla:
@magento run all tests

MediaGallerySynchronization/Model/FileEncryption/Sha1File.php
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+declare(strict_types=1);
+
+namespace Magento\MediaGallerySynchronization\Model\FileEncryption;
+
+use Magento\MediaGallerySynchronizationApi\Model\FileEncryptionInterface;
+
+class Sha1File implements FileEncryptionInterface
```
sivaschenko: I don't think this additional class and interface are needed. It's enough to have only GetContentHash and GetContentHashInterface.

MediaGallerySynchronizationApi/Model/GetContentHashInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+interface GetContentHashInterface
+{
+    /**
+     * Get hashed value of image content.
+     *
+     * @param string $filepath
+     * @return string
+     */
+    public function execute(string $filepath): string;
```
sivaschenko:
```suggestion
    public function execute(string $content): string;
```

It would be better to work with file content instead of the file (we wouldn't like to include file path/name in the hash) as it will not allow us to use this hash for duplicates detection

sivaschenko (changes requested):
Thanks for the updates @joweecaquicla Please see my comments

sivaschenko:
@joweecaquicla can you please cover `GetContentHashInterface` with the integration test.

The test cases for the integration test:
 - Have two files with the same content.
 - Get hashes from these two files using `GetContentHashInterface`
 - The hashes should be equal

MediaGallerySynchronizationApi/Model/GetContentHashInterface.php
joweecaquicla: If we're gonna go with hashing the file content, then we'll go with `sha1()` instead of `sha1_file()` since the latter uses file path as a parameter instead of the entire file content.

joweecaquicla:
@magento run all tests

sivaschenko (dismissed):

sivaschenko:
Covered with integration tests, doesn't require manual testing

joweecaquicla:
@magento run all tests

sivaschenko:
@magento run Integration Tests

chalov-anton:
:heavy_check_mark: QA Passed

Test on Cucumber Studio - https://studio.cucumber.io/projects/131313/test-plan/folders/1320712/scenarios/4942139
(Test State - _In Progress_)

PR #1547, opened 2020-06-29
ASI-1543: represent an error message text into the console

### LINKED ISSUE #1543: Uncaught JS Error after authorization failure on licensing image from media gallery
### Preconditions

3. Provided _API Key (Client ID)_ 
client_key

### Steps to reproduce (*)

1. Go to **Content - Media Gallery**, click **Search Adobe Stock**
2. Save an _Unlicensed image_ Preview
(_the image appeared in Media Gallery_)
3. Go to _Stores - Configuration - Advanced - System - Adobe Stock Integration_ and **remove** the _API Key (Client ID)_. **Save Config**
no_keys
4. Go to **Content - Media Gallery** and select the previously saved Unlicensed image
5. Click "three dots" and select **Lisense**

### Expected result (*)
Adobe error is displayed in an opened window, after closing the window no additional alerts of console errors are displayed

### Actual result (*)
Adobe error appears.
After closing the window:
`Uncaught TypeError: this.imageModel is not a function` error appears in dev console

uncought

### Video!

Peek 2020-06-26 13-09

coderimus:
### Description (*)
This PR provides changes related to the #1543 
As a suggestion, I just represent the error message text received from the Adobe auth service directly into the console.

Error from the service
image

Console representation
image

@chalov-anton please, validate this approach.

#1543 

#1543

### DIFF
--- a/AdobeIms/view/adminhtml/web/js/action/authorization.js
+++ b/AdobeIms/view/adminhtml/web/js/action/authorization.js
@@ -113,6 +113,11 @@ define([
                 if (authWindow.closed) {
                     clearTimeout(stopWatcherId);
                     clearInterval(watcherId);
+
+                    // eslint-disable-next-line max-depth
+                    if (window.adobeIMSAuthWindow && window.adobeIMSAuthWindow.closed) {
+                        deferred.reject();
+                    }
                 }
             }
         }
--- a/AdobeIms/view/adminhtml/web/js/action/authorization.js
+++ b/AdobeIms/view/adminhtml/web/js/action/authorization.js
@@ -113,11 +113,6 @@ define([
                 if (authWindow.closed) {
                     clearTimeout(stopWatcherId);
                     clearInterval(watcherId);
-
-                    // eslint-disable-next-line max-depth
-                    if (window.adobeIMSAuthWindow && window.adobeIMSAuthWindow.closed) {
-                        deferred.reject(new Error('Authentication window was closed.'));
-                    }
                 }
             }
         }
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/mediaGallery/grid/columns/image/licenseActions.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/mediaGallery/grid/columns/image/licenseActions.js
@@ -98,13 +98,11 @@ define([
                     this.imageModel().addMessage('success', $.mage.__('The image has been licensed.'));
                 }.bind(this)).fail(function (error) {
                     if (error) {
-                        console.log(error.message);
+                        this.imageModel().addMessage('error', error);
                     }
                 });
             }.bind(this)).fail(function (message) {
-                if (message) {
-                    console.log(message);
-                }
+                this.imageModel().addMessage('error', message);
             });
         }
     });
--- a/AdobeStockImageAdminUi/view/adminhtml/web/js/mediaGallery/grid/columns/image/licenseActions.js
+++ b/AdobeStockImageAdminUi/view/adminhtml/web/js/mediaGallery/grid/columns/image/licenseActions.js
@@ -98,11 +98,13 @@ define([
                     this.imageModel().addMessage('success', $.mage.__('The image has been licensed.'));
                 }.bind(this)).fail(function (error) {
                     if (error) {
-                        this.imageModel().addMessage('error', error);
+                        console.log(error.message);
                     }
                 });
             }.bind(this)).fail(function (message) {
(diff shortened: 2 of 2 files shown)


AdobeStockImageAdminUi/view/adminhtml/web/js/mediaGallery/grid/columns/image/licenseActions.js
```diff
@@ -98,11 +98,13 @@ define([
                     this.imageModel().addMessage('success', $.mage.__('The image has been licensed.'));
                 }.bind(this)).fail(function (error) {
                     if (error) {
-                        this.imageModel().addMessage('error', error);
```
Nazar65: Can we refactor this method to be a "deferred" "promise" ? and catch any error ? and then display it in our grid messages ? and show the origin message about wrong credentials

Nazar65 (changes requested):

AdobeStockImageAdminUi/view/adminhtml/web/js/mediaGallery/grid/columns/image/licenseActions.js
coderimus: Good point @Nazar65 . Yes, I think it is possible but in the case of the Sign In the error is shown in the Adobe Stock Sign In window and the text we get from it describes only Authentification window was closed.To show the same error message twice think not so useful. @chalov-anton 

This is the signing window with an error message. Looks informative.
image

chalov-anton: @coderimus Agree, the text in the message looks informative enough.

chalov-anton:
@coderimus 
Please, hide the message from Adobe into the console, and display an Attention message with a text as we have in Search Adobe Stock.
image

Something like this would be informative :+1: 
attention

@sivaschenko What do you think?

sivaschenko:
@chalov-anton agree, @coderimus I don't see the reason why we would log the error to the console instead of showing appropriate message to the user.

sivaschenko:
@coderimus any updates on this PR? @chalov-anton do you think we should show any error popups at all if the auth window was closed by the user? (I believe the user will be able to see the appropriate error in the auth window itself before closing it)

chalov-anton:
@sivaschenko I think yes, we should. Because the auth window is not informative. It contains an error code and text: _There was an error with this action. Try again later_
85845185-8e6d3100-b7ac-11ea-8d6a-61188e8281af

If we are able to display an appropriate text in the adobe auth window, then I think no, the popup is not needed after the "auth" is closed.

coderimus:
@sivaschenko @chalov-anton 
We have 2 variants when the auth window opens:
- "Error There was an error with this action. Try again later" it is on the domain level and our regexp parsing doesn't even start. And this error looks like an error on the Adobe Stock side and doesn't relate to the API credentials. You can get it event of you have correct API keys.
- More clear error about incorrect client_id

I am trying to identify how we can catch them and send them back to user but, personally, I do not think that we should show something after the auth window closed. Because in all cases this will be duplication.

Situation 1: Incorrect CLIENT_ID
- Auth window shows a message about the  incorrect client id
- We show the same

Situation 2: Sign In service down on the Adobe Side
- Auth window shows a message about the failure
- We duplicate this info

sivaschenko:
@chalov-anton @coderimus all information that we have is displayed in the auth window. As far as I know there is no way to retrieve additional details

coderimus:
@sivaschenko just an update: I am working on this issue at this moment.

AdobeIms/view/adminhtml/web/js/action/authorization.js
```diff
@@ -113,11 +113,6 @@ define([
                 if (authWindow.closed) {
                     clearTimeout(stopWatcherId);
                     clearInterval(watcherId);
-
-                    // eslint-disable-next-line max-depth
-                    if (window.adobeIMSAuthWindow && window.adobeIMSAuthWindow.closed) {
-                        deferred.reject(new Error('Authentication window was closed.'));
```
sivaschenko, line 119: @coderimus we have to reject the promise when window is closed even if we wouldn't like to display the error

sivaschenko (changes requested):
Thanks for the updates @coderimus , please see my comment

coderimus:
@sivaschenko thank you for the review. I added the requested changes.

sivaschenko:
Hi @coderimus , I am closing this PR as Adobe IMS has migrated to https://github.com/magento/adobe-ims

PR #1545, opened 2020-06-26
[Stub] Provide save asset file support

coderimus:
### Description (*)
This PR provides the save image emulation. This achieved by using the `Magento\Framework\Filesystem\Driver\File` functionality instead of the `Magento\Framework\Filesystem\Driver\HTTPS`.

### Fixed Issues (if relevant)
#1544 

### Manual testing scenarios (*)
N/A

PR #1549, opened 2020-06-30
[WIP] [Stub] Added ability to download images via http

sivaschenko:
Added ability to download images via http for the stub


coderimus:
@magento run Functional Tests EE

AdobeStockStub/Model/FileGenerator.php
```diff
@@ -112,22 +82,16 @@ private function constructStubFileData(int $iterator): array
...
+    private function getImageUrl(): string
     {
-        if (empty($this->stubImages)) {
-            $this->stubImages = [
-              $this->assetRepository->getUrl('Magento_AdobeStockStub::images/1.png'),
-              $this->assetRepository->getUrl('Magento_AdobeStockStub::images/2.png'),
-              $this->assetRepository->getUrl('Magento_AdobeStockStub::images/3.png')
-            ];
-        }
+        return $this->assetRepository->getUrl('Magento_AdobeStockStub::images/' . rand(1, 3) . '.png');
```
coderimus, line 95: Working on test coverage with the stub module I found that the number of images should be > than 3 :) Because one of the tests sorts images and asserts that the image # 2 in ASC sort will not be the same when DESC is applied.
ASC            DESC
1,2,3           3,2,1
So I introduced next: https://github.com/magento/adobe-stock-integration/pull/1545/files#diff-16e3c853263dc10709168cd7b54e87baR118

coderimus (changes requested):
@sivaschenko thank you for you investigation and provided changes! Please, check my comment related to the stub images number.

AdobeStockStub/Model/FileGenerator.php
coderimus: @sivaschenko but this is mot critical in this PR. If it solves the main problem I will adjust this part with my changes in another PR.

coderimus:
@sivaschenko I tested locally. Doesn't work for me :(

sivaschenko:
Hi @coderimus I am not able to reproduce this issue on my local. Considering that it is reproducible for you, can you please remove the error suppression from https://github.com/magento/magento2/blob/735579d8aa14f7275d92818848f4a405191e7cb2/lib/internal/Magento/Framework/Filesystem/Driver/Http.php#L88
And put a debugger breakpoint in the https://github.com/magento/adobe-stock-integration/blob/261458c09c1596fb16a8915589560cdab402de2d/AdobeStockImage/Model/SaveImageFile.php#L67
After that try saving an image with debugger and share the exception.
That will be really helpful, thank you!

sivaschenko:
@magento run Functional Tests CE,Functional Tests EE,Functional Tests B2B

coderimus:
@sivaschenko thank you for the update and testing. I will test this today and report you back.

coderimus:
@sivaschenko I rebuild my local environment and run the second one containerized. Problem solved with your PR. I was able to save images! That's cool. Think that the problem was in my local permissions. Thank you for your assistance! 👍

coderimus (approved):

sivaschenko:
@coderimus thanks for checking it, but the MFTF tests are failing because of this issue)

PR #1611, opened 2020-07-20
[Refactoring] Move MediaGallery related modules to Magento core

lenaorobei:
### Description (*)
Review fixes after https://github.com/magento-engcom/magento2ce/pull/3943 

1. Related to magento/adobe-stock-integration#1578

Need to check following areas
1. Check messaging for images deleting + batch
2. Media gallery and media content synchronization
3. Make sure Enhanced media gallery is disabled by default.


lenaorobei:
@magento run all tests

lenaorobei:
@magento run all tests

lenaorobei:
@magento run all tests

MediaGalleryRenditionsApi/Api/ConfigInterface.php
```diff
@@ -6,7 +6,7 @@
 
 declare(strict_types=1);
 
-namespace Magento\MediaGalleryRenditionsApi\Model;
+namespace Magento\MediaGalleryRenditionsApi\Api;
```
sivaschenko: Th config interface was not meant to be used outside of the renditions module (but can be implemented by extensions). The proper namespace for such interface in the Api module is `Model`. What do you think @sidolov @lenaorobei

MediaContentSynchronization/Model/Consumer.php
```diff
@@ -12,7 +12,7 @@
 /**
  * Media content synchronization queue consumer.
  */
-class Consume
+class Consumer
```
sivaschenko: I wouldn't change the naming to keep the consistency. I.e. for Magento, it's more common to use action based name (`SaveAsset`) rather than object based (`AssetSaver`), especially when the only public method is `execute`. What do you think @lenaorobei @sidolov ?

AdobeStockImageAdminUi/etc/config.xml
```diff
@@ -1,16 +0,0 @@
-<?xml version="1.0"?>
-<!--
-/**
- * Copyright © Magento, Inc. All rights reserved.
- * See COPYING.txt for license details.
- */
--->
-<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:module:Magento_Store:etc/config.xsd">
```
sivaschenko: That config.xml change has already been merged to 2.0-develop

sivaschenko (changes requested):
Thanks for the PR, please see my review comments

AdobeStockImageAdminUi/etc/config.xml
lenaorobei: Not sure I get what you mean. I removed this config from `AdobeStockImageAdminUi` but it is present in `MediaGalleryUi` with `0` value. 
https://github.com/magento/adobe-stock-integration/pull/1611/files#diff-7c5661c33a5e5b316c37db653ae2700cL12

MediaContentSynchronization/Model/Consumer.php
sidolov: In this case we need to rename all Writers and Readers for consistency

MediaGalleryRenditionsApi/Api/ConfigInterface.php
sidolov: We need to consider removing the interface and left only implementation marked as api

lenaorobei:
@magento run all tests

MediaGalleryRenditionsApi/Api/ConfigInterface.php
lenaorobei: I removed the interface. Did not mark class with `@api` though, since it is patch release.

MediaContentSynchronization/Model/Consumer.php
lenaorobei: I feel like `Consumer` sounds more natural. Left as is.

MediaGalleryRenditionsApi/Api/ConfigInterface.php
lenaorobei: @sivaschenko looks like we don't need `MediaGalleryRenditionsApi ` module for now at all.

sivaschenko: The interface is needed to be able to provide an alternative implementation.

MediaContentSynchronization/Model/Consumer.php
sivaschenko: Considering that there are different opinions regarding the class naming, shouldn't we exclude this change from refactoring and leave the name as it is? Or is that an important change?

AdobeStockImageAdminUi/etc/config.xml
sivaschenko: I guess this can be marked as resolved after the branch was updated

MediaGalleryRenditionsApi/Api/ConfigInterface.php
lenaorobei: Will move class to `MediaGalleryRenditionsApi/Model`.

lenaorobei:
@magento run all tests

MediaContentSynchronization/Model/Consumer.php
lenaorobei: Changed to `Consume` 🙂

lenaorobei:
@magento run all tests

lenaorobei:
@magento run all tests

lenaorobei:
@magento run Functional Tests CE, Sample Data Tests EE

sivaschenko (dismissed):

chalov-anton:
QA PAssed

Checked the following areas: 

:heavy_check_mark: Correct warning messages on deleting. Correct _number of images_ counter, and _used in_ counter
:heavy_check_mark: Performed `media-gallery: sync` and `media-content:sync` The images from _pub/media_ appeared in Media Gallery. The Used In is displayed on Image Details page
:heavy_check_mark: On a"fresh" installation, the _Media Gallery Enabled_ is set to **No** in Stores -> Configuration...

lenaorobei:
@magento run all tests

sivaschenko:
@magento run all tests

sivaschenko:
@magento run all tests

sivaschenko:
@magento run all tests

PR #1645, opened 2020-07-24
#1614: [Refactoring] PHP

### LINKED ISSUE #1614: [Refactoring] PHP
Related to https://github.com/magento/adobe-stock-integration/issues/1578

As per code review we need address following issues in PHP files.

- `MediaContentSynchronization/Console/Command/Synchronize.php:61`, `MediaGallerySynchronization/Console/Command/Synchronize.php:63` - Get rid of area emulation
- `MediaGalleryUi/Model/AssetDetailsProviderInterface.php:23` - Consider using AssetDetails data object instead of plain array
- `MediaGalleryUi/Plugin/ResizeSyncedFiles.php:36` - add return type
- `MediaGalleryUi/Ui/Component/Listing/Filters/Options/UsedIn.php` - It does not make sense to configure options via DI, it's better to return a plain array with all required information. Any 3d party developer is able to create a plugin in the case when the list should be extended. Also, localization will not work for values stored in the di.xml config.
- `MediaGallerySynchronizationApi/Api/SynchronizeFilesInterface.php:18` - Please, don't use generic PHP interfaces. It's better to introduce our own interface for that.
- `MediaGallerySynchronizationApi/Model/FetchBatchesInterface.php:22` - What's the purpose of this field? Could you please add a description.
- `MediaGallerySynchronizationApi/Model/GetContentHashInterface.php` - What's the reason to keep this interface in the API module? How external developers can use it/extend? It looks like internal implementation and shouldn't be exposed via API.
- `MediaGallerySynchronization/Model/Synchronize.php:67` - Move logic to the pool initialization. According to our technical vision - any object in Magento should be ready for use after initialization, it means that pool must contain only valid elements and client code doesn't need to perform any checks.
- `MediaGallerySynchronizationApi/Model/SynchronizerPool.php:30` - Add synchronizers validation.
- `MediaGalleryRenditionsApi/Model/ConfigInterface.php:` - Why this interface located under the Model folder instead of Api?
- `MediaGallerySynchronization/Model/Filesystem/SplFileInfoFactory.php` - Define our own interface and object instead of generic
- `MediaGallerySynchronization/Model/CreateAssetFromFile.php:24` - The class should be refactored to avoid coupling.
- `MediaGalleryIntegration/etc/adminhtml/di.xml:17` - Should be moved to MediaGalleryCatalogIntegration module
- `MediaGalleryIntegration/Model/OpenDialogUrlProvider.php:16` - Is it required to extend from DataObject? I don't see any use cases.
- `MediaContentSynchronization/Model/ResourceModel/GetOutdatedRelations.php:98` - Used incorrect exception type
- `MediaContentSynchronizationApi/Api/SynchronizerInterface.php` - Should be removed
- `MediaContentSynchronizationApi/Model/Entities.php:13` - Rename to GetEntities to be consistent with interface
- `MediaGalleryCatalogUi/Model/Listing/DataProvider.php:167` - get rid of hardcoded `entity_id`
- `MediaGalleryCatalogUi/Ui/Component/Listing/Columns/Thumbnail.php:67` - DataObject? According to the method signature, it expects the Product model.

yolouiese:
### Need to verify (*)
- If no issues will occur after fixes and changes have applied on files mentioned in this ticket

### Description (*)
Related to magento/adobe-stock-integration#1614

As per code review we need address following issues in PHP files.

- `MediaContentSynchronization/Console/Command/Synchronize.php:61`, `MediaGallerySynchronization/Console/Command/Synchronize.php:63` - Get rid of area emulation
- `MediaGalleryUi/Plugin/ResizeSyncedFiles.php:36` - add return type
- `MediaGallerySynchronizationApi/Api/SynchronizeFilesInterface.php:18` - Please, don't use generic PHP interfaces. It's better to introduce our own interface for that.
- `MediaGallerySynchronizationApi/Model/FetchBatchesInterface.php:22` - What's the purpose of this field? Could you please add a description.
- `MediaGallerySynchronization/Model/Synchronize.php:67` - Move logic to the pool initialization. According to our technical vision - any object in Magento should be ready for use after initialization, it means that pool must contain only valid elements and client code doesn't need to perform any checks.
- `MediaGallerySynchronizationApi/Model/SynchronizerPool.php:30` - Add synchronizers validation.
- `MediaGallerySynchronization/Model/CreateAssetFromFile.php:24` - The class should be refactored to avoid coupling.
- `MediaGalleryIntegration/etc/adminhtml/di.xml:17` - Should be moved to MediaGalleryCatalogIntegration module
- `MediaContentSynchronization/Model/ResourceModel/GetOutdatedRelations.php:98` - Used incorrect exception type
- `MediaContentSynchronizationApi/Model/Entities.php:13` - Rename to GetEntities to be consistent with interface
- `MediaGalleryCatalogUi/Model/Listing/DataProvider.php:167` - get rid of hardcoded `entity_id`


MediaGallerySynchronizationApi/Model/FetchBatchesInterface.php
```diff
@@ -19,7 +19,7 @@ interface FetchBatchesInterface
      *
      * @param string $tableName
      * @param array $columns
-     * @param string|null $modificationDateColumn
+     * @param string $modificationDateColumn
```
Nazar65: The modification date column can be null, this column needed to fetch only those entities wich was changed since last execution, so the Generator compare modification date column and if the value bigger then in last execution time they will include this item to the batch. this behavior was implemented to avoid fetch again items thats were fetched in previous synchronization

Nazar65: and if you don't like fetch items with modificationDate column you can skip this behavior by putting null to this parameter

Nazar65 (changes requested):

MediaGallerySynchronizationApi/Api/SynchronizeFilesInterface.php
```diff
@@ -15,7 +15,7 @@ interface SynchronizeFilesInterface
     /**
      * Create media gallery assets based on files information and save them to database
      *
-     * @param \SplFileInfo[] $items
+     * @param array $items
```
Nazar65:
```suggestion
     * @param string[] $paths
```

MediaGalleryUi/Plugin/ResizeSyncedFiles.php
```diff
@@ -32,6 +32,7 @@ public function __construct(Storage $storage)
      * @param \Closure $closure
      * @param array $files
      * @SuppressWarnings(PHPMD.UnusedFormalParameter)
+     * @return mixed
```
Nazar65:
```suggestion
     * @return void
```

MediaGallerySynchronizationApi/Model/FetchBatchesInterface.php
yolouiese: Hi @Nazar65, thanks for the review. I'll update my pull request with the requested changes. Thank you!

MediaGalleryUi/Ui/Component/Listing/Filters/Options/UsedIn.php
```diff
@@ -32,6 +32,34 @@ public function __construct(array $options = [])
      */
     public function toOptionArray(): array
     {
+        return $this->usedInOptions();
```
sivaschenko: please revert these changes, options should be configurable via DI.xml

MediaGalleryUi/etc/adminhtml/di.xml
```diff
@@ -41,30 +41,4 @@
             <argument name="collectionProcessor" xsi:type="object">Magento\MediaGalleryUi\Model\Api\SearchCriteria\CollectionProcessor</argument>
         </arguments>
     </type>
-    <type name="Magento\MediaGalleryUi\Ui\Component\Listing\Filters\Options\UsedIn">
```
sivaschenko: please keep this configuration

sivaschenko (changes requested):
Hi @yolouiese here are a couple of comments, please let me know when the PR is ready for review

Please keep the renditions config in `MediaGalleryRenditionsApi/Model/Config.php`

yolouiese:
@magento run all tests

yolouiese:
@magento run all tests

sivaschenko:
@magento run all tests

sivaschenko:
@magento run all tests

sivaschenko (approved):

PR #1804, opened 2020-09-03
Moved renditions modules to magento2

sivaschenko:
Moved renditions modules to magento2


sivaschenko:
@magento run all tests

